PR #2, opened 2020-03-08
Add a new-packages file and CI verification

f-f:
This PR adds a `new-packages.json` file, which will contain all the new packages that cannot be published on the Bower registry.
It already contains the package from https://github.com/purescript/package-sets/pull/552

The reasoning for having a separate file instead of putting everything in the `bower-packages.json` is that I'd like that file to be a snapshot of all the packages that are on the Bower registry, and nothing more.

It also adds a small CI setup that joins the two files containing packages, and verifies that there are no name clashes.

This implements the "stopgap solution" mentioned [here](https://discourse.purescript.org/t/the-bower-registry-is-no-longer-accepting-package-submissions/1103/4) until we get a more automated registry working

cc @hdgarrood


hdgarrood (approved):
Thanks!

PR #32, opened 2020-06-17
Add package: quickcheck-mbt

mikesol:
I've created a new package called `purescript-quickcheck-mbt`. It is close in spirit to Haskell's `quickcheck-state-machine` with a few tweaks. Thanks for your consideration!


f-f (approved):
Thanks :slightly_smiling_face:

PR #38, opened 2020-07-02
Add halogen-svg-elems

JordanMartinez:
This is a fork of [Statebox](https://github.com/statebox/purescript-halogen-svg)'s repo, which is a fork of the [original repo](https://github.com/kwohlfahrt/purescript-halogen-svg). I'm adding my fork here, so i can implement the next few Elm Examples in the cookbook repo. 

If Statebox wants to push their fork here in the future in place of mine, by all means, please do.

### DIFF
--- a/new-packages.json
+++ b/new-packages.json
@@ -20,5 +20,5 @@
   "purescript-http-types-basic": "https://github.com/meeshkan/purescript-http-types.git",
   "purescript-graphql-gen": "https://github.com/meeshkan/purescript-graphql-gen.git",
   "purescript-graphql-validator": "https://github.com/meeshkan/purescript-graphql-validator.git",
-  "purescript-halogen-svg": "https://github.com/JordanMartinez/purescript-halogen-svg.git"
+  "purescript-halogen-svg-elems": "https://github.com/JordanMartinez/purescript-halogen-svg-elems.git"
 }
--- a/new-packages.json
+++ b/new-packages.json
@@ -19,5 +19,6 @@
   "purescript-graphql-parser": "https://github.com/meeshkan/purescript-graphql-parser.git",
   "purescript-http-types-basic": "https://github.com/meeshkan/purescript-http-types.git",
   "purescript-graphql-gen": "https://github.com/meeshkan/purescript-graphql-gen.git",
-  "purescript-graphql-validator": "https://github.com/meeshkan/purescript-graphql-validator.git"
+  "purescript-graphql-validator": "https://github.com/meeshkan/purescript-graphql-validator.git",
+  "purescript-halogen-svg": "https://github.com/JordanMartinez/purescript-halogen-svg.git"
 }


JordanMartinez:
Seems the name is already taken here: 
```
"purescript-halogen-svg": "https://github.com/AidanDelaney/purescript-halogen-svg.git",
```

thomashoneyman:
I'm actually quite surprised that the @AidanDelaney version compiles with this package set, as it looks like it's built against Halogen 2. The Statebox fork was the most up-to-date prior to your changes, but that looks like it's also stagnated recently.

If you plan to maintain your fork over time I think it's reasonable for your fork to take over the namespace. However, if Statebox wants the namespace in the future I think they have a claim to it as the original author of the library essentially blessed them to continue development of it.

I'm not aware of official policies that cover this scenario.

AidanDelaney:
Hi.  My `purescript-halogen-svg` is very much out-of-date.  I would welcome the reuse of this namespace by a more up-to-date implementation.  I am happy for @thomashoneyman or @JordanMartinez to make the decision of which package is known as `purescript-halogen-svg` and I will agree with thier judgement.

JordanMartinez:
> If you plan to maintain your fork over time I think it's reasonable for your fork to take over the namespace. However, if Statebox wants the namespace in the future I think they have a claim to it as the original author of the library essentially blessed them to continue development of it.

If I did take over that namespace and they wanted it back, it would create unnecessary conflict right? What if I published that library under a different name? Perhaps `purescript-halogen-svg-elems`? Then, they could still claim it in the future without us having to discuss how to handle possible issues (besides a package set one). If they don't, that name is still readable enough to explain what the library is: SVG elements for Halogen.

thomashoneyman:
That's fine with me!

f-f:
In order to not break people's builds I would like to minimize the amount of (1) package removals and (2) name transfers: i.e. ideally we'd not delete packages at all (only exceptions are legal things like DMCA takedowns) and we'd not have new packages replace old ones.

The problem that we'd have if we go ahead with this is that people that were using `AidanDelaney/purescript-halogen-svg` in their old Bower builds then they'd get `JordanMartinez/purescript-halogen-svg` once they migrate to a solver using the new registry? (note: this is unlikely since it'd require that they did not specify bounds, but it is possible in principle)

So I'd agree with picking a different name :+1:

JordanMartinez:
Ok, i've renamed it to `purescript-halogen-svg-elems`

thomashoneyman:
Separately, I do think the Statebox fork is the “official” one for the halogen-svg package and we could update the package in the sets to point at it. But thanks for switching names, Jordan!

JordanMartinez:
> I do think the Statebox fork is the “official” one for the halogen-svg package

I agree. I don't plan on doing more work on their fork, but if someone does submit a PR cleaning up things (or I discover I need more stuff to implement the Elm Examples in the cookbook), then it might include more additions.

JordanMartinez:
Can this get merged btw?

f-f (approved):

f-f:
@thomashoneyman packages in `package-sets` point to the same repo that they point to in the registry, so Statebox's repo should be published to the registry first (under another name than the current one)

PR #39, opened 2020-07-07
Update `purescript-halogen-svg` to point to statebox fork

JordanMartinez:
See #38 for more context.

Erik from Statebox was on the PS Meetup yesterday. Their fork isn't inactive as I thought. Also, it was my understanding that he would like to push their fork to the package set. I guess he hadn't been notified about the issue I opened in their corresponding fork. So, I thought I'd help out by updating the name to use their fork since Thomas had mentioned doing that in the PR linked above.

### DIFF
--- a/bower-packages.json
+++ b/bower-packages.json
@@ -569,7 +569,7 @@
   "purescript-halogen-select": "https://github.com/citizennet/purescript-halogen-select.git",
   "purescript-halogen-selects": "https://github.com/slamdata/purescript-halogen-selects.git",
   "purescript-halogen-storybook": "https://github.com/rnons/purescript-halogen-storybook.git",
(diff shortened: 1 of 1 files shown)


thomashoneyman (approved):

f-f:
How I'd like this to happen would be that @AidanDelaney would move his repo to Statebox (or if that's not possible, that Statebox would get access to that repo), rather than just renaming this here (cc @epost)

This is because if we just rename this here then `package-sets` would point to a different set of versions than the Bower solver, and we don't want that to happen. Makes sense?

f-f:
It seems like this has been inactive for a while - I'm tyding up our backlog to make it more actionable, and I think we can close this for now, and reopen it if we want to move forward with this change in the future.

PR #53, opened 2020-08-23
Proposal: Allow for multiple backends to be specified

csicar:
It might be useful to support specifying, that a package is compatible with 
multiple or all backends.

* If the `backends : []` is set, the registry's CI could also check, that no `foreign import` are used in the package itself

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,17 +119,19 @@ let TargetType =
       -- Optional output folder where the compiler will put its result.
       -- If not specified, the compiler will use "output"
       , output : Optional Text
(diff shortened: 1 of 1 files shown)


f-f:
In #76 we figured that at this point in time we won't attempt to add support for backend-specific things (such as "native dependencies"), as the cost of getting it wrong now is much higher than the opportunity cost of adding such support for these things at a later stage of the design of the registry, e.g. after a first version is up and running. So I'll close this for now, even though I agree that we definitely support this at some point

PR #65, opened 2020-09-10
Add point-free

ursi:
As far as I can tell, if I want to add a package to the registry, the first step is to write it in here.


thomashoneyman:
That’s right!

Issue #20, opened 2020-04-28
Backend specific dependencies


kl0tl:
Some packages have implicit dependencies on JavaScript packages that aren’t enforced anywhere and must be propagated manually. For instance `react-basic` unsurprisingly depends on `react` and `react-dom`, `uuid` depends on `uuid` and `uuid-validate`, …

At work we circumvent this by installing our packages with both psc-package (for PureScript sources and dependencies) and npm (an empty "files" array in package.json ensuring this only installs JavaScript dependencies). As you can imagine, versions installed by npm frequently differ from those in our internal package set so we hope for the best and things mostly work because we don’t update or JavaScript dependencies often. At least we only have one npm dependency to worry about.

I’d like to propose tracking backend specific dependencies and installers in packages targets to solve this, by giving the following type to backends instead of `Optional Text`:

```dhall
{ compile : Optional Text
, install : Optional Text
, dependencies : Map Text Text
}
```

Spago could then automatically install the backend dependencies with its specified `install` command (if any).

An alternative would be to define backends as an union so unambiguous fields can be ommited and types can be made as precise as needed per backend:

```dhall
< JavaScript : { installer < Npm | Yarn >, dependencies : Map Text Text } >
```

There’s probably better ways to do this and the registry perhaps doesn’t need to take this into account from the start, in any case I’m curious to hear other people thoughts and solutions about that.

hdgarrood:
I would much rather the registry not take this into account to start with, and perhaps not ever. I think it is very likely that if we tried, we wouldn't quite get it right at first, separate tools would arise to fill the gap, and then it would be too difficult to change the registry to address these problems, so there would just be a half-done implementation inside the registry which we'd have to maintain but which barely anyone actually used. In particular, there is not just one way of retrieving JS dependencies: you can use a CDN, you can install it using one of any number of npm clients (npm, yarn, etc), you can just download a particular version to your server and serve it yourself, etc. I think this design might force us to pick one particular approach, which to me seems too opinionated for a registry. In any case it definitely doesn't feel right to me for a PureScript package in the registry to be able to say "I should be installed via npm/yarn". That's something that should be up to the package consumer. See also, previous sort-of related discussions: https://github.com/purescript/purescript/issues/1850,  https://github.com/purescript/purescript/issues/2449

kl0tl:
[...]

PR #76, opened 2020-09-27
Registry Draft v2

f-f:
This pull request introduces a major revision of the "new PureScript Registry draft" that we currently have in the README of the `master` branch.
I'll call that "Draft v1", while this should be considered the "Draft v2".
Once this work graduates from the "draft" status (eventually after more revisions of this draft), it will be called "Registry v1" - this is important to note because we'll tie the version number to the breaking changes to the Dhall types of the Registry, currently in the [`v1`](./v1) folder.

## How to review this Pull Request

Before I get into the explanation of what has changed and how, I'd like to detail how I'd like the lifetime of this pull request to look like, and what I'd suggest the review workflow to be:
- since the changeset is admittedly giant, I don't expect anyone to fully look at it, and "reviewing it" in the traditional GitHub meaning is highly discouraged since that'd be a painful experience (of course you can still go ahead if you'd like that).
 Since 90% of the work here consists of rewriting the spec in the README, I'd strongly suggest [just having a read at the new draft](https://github.com/purescript/registry/blob/v1/README.md)
- because we are still in draft stage, it's least confusing for everyone if this work gets merged as soon as possible in `master`, as I'd find highly undesirable to be stuck with an outdated "draft v1" in there and the updated work only shown in this PR.
  However I believe it's useful to have this PR open so we have time to catch obvious oversights - so __I'll wait for 3 days, and then merge to `master`__
- I'd like to encourage questions, requests for clarifications, etc in GitHub reviews, but we'll move all of these discussions over to Issues once this is merged (i.e. we should not hold the merge on anything short of a fundamental design flaw)

## Notable changes

I'll assume here that you're familiar with the current design, and I'll summarize here a list of things that have changed since then:
- most importantly, the package Manifest storage format changes from Dhall to JSON
- however, we'll still conform the JSON Manifests to Dhall types to ensure they follow the Schema we specify here
- a contract of _forward-compatibility_ has been defined for future versions of the Manifest schema
- it will be required for the package Manifest to be versioned in the repository sources (but we'll generate those for all the Bower packages, so the transition will be smooth)
- we propose the Manifest file to be called `purs.json`
- we won't store the package Manifests in this repo, and the single source of truth for them will be the package tarballs. However, we'll store a copy of that in a `registry-index` repository, for easier bulk-access.
- we introduce a `Metadata` file for every package, to be stored in this repo and to serve as permanent storage for data needed for registry operations. This includes storing the SHA256 of the package tarballs.
- there's no focus anymore on storing the package tarballs as artifacts in GitHub Releases, and instead we'll start off with a S3-like storage maintained by the Package Team
- and at the same time the contract for adding new mirrors of Registry storage has been clarified, both for package managers and for anyone that would like to mirror the registry 
- source locations now support a "subdirectory" component, freeing packages from having to be in the root of the repository, and allowing multiple packages from the same repository
- the role of Trustees has been clarified
- an attempt to support `nativeDependencies` has been made, and subsequently abandoned. See #20 for context, and [this commit](https://github.com/purescript/registry/commit/928c069b917cd5e62cbe157293fb8658f9756a5d) for the (removed) code that was implementing this.
- the work on the "CI infrastructure" that is supposed to power all the Registry operations has been started. It was initially thought as a single Haskell executable, but for ease of operation and maintenance this has changed into a modular PureScript codebase, to be stored in this repo. You can see it [here](https://github.com/purescript/registry/tree/v1/ci)
- the "Registry API" for package publishers has been specified, and it will consist of "JSON over GitHub issues". The possible interaction with package managers has been specified too.
- the "implementation plan" has been expanded and detailed, see next section for more

Some of these changes have been discussed in issues in this repo, so I am not listing all the reasons for all the changes above (after all it's supposed to be a summary), so feel free to ask for more context and I'll do my best to link to relevant places and otherwise detail that.

## Where do we go from here

You can find more details for the implementation plan in the [relevant section of the draft](https://github.com/purescript/registry/tree/v1#implementation-plan).

Quoting from "what is happening now":

> 1. we're figuring out the last details of the package Manifest, which is the big blocker for proceeding further, since it will be baked into all the tarballs uploaded to the storage.
> 2. setting up a storage backend maintained by the Packaging Team
> 3. writing up the CI code to import the Bower packages as described above

(1) and (3) are mostly this PR, (2) is happening but it's blocked by (1).

To summarize: we need to commit to the first version of the `Manifest` Schema before we move forward with the rest of this.


README.md
```diff
@@ -103,321 +129,316 @@ A "compilation target".
...
+CI will verify that _all_ the following conditions hold:
+- the author of the issue is either one of the maintainers or one of the [Registry Trustees](#Registry-Trustees)
+- the version is less than __1 week old__
 
+If these conditions hold, then CI will:
+- move that package version from `published` to `unpublished` in the package `Metadata`
+- delete that package version from the storages
 
-## Unpublishing a package/release
+Unpublishing is allowed for security reasons (e.g. if some package was taken over maliciously),
```
hdgarrood, line 221: Malicious code poses a risk to anyone using the registry and so I think we should always remove malicious code once we become aware of it, no matter how much time has passed. The possibility of malicious code ending up in the registry also makes me wonder if we might want to seek legal advice and put together a proper terms of service document, saying things like "you acknowledge that not all of the code in the registry has been vetted at that whenever you download and use code from the registry you do so at your own risk".

README.md
```diff
@@ -103,321 +129,316 @@ A "compilation target".
...
+## Package metadata
 
-Proposed window for possible unpublishing: one week from the publishing of that version.
+Every package will have its own file in the `packages` folder of this repo.
+
+You can see the schema of this file [here](./v1/Metadata.dhall), and the main reasons for this file to exist are to track:
+- the upstream location for the sources of the package
+- published versions and the SHA256 for their tarball as computed by [our CI](#Adding-a-new-package).
+  Note: these are going to be sorted in ascending order according to [SemVer](https://semver.org) - when in
+  doubt the sorting provided by [the `semver` package on NPM](https://www.npmjs.com/package/semver#comparison) is correct.
```
hdgarrood, line 234: I don't think there's a need to refer to the `semver` package on npm, is there? The spec tells you how to sort versions unambiguously in all cases except for when the versions have the same major, minor, patch, and prerelease data, but their build metadata differs. However, I think this case should not come up, as the registry should not accept a package submission whose version is the same as an already-uploaded version in all respects except for the build metadata.

README.md
```diff
@@ -103,321 +129,316 @@ A "compilation target".
...
+- unpublished versions together with the reason for unpublishing
+- GitHub handles of package maintainers, so that we'll be able to contact them if any action is needed for any of their packages
```
hdgarrood: Is there a missing word here? "GitHub handles of package maintainers"

README.md
```diff
@@ -103,321 +129,316 @@ A "compilation target".
...
-..and then we'd use that package set as the provider of our package versions.
+Trustees will try to contact the maintainer of a package for __4 weeks__ before publishing a new revision, except if the author
+has opted out from this process, in which case they won't do anything.
 
-### Overriding a Package
+Trustees will __not__ change the source of a package, but only its metadata in the `Manifest` file.
 
-Building on the example above, what if we want to override the version of `effect`
-in the set with a local package?
+Trustees are allowed to publish __new revisions__ (i.e. versions that bump the `pre-release` segment from SemVer), to:
```
hdgarrood, line 275: I'm not sure using the semver prerelease segment will work so well for this. Firstly, prerelease versions are considered less than non-prerelease versions according to semver, so v1.0.0 compares greater than v1.0.0-r1. Secondly, doesn't this prevent package authors from using the prerelease segment for their own purposes? I wouldn't really mind if we reserved the prerelease segment for our own use, but in that case we would be diverging from semver (if we don't allow authors to use all of the components of semver, we aren't really using semver) so I think we ought to be more upfront about that if that's what we go with.

hdgarrood (commented):
This is starting to look really nice :heart:

README.md
```diff
@@ -1,99 +1,125 @@
...
-- with version bounds for packages: so that packages can express compatibilities between themselves
-- with a way for trusted editors to edit version bounds without republishing: so that faulty bounds can be corrected immediately without having to wait for package authors
-- easy to implement: the less code we have to maintain, the better
-- easy to secure: since this is a sensitive piece of the infrastructure, the smaller the attack surface is, the better
-- easy to host: we'd eventually like to have multiple copies of this registry available, so it should be easy to replicate the hosting setup
+Here's a non-comprehensive list of desiderable properties/goals that we'd generally like to
+achieve when talking of a PureScript registry, and that we think this design covers:
+- **independent**: we're coming from a situation when relying on a third party registry has
+  faulted us, so we'd need something that we can control and possibly host ourselves.
+- **immutabile**: published packages are immutable - once a version has been published
```
JordanMartinez: Small typo: immutabile -> immutable

f-f: More like "accidental Italian" :smile:
Thanks for the catch!

README.md
f-f: Would "GitHub usernames of package maintainers" sound better?

Tangent: I thought about putting email addresses in there, but I am not sure about having them in clear in this way, but other registries do it as well.
On the other hand, it would be hard for us maintainers of the Registry to be able to backfill email addresses, while it would be quite easy to fill in GitHub usernames. So I went with that in the end.

hdgarrood: Ohh, I see what you mean, I thought "handles" was being used as a verb lol. Yeah, I think "usernames" would avoid that confusion at least.

README.md
f-f: The keyword here is "when in doubt". The bottom line here is that we go by the standard, but in practice our CI will use the NPM package (which is the main implementation of the standard anyways), so I thought it was useful to refer to that here so that in case of inconsistencies it would be easy to verify if the problem is in the standard, in this implementation or in other implementations.

README.md
f-f: 1. I agree on allowing us to remove malicious code in any case, but in that case we'd become judges of "what's malicious"? I think a good policy could be "something that has a [CVE](https://cve.mitre.org/) assigned", but that might be too strict as PureScript packages might not be considered "common". We could get in touch with them to coordinate on this though, they are supposed to be _very_ responsive by definition :slightly_smiling_face: 
2. I also agree on having some legalese that discharges us of responsibility of what we host, but I wonder how NPM does this?

README.md
f-f: Great points. I wouldn't like us to diverge from SemVer, but other than that I have no strong opinion on how to do this.
Do you have a concrete idea that we could write down here?

README.md
f-f: Fixed!

README.md
hdgarrood: I think what counts as malicious needs to be based on the trustees' judgement. I don't think "has a CVE assigned" would work, because CVEs are for vulnerabilities, many of which are not malicious, and also as I understand it it's not too hard to get a CVE assigned for a vulnerability. I think it's quite rare than an accidental vulnerability should result in a package being taken down from the registry, because it's likely that a large chunk of the users of any given library aren't using it in such a way that the vulnerability affects them, or are using it in a context where security isn't a concern. For example, this vulnerability probably wouldn't be worth breaking people's builds over, since it's rare than an attacker controls the HTTP request method: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-26116

README.md
hdgarrood: I just looked over the semver spec again and it does say that prerelease identifiers and build metadata are optional so actually I take that back, we are well within our right to only accept what the spec describes as "normal" versions into the registry (i.e. those which don't have prerelease identifiers or build metadata), and since it's quite rare to use prerelease identifiers and build metadata in practice (at least for libraries), I think it may even be a good idea to reject versions which use them. I also think it would be nice to set the registry architecture up so that it's impossible for revisions to affect anything other than the package metadata. What Hackage do is handle revisions of metadata for a version separately from the version tarball itself, so that there's only ever one tarball for a version, and revisions of the metadata are stored separately. It then becomes the job of the registry client to fetch the most recent version of the metadata alongside the package tarball. That approach sounds sensible to me, and I think it suggests that metadata revisions are separate from package versions. So maybe this is just a question for the package index?

README.md
f-f: I see, yeah. I went looking for NPM's policy on this, and it looks like [they consider "malicious code" as "unacceptable content"](https://www.npmjs.com/policies/open-source-terms#acceptable-content), so I think we could go for this as well.

Moreover, I pretty much agree with all [their legal policies](https://www.npmjs.com/policies/) as I think it's pretty sensible stuff (the ones I read at least). So we could have a read at them and just adopt the ones we like without even having to talk with a lawyer (and talk with one whenever we want to diverge from there)

README.md
```diff
@@ -1,99 +1,125 @@
...
-  from the security standpoint.
-- **first class support for Package Sets**: see the [relevant section for more info](#Package-Sets).
+A `Manifest` stores all the metadata (e.g. package name, version, dependencies, etc)
+for **a specific  version** of **a specific package**.
 
-## The `Package` schema
+Packages are expected to version in their sources a `purs.json` file, that conforms
+to the [`Manifest` schema](./v1/Manifest.dhall) to ensure __forwards-compatibility__
+with future schemas.
+This means that new clients will be able to read old schemas, but not viceversa.
```
thomashoneyman:
```suggestion
This means that new clients will be able to read old schemas, but not vice-versa.
```

README.md
```diff
@@ -1,99 +1,125 @@
...
+And the reason why forward (rathen than backwards) compatibility is needed is because
```
thomashoneyman:
```suggestion
And the reason why forward (rather than backwards) compatibility is needed is because
```

README.md
```diff
@@ -1,99 +1,125 @@
...
+package manifests are baked in the (immutable) package tarballs forever, which means that
+any client (especially old ones) should always be able to read that manifest.
 
-See [the `Package` type definition](./v1/Package.dhall) for an up to date representation.
+This means that the only changes allowed to the schema are:
```
thomashoneyman, line 82: Is it possible to "deprecate" a field by replacing it with a new one, with tools using the new one but falling back to the old one?

README.md
```diff
@@ -103,321 +129,316 @@ A "compilation target".
 Every target can have its own dependencies, source globs, etc.
 By convention a package needs to have at least one target called `lib`.
```
thomashoneyman, line 130: Is this part just enforced by CI? Or is there any reason to make `TargetType` either a) `Lib` (with an enforced "src/**/*.purs" source) or b) some other target?

thomashoneyman:
On the whole 👍!

README.md
f-f: That could theoretically work, but the reason why we have to leave old fields in is so that old clients will be able to read them (as they cannot rely on new fields). In this sense we'll never be able to stop writing data to the old fields, so I guess the answer to your question is "not really". Makes sense?

thomashoneyman (approved):

README.md
f-f: I am not quite comfortable with Hackage's approach of storing package sources and metadata revisions separately, as I think it has a couple of problems:
1. there is only one tarball for a version, which means that "hashing the package sources" doesn't guarantee integrity and security by itself, as you'd need to fetch a metadata file as well, and you'd need to guarantee integrity (i.e. hash it and distribute the hashes) either of this file or of the whole bundle. In the latter case, one then also needs to define how to compose the files to compute the hash. Preserving integrity of the metadata as well is necessary because things like "changing version bounds" could introduce malicious code if not supervised, and not guaranteeing integrity for them would mean that people would get them in their build believing that the version hasn't changed (because the hash did not), rendering builds nondeterministic and insecure.
2. doing this would mean that the `registry-index` repo would become a source of truth, which means that we'd either have to maintain these two repos in sync (e.g. the moment we add metadata in there we also need to sync the list of versions here, etc) or unify them
3. ...but would storing metadata in here mean that packages wouldn't version that? If metadata is not stored alongside the package sources, then there would be no way to use a package without going through the Registry (and I can see some people not wanting to do that because of security or other company concerns), as they'd have to figure out how the package is defining dependencies, bounds, etc. If we store the metadata alongside the sources instead, then we have the question of "which metadata is correct, the one here or the one in the registry?". This last reason alone is why the current design ditches this aspect (i.e. storing manifests in here and optionally in packages) from the previous draft.

About prerelease identifiers: some packages have been using the pre-release segment (e.g. [Halogen](https://github.com/purescript-halogen/purescript-halogen/releases/tag/v5.0.0-rc.9) or [aff](https://github.com/purescript-contrib/purescript-aff/releases/tag/v4.0.0-rc.6)) and I think it's good to allow them to, as it has an important role in package versioning.

I also read again through the SemVer spec, and found an issue which might actually help us here: it looks like it considers two version with everything equal but the _build metadata_ as equals in the sorting. I.e. it ignores _build metadata_ when sorting. This is a problem for us because we _need_ get a stable sorting of releases (to figure out the last version of a package), so we would either have to:
- disallow releases with _build metadata_ altogether
- or reserve the _build metadata_ segment _only_ for Trustees to cut new revisions (so going back to that idea from the previous version of the draft), and expand the sorting so that it would consider version with build metadata as "later" than without. The ordering of different build metadata segments should be no problem, as the spec already defines how to sort prerelease segments, and their grammar is the same, so we could just reuse that.
  This behaviour seems to be what `apt` does as well - distros patch upstream packages and add build metadata so that the package manager picks the patched versions as newer - and I think it works very well there

Note that both of the above options means that we'll slightly diverge from SemVer, but I'd consider this quite fine, since it's basically just getting rid of undefined behaviour.

README.md
f-f: I envisioned this as being just a CI check, but we could do something fancier as well here.
Right now the Dhall types are fairly basic, because there are some constraints on their values, which need to:
1. have a decent JSON representation
2. make it easy to guarantee forwards-compatibility

README.md
hdgarrood: 1. Aren't we intending to provide access to individual package manifest files without downloading a whole package tarball anyway, via the package index? Ensuring the integrity of those package manifest files is already a problem we'll need to deal with, surely?
2. In that case, could we store package manifest files in the storage backend alongside the package tarballs and call that the source of truth for them? Then, the package index would always be derived from the storage backend, and would not be a source of truth?
3. I think disallowing build metadata makes sense. We could also require that each package may only upload one version with the same major, minor, patch, and prerelease components. For example, if you've already uploaded `1.0.0+abc`, then I think the registry should reject a subsequent upload of `1.0.0+def`. I think defining an ordering for the build metadata would be a more serious violation of the semver spec, because the spec specifically says that you mustn't do that. If we go against this, I think it is likely to cause funny behaviour in clients which implement the semver spec accurately: for example, you couldn't have versions `1.0.0+abc` and `1.0.0+def` in a Set together. The only way I can make sense of the build metadata ordering requirement from the semver spec is if package registries should be refusing package uploads which differ from an existing version only in the build metadata component, i.e. uploading two different versions with the same major, minor, patch, and prerelease components should be disallowed.

f-f: How would Trustees publish revisions if we disallow build metadata? And why would sorting by that a "serious violation"? SemVer doesn't say "you shouldn't do that because it's bad", it just says "we don't do that in SemVer".
Registry clients are supposed to implement a spec that we define here. If we say "it's SemVer plus ordering by build metadata", then that is the spec.

hdgarrood: By treating revisions as a separate thing from package versions? It’s a more serious violation in my mind because it’s not just filling a gap in the spec, it’s going against something the spec explicitly says. The versioning libraries that exist aren’t “semver plus build metadata,” they’re just semver.

README.md
f-f: Opened #79 to track this

README.md
f-f: Opened #80 to track this, let's move over there

f-f:
@hdgarrood @thomashoneyman I merged this and opened issues for the review comments where we figured out action points or where I felt there's the need for more discussion, feel free to open more issues if you feel like I didn't address some concern or if I missed some other actionable item.

Thanks everyone for the thoughtful reviews! :slightly_smiling_face:

PR #81, opened 2020-10-08
Update URL for `purescript-css`

maxdeviant:
[`purescript-css`](https://github.com/purescript-contrib/purescript-css) now lives in the `purescript-contrib` organization, so the repository URL needed to be updated appropriately.


thomashoneyman:
@f-f are we maintaining the original urls and relying on redirects?

f-f:
The ‘bower-packages.json‘ file is meant as an archive and not supposed to be updated. It is also temporary.
So I'd prefer to not update it at all if things are not broken, since its main role is to just be a mapping from names to repos (I.e. no one uses this to download stuff AFAIK)

PR #97, opened 2020-11-20
fix link for polymorphic-vectors, canvas-action and game

artemisSystem:


f-f:
As a policy we don't change URLs for published packages, what's the reason for this PR?

artemisSystem:
I changed my github username, so the old links could stop working at any time. It also won't let me publish updates to pursuit currently, because the link in the registry is wrong

hdgarrood:
@f-f Can you remind me what the reason was for disallowing repo relocations? I’m not sure this will be a workable policy long term; I think we probably need to find a way to allow this eventually

f-f:
Sorry for delay in getting back on this, I was on vacation last week.

@artemisSystem changing username is a good reason to change the URLs here, so let's go ahead with this

@hdgarrood the reason for (mostly) forbidding repo relocation is basic security precautions: if we allow people to just point packages to different locations it's easier to introduce malicious takeovers. Makes sense?
I am in favour of having it happen for some cases though, and we might be able allow this (and even automate it) once the registry is not in a transitional state anymore, since we'll have people directly publishing to us rather than us trying to sync with Bower.

f-f (approved):

hdgarrood:
Yes, that makes sense, thanks. :+1:

f-f:
We cannot merge this yet though, as we first need to fix CI applying the same fix as https://github.com/purescript/package-sets/pull/749

artemisSystem:
I see, thanks 🙂

PR #98, opened 2020-11-26
Add abc-parser

newlandsvalley:


thomashoneyman:
I’m not able to do this right now, but this is failing because of the deprecated add_path command in GitHub Actions — the same issue as in the package-sets repository. We’ll have to apply the same update here and then merge your pr!

newlandsvalley:
Thanks, @thomashoneyman - yes I thought this was the case.  Not something I feel confident about fixing myself, but there's no rush.

Issue #16, opened 2020-04-25
Support packages with sources not in `src`


f-f:
For any package published on Bower right now it is possible to assume that one would get their source files with the following glob: `src/**/*.purs`

In https://github.com/purescript/spago/issues/288 we explored what it would take to lift that limitation (it's a good read to get some context on what follows here)

One of the main blockers for that was the assumption coming from Bower registry that "one package = one repo". Since here we're moving away from the git-based model and towards packaging sources in immutable archives, I thought we could take this occasion to allow publishing packages with sources in arbitrary paths.
This would allow things like publishing several package from the same monorepo.

There has been some discussion in #4 about this, and we're spinning off this issue to keep track of a concern expressed there:

> The issue I'm worried about with `sources` is basically for setups which don't use Spago; I want to avoid making life difficult for people who are still using other things like psc-package or pulp/bower. Yes, it's probably not too much effort to switch a build script of just a few lines in most cases, especially with `spago init`, but it's harder if the assumption is built into more complex tooling, such as, say, editor plugins. This assumption is built into `purs publish` right now too. It becomes significantly more complicated if these build scripts or tooling can't/don't want to assume the presence of Spago on the machine in use, or don't want to add Dhall as a dependency.
>
> I would be happy if we could have the curator require that the `sources` for the `lib` target for any published package are just `["src/**/*.purs"]`, at least to start with; I could see us relaxing that later once most of the ecosystem is using this registry.

hdgarrood:
> This would allow things like publishing several package from the same monorepo.

This ought to be possible anyway with the new design even if we require source files to be in `src`, right? If you have multiple packages in a monorepo, I would expect that they would each be in separate directories with separate package manifest files, and the `sources` key in the package manifest file would be interpreted relative to the directory in which the manifest file lives.

f-f:
There is now (as of #76) support for "multiple manifests in the same repo" via the `subdir` key in the `Repo` type.

However, I think the concern expressed of "we should limit to one package only until the ecosystem is using this" is still valid, and more specifically I'd like to have support for this in Spago before we allow the publishing of packages that specify a `subdir`. Makes sense?

hdgarrood:
I still don't understand what we gain by making source paths configurable by a package author, over having the rule be "the visible source files are always those which are located in the `src` folder relative to the package manifest file, and only those". That still permits managing a number of packages together in a monorepo, and as far as I'm aware that is the only concern which has been brought up in the context of this issue.

[...]

PR #107, opened 2020-12-19
Support multiple package versions in monorepos

f-f:
..by specifying the `version` field inside the `Manifest`, and not relying anymore on the assumption "one git tag == one release", but instead using any ref to fetch the sources.

This is related to #16

### DIFF
--- a/examples/mysql/v4.1.1.json
+++ b/examples/mysql/v4.1.1.json
@@ -25,10 +25,11 @@
     }
   },
   "repository": {
-    "version": "v4.1.1",
+    "ref": "v4.1.1",
     "githubOwner": "oreshinya",
     "githubRepo": "purescript-mysql"
   },
   "name": "mysql",
+  "version": "v4.1.1",
   "license": "MIT"
 }
--- a/examples/mysql/v4.1.1.json
+++ b/examples/mysql/v4.1.1.json
@@ -26,7 +26,6 @@
     }
   },
   "repository": {
-    "ref": "v4.1.1",
     "githubOwner": "oreshinya",
     "githubRepo": "purescript-mysql"
   },
--- a/examples/mysql/v4.1.1.json
+++ b/examples/mysql/v4.1.1.json
@@ -1,4 +1,5 @@
 {
+  "version": "v4.1.1",
   "targets": {
     "lib": {
       "sources": [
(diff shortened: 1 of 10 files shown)


examples/mysql/v4.1.1.json
```diff
@@ -25,7 +26,7 @@
     }
   },
   "repository": {
-    "version": "v4.1.1",
+    "ref": "v4.1.1",
```
thomashoneyman: Should this be a commit rather than the tag?

f-f: Great catch - the code that fetches Bower releases will have to be updated slightly, but a commit is definitely better here

f-f: Thinking more about it I don't think we can include a commit hash here - if someone is versioning the manifest in their repo, they wouldn't be able to publish at the next commit because they wouldn't know which commit hash is coming when they update the manifest. So one would have to do two commits to publish something - at this point I'm doubting of the utility of this field here at all, is there any value in storing in the Manifest which ref a version is from?
We already have it in the `Update` data structure:

https://github.com/purescript/registry/blob/05d0cf31de06ff1ea4ed99103499e3b8258007ed/v1/Operation.dhall#L11

thomashoneyman: I'm not making the connection between the `Update` data structure and a way to look up a ref associated with a version in the future. This sounds like we'd only ever store the version number without storing the commit, too. Am I missing something?

f-f: Right, I'll try to give more context for my reasoning.

Our main requirement is: we need to tell the registry the exact ref to use when pulling a package version for upload.
An optional but desirable corollary is that we'd like to store these refs in a somewhat permanent storage.

Now, it makes sense that we tell the Registry which ref to use through its API, i.e. when sending an `Addition` or `Update`. 
But this information is currently not stored anywhere - however storing such ref into the package Manifest is clumsy, because one would need two commits to publish something: one where they bump the `version` field, and one where they update this `ref` field to have the commit hash of the previous commit.

So here I'm proposing that we remove entirely this `ref` field from the `Manifest`, and include the `ref` only in the various API `Operations`. Then we could store such `ref`s in a more "permanent" way in the package `Metadata` here in the Registry repo.

Makes sense?

thomashoneyman: I see, it’s still in the metadata.

thomashoneyman (approved):

Issue #95, opened 2020-11-11
Implement Bower mass-import


f-f:
We already have some code that [can get all the Bower package manifests and generate our manifests](https://github.com/purescript/registry/blob/518de1833784650e56c97bb58eb122b7bdc21449/ci/src/BowerImport.purs#L1), so in order to be able to do the mass-import of Bower packages to the Registry we need to implement some code that:
- downloads the tarball for a commit hash from a GitHub repo
- sticks the generated manifest inside
- repackages the whole thing
- computes the sha256 of the new tarball

Note that we are planning to have CI checks for new packages, and would be good to run them for imported packages too:
- all the checks listed in #23 
- the name check in #3 

Plus we should convers the semver ranges in the dependencies to simpler ones when generating the manifest, see #43

JordanMartinez:
Based on [the docs](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#download-a-repository-archive-tar), we'll need to download an archive via `https://api.github.com/repos/USER/REPO/tarball/REF`. For example, https://api.github.com/repos/purescript/purescript-prelude/tarball/v4.1.0

We can add the manifest inside the file via `tar --append v4.1.0.json --file=downloaded-archive.tar` and compute the sha256 sum via `sha256sum result.tar`.

Not sure whether this is an issue, but the downloaded tar will store all of its content in top-level folder:
```
$ tar -tf prelude.tar 
purescript-purescript-prelude-7a691ce/.eslintrc.json
purescript-purescript-prelude-7a691ce/.gitignore
purescript-purescript-prelude-7a691ce/.travis.yml
purescript-purescript-prelude-7a691ce/LICENSE
purescript-purescript-prelude-7a691ce/README.md
purescript-purescript-prelude-7a691ce/bower.json
purescript-purescript-prelude-7a691ce/package.json
purescript-purescript-prelude-7a691ce/src/Control/Applicative.purs
purescript-purescript-prelude-7a691ce/src/Control/Apply.js
...
```

Correct me if I'm wrong, but I'm assuming we would want to remove the toplevel `purescript-purescript-prelude-7a691ce` directory, correct?
```
purs.json -- package manifest file
.eslintrc.json
.gitignore
.travis.yml
LICENSE
README.md
bower.json
package.json
src/Control/Applicative.purs
...
```

f-f:
[...]

Issue #84, opened 2020-10-12
Create a visual flowchart of how package addition, unpublishing, and version changing occurs


JordanMartinez:
This is one of my tasks this week from the PS Contrib working group.

JordanMartinez:
Based on the [Adding a new package section](https://github.com/purescript/registry#adding-a-new-package), this is my understanding:

![Registry - Adding a new package](https://user-images.githubusercontent.com/8413037/96505138-8e38eb80-120a-11eb-8392-07c70f8edfc4.jpg)

f-f:
@JordanMartinez this is really great! :clap:

milesfrain:
A legend for what the blue/gold colors represent (and possibly more colors / categories) would be helpful.

JordanMartinez:
Yellow represents actions in the workflow. Blue represents the state of the github issue.

thomashoneyman:
Closing this because we don't have plans to implement it, though contributions to the SPEC.md file are always welcome!

PR #110, opened 2020-12-21
Implement package upload to our storage backend on Digital Ocean

### LINKED ISSUE #94: Implement package upload to storage backend
We should have some code that given a tarball and package info (name, version), then uploads it to our S3-compatible storage backend (which is going to be on PureScript's org Digital Ocean account)

f-f:
FIx #94

### DIFF
--- a/ci/src/PackageUpload.purs
+++ b/ci/src/PackageUpload.purs
@@ -1,0 +1,57 @@
+module PackageUpload where
+
+import Prelude
+
+import Data.Array as Array
+import Effect (Effect)
+import Effect.Aff as Aff
+import Effect.Class (liftEffect)
+import Effect.Class.Console (log)
+import Node.FS.Aff as FS
+import S3 as S3
+
+type PackageInfo =
+  { name :: String
+  , version :: String
+  , revision :: Int
+  }
+
+type Path = String
+
+upload :: PackageInfo -> Path -> Effect Unit
+upload { name, version, revision } path = Aff.launchAff_ $ do
+  -- first check that the file path exists
+  fileContent <- do
+    fileExists <- FS.exists path
+    if fileExists
+    then FS.readFile path
+    else Aff.throwError $ Aff.error $ "File doesn't exist: " <> show path
+  -- connect to the bucket
+  let bucket = "purescript-registry"
+  log $ "Connecting to the bucket " <> show bucket
+  s3 <- liftEffect $ S3.connect "ams3.digitaloceanspaces.com"
+  -- check that the file for that version and revision is there
+  let filename
+         = name <> "/"
+        <> version
+        <> (if revision == 0 then "" else "_r" <> show revision)
+        <> ".tar.gz"
+  let listParams = { "Bucket": bucket, "Prefix": name <> "/" }
+  publishedPackages <- map _."Key" <$> S3.listObjects s3 listParams
+  if Array.elem filename publishedPackages
+  -- if the release is already there we crash
+  then Aff.throwError $ Aff.error $ "The package " <> show filename <> " already exists"
+  -- if it's not, we upload it with public read permission
+  else do
+    log $ "Uploading release to the bucket: " <> show filename
+    let putParams = { "Bucket": bucket, "Key": filename, "Body": fileContent, "ACL": "public-read" }
+    void $ S3.putObject s3 putParams
+    log "Done."
+
+{-
+
+main :: Effect Unit
+main = do
+  upload { name: "aff", version: "5.1.2", revision: 0 } "../examples/aff/v5.1.2.json"
+
+-}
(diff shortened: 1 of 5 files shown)


ci/src/S3.js
```diff
@@ -0,0 +1,48 @@
+const AWS = require('aws-sdk');
+
+if ("SPACES_KEY" in process.env && "SPACES_SECRET" in process.env) {
```
thomashoneyman, line 3: Are these environment variables we would implement as repository secrets, taken from the Digital Ocean settings?

ci/src/PackageUpload.purs
```diff
@@ -0,0 +1,57 @@
...
+import Effect.Class (liftEffect)
+import Effect.Class.Console (log)
+import Node.FS.Aff as FS
+import S3 as S3
+
+type PackageInfo =
+  { name :: String
+  , version :: String
+  , revision :: Int
+  }
```
thomashoneyman, line 16: Do we need any validation on this before we try and upload it? If we're relying on prior checks having run in CI before we get to this step then perhaps it would be useful to note that in a comment here.

ci/src/PackageUpload.purs
```diff
@@ -0,0 +1,57 @@
...
+        <> ".tar.gz"
+  let listParams = { "Bucket": bucket, "Prefix": name <> "/" }
+  publishedPackages <- map _."Key" <$> S3.listObjects s3 listParams
+  if Array.elem filename publishedPackages
+  -- if the release is already there we crash
+  then Aff.throwError $ Aff.error $ "The package " <> show filename <> " already exists"
+  -- if it's not, we upload it with public read permission
+  else do
+    log $ "Uploading release to the bucket: " <> show filename
+    let putParams = { "Bucket": bucket, "Key": filename, "Body": fileContent, "ACL": "public-read" }
```
thomashoneyman: Should we export a more PureScript-y compatibility type from `S3`? ie. we could write:

```purs
let putParams = { bucket, key: filename, body: fileContent, acl: PublicRead }
```

if we adjusted the `S3` module to be more like this:

```purs
module S3
  ( PutParams
  , putObject
  , -- not exporting js* functions and JS* types
  ) where

-- These functions and types can use normal PureScript naming conventions and
-- richer types

type PutParams =
  { bucket :: String
  , key :: String
  , body :: BufferOrString
  , acl :: ACL
  }

putObject :: S3 -> PutParams -> Aff PutResponse
putObject s3 params = jsPutObject s3
  { bucket: params.bucket
  , key: params.key
  , body: case params.body of
      ...
  , acl: case params.acl of
      PublicRead -> "public-read"
      ...
  }

-- Then the definitions below can be responsible for matching the FFI without also
-- having to be flexible for PureScript

type JSPutParams =
  { "Bucket" :: String
  , "Key" :: String
  , "Body" :: Buffer -- TODO: the SDK also accepts a string here
  , "ACL" :: String -- TODO: this should really be an enum
  }

jsPutObject :: S3 -> JSPutParams -> Aff PutResponse
jsPutObject = ...
```

ci/src/PackageUpload.purs
```diff
@@ -0,0 +1,57 @@
...
+    void $ S3.putObject s3 putParams
+    log "Done."
+
+{-
+
+main :: Effect Unit
+main = do
+  upload { name: "aff", version: "5.1.2", revision: 0 } "../examples/aff/v5.1.2.json"
```
thomashoneyman, line 53: Is there a way to upload in batch? This is giving me the sense we'd make a request per-package, which seems like it might be an issue.

ci/src/S3.js
f-f: That's right. I have a pair that I used locally, I can just add these to the secrets once this is in use

ci/src/PackageUpload.purs
f-f: No validation needed here - the broader context is that one would run the `upload` function as part of the [package addition flow](https://github.com/purescript/registry/issues/84#issuecomment-712407673), and as you can see we should run this "upload to the storage backend" right after we run the checks. So this means that we should do all the validation in there, and here only check if the tarball we're trying to upload doesn't already exist (which we do).
I believe we don't even need a comment here to explain this, because I'd expect that we'll shuffle this code around and organize it better under some `API` namespace to signal that this is part of the flow, or something like that. I don't even expect this type definition to stay around here, as we'll need it in other places too. We'll likely put `newtype`s in there too I think.

ci/src/PackageUpload.purs
f-f: Yeah, I was lazy here :smile: 
I pushed a nicer interface in the last commit

ci/src/PackageUpload.purs
f-f: This is going to be part of the upload flow for a single package version, and if we go for what I detailed in https://github.com/purescript/registry/issues/95#issuecomment-748678539 then we won't have any batch-like process anywhere

thomashoneyman (approved):

PR #126, opened 2021-02-13
Add missing web libraries and update stale repository URLs

kl0tl:
This pull request updates “stale” packages (whose repository URL redirects to another) and register missing web libraries with the help of some scripts.

<details>

```bash
#! /usr/bin/env bash
set -euo pipefail

function redirect_url {
  local repository_url; repository_url="$1"
  curl --head --output /dev/null --write-out '%{redirect_url}' --silent "${repository_url%.git}"
}

function main {
  local registry="$1"
  local org="$2"
  local repository_url
  local repository_redirect_url
  echo "Updating stale packages in $registry"
  for package_name in $(jq --raw-output --arg org "$org" 'with_entries(select(.value | test("/" + $org + "/"))) | keys[]' "$registry"); do
    repository_url="$(jq --raw-output --arg key "$package_name" '.[$key]' "$registry")"
    repository_redirect_url="$(redirect_url "$repository_url")"
    if [ -n "$repository_redirect_url" ]; then
      echo "$package_name"
      jq --arg package_name "$package_name" --arg repository_url "$repository_redirect_url.git" '.[$package_name] = $repository_url' "$registry" > "$registry.tmp"
      mv "$registry.tmp" "$registry"
    fi
  done
  git --no-pager diff "$registry"
}

case "${1-unexpected}" in
  '*'|purescript|purescript-contrib|purescript-node|purescript-web )
    main bower-packages.json "$1"; main new-packages.json "$1";;
  * ) echo "Expecting one of *, purescript, purescript-contrib, purescript-node or purescript-web.";;
esac
```

```bash
#! /usr/bin/env bash
set -euo pipefail

function main {
  local org="$1"
  local org_packages; org_packages="$(dhall-to-json <<< "https://raw.githubusercontent.com/purescript/package-sets/prepare-0.14/src/groups/$org.dhall")"
  local repository_url
  for package_name in $(jq --raw-output 'keys[]' <<< "$org_packages"); do
    for registry in bower-packages.json new-packages.json; do
      if [ "$(jq --arg package_name "purescript-$package_name" 'has($package_name)' "$registry")" = true ]; then
        continue 2
      fi
    done
    repository_url="$(jq --raw-output --arg package_name "$package_name" '.[$package_name].repo' <<< "$org_packages")"
    echo "Register $repository_url"
    jq --arg package_name "purescript-$package_name" --arg repository_url "$repository_url" '.[$package_name] = $repository_url' new-packages.json > new-packages.json.tmp
    mv new-packages.json.tmp new-packages.json
  done
  git --no-pager diff new-packages.json
}

case "${1-unexpected}" in
  purescript|purescript-contrib|purescript-node|purescript-web )
    main "$1";;
  * ) echo "Expecting one of purescript, purescript-contrib, purescript-node or purescript-web.";;
esac
```

</details>

I also renamed some packages for consistency with how they are named in their bower.json file (for instance `purescript-STsequence` was renamed to `purescript-stsequence` when its repository was relocated) and removed redundant packages (for instance `purescript-digraph` was initially registered as `purescript-dgraph` before being renamed and registered again, making `purescript-dgraph` redundant), but I’m not familiar enough with the registry to know if this is wanted.

This will ensure that core, contrib, node and web libraries are updated with their actual repository URL in the upcoming mass release.


f-f:
@kl0tl thank you for this! The URL updates are good, as are the renames for lowercasing things - however, I think we should not remove already registered names.
This is because doing so is a big security risk: if you have a package `foo` that someone's build depends on, if it gets de-registered then someone else could register the same name and publish malicious code under that name, and the owner of the build above might not realize. So as a rule we should never de-register names (we are going to have an unpublishing mechanism though, so that the name can be registered but versions of that package can be made unavailable)

kl0tl:
Ok! I kept the redundant packages: `purescript-dgraph`, `purescript-elm-compat`, `purescript-flarecheck`, `purescript-foreign-extra`, `purescript-functor-optics`, `purescript-heterogenous`, `purescript-hibachi`, `purescript-hyper-routing`, `purescript-hyper-routing-server`, `purescript-hyper-routing-xhr`, `purescript-nemo`, `purescript-node-dotenv`, `purescript-polyform-validators`, `purescript-profunctors`, `purescript-react-carpenter`, `purescript-react-cedar`, `purescript-sqlsquare`, `purescript-unix`, `purescript-unordered-containers` are now aliases to what they were renamed to. Should we also keep `purescript-aws-serverless-express` and `purescript-errorcontrol`? They’re both registered with https://github.com/purescript/purescript-prelude.git, which is clearly a mistake I think.

You mentioned lowercasing things, does this mean we should also rename `purescript-TypeAhead` to `purescript-typeahead`?

Should we keep the `purescript-polyform-batteries`, `purescript-kushikatsu` and `purescript-pselm-core` packages I registered to match the name of their repository? Should they be defined in new-packages.json instead? Or should we let their owners do it themselves perhaps?

f-f:
> Should we also keep `purescript-aws-serverless-express` and `purescript-errorcontrol`? They’re both registered with https://github.com/purescript/purescript-prelude.git, which is clearly a mistake I think.

Good call, let's remove these

> You mentioned lowercasing things, does this mean we should also rename `purescript-TypeAhead` to `purescript-typeahead`?

I mentioned this because the new registry rules will forbid packages with uppercase characters, so `TypeAhead` would not get in, but `typeahead` would. I am not sure if these two packages would be the same from the Bower perspective, nor what's their Bower name?

> Should we keep the `purescript-polyform-batteries`, `purescript-kushikatsu` and `purescript-pselm-core` packages I registered to match the name of their repository? Should they be defined in new-packages.json instead? Or should we let their owners do it themselves perhaps?

I am not sure I understand what's the issue with these packages, could you detail?

kl0tl:
> I mentioned this because the new registry rules will forbid packages with uppercase characters, so `TypeAhead` would not get in, but `typeahead` would. I am not sure if these two packages would be the same from the Bower perspective, nor what's their Bower name?

This package (`purescript-TypeAhead`) is registered with a lowercase repository URL (https://github.com/mckeankylej/purescript-typeahead.git) and is named `purescript-typeahead` in its bower.json but `bower install purescript-typeahead` installs https://github.com/ascjones/purescript-typeahead instead (also named `purescript-typeahead` in its bower.json), so perhaps we should keep `purescript-TypeAhead` in bower-packages.json for now and wait for the author or consumers of this package to register it as `typeahead` once the new registry is up?

> I am not sure I understand what's the issue with these packages, could you detail?

- `purescript-polyform-validators` was registered with https://github.com/lambdaterms/purescript-polyform-validators.git. It is named `purescript-polyform-batteries` in its bower.json though and its repository URL redirects to https://github.com/purescript-polyform/batteries-core.git.
- `purescript-hibachi` was registered with https://github.com/justinwoo/purescript-hibachi.git. It is named `purescript-kushikatsu` in its bower.json though and its repository URL redirects to 
-  https://github.com/justinwoo/purescript-kushikatsu.git.
- `purescript-elm-compat` was registered with https://github.com/rgrempel/purescript-elm-compat.git. It is named `purescript-pselm-core` in its bower.json though and its repository URL redirects to https://github.com/pselm/core.git.

Should we register these packages again as `purescript-polyform-batteries`, `purescript-kushikatsu` and `purescript-pselm-core` in new-packages.json?

thomashoneyman:
@kl0tl @f-f I would like to merge this as-is, and we can open an issue about the three packages that may need to be adjusted. That at least unblocks releasing the 0.14 libraries, and the three libraries in question can always be adjusted later.

thomashoneyman (approved):

f-f:
@thomashoneyman sounds good, let's go ahead and take care of the remaining things in a follow-up pull request

f-f (approved):

Issue #124, opened 2021-01-31
Manifest description


jamesdbrock:
It would be nice if Metadata had a `description` field for use by Pursuit, see https://github.com/purescript/pursuit/issues/431

https://github.com/purescript/registry/blob/bfa073019247439c000feb304555675711948514/v1/Metadata.dhall#L18-L25

f-f:
Hey @jamesdbrock! It would be indeed good to have a `description` field, but I think we should have it in the `Manifest`, rather than the `Metadata`. Here:

https://github.com/purescript/registry/blob/bfa073019247439c000feb304555675711948514/v1/Manifest.dhall#L13-L24

The difference being that the `Manifest` is totally in the hands of the package authors as it's versioned together with the package sources, while the `Metadata` resides in this repo, it's meant to be 100% handled by our CI automation (as it contains information for CI itself), and contains sensitive information (the hashes of the package tarballs), so pull requests to it will likely be forbidden.

Issue #80, opened 2020-10-07
Allowed versions and Trustees publishing


f-f:
In https://github.com/purescript/registry/pull/76#discussion_r495608601 we figured that using SemVer's "prerelease segment" for Trustees to publish versions, as the way the spec orders them is not the one we'd like.

I'll report here the discussion from the thread, as a starting point for further discussion:

> @hdgarrood: I'm not sure using the semver prerelease segment will work so well for this. Firstly, prerelease versions are considered less than non-prerelease versions according to semver, so v1.0.0 compares greater than v1.0.0-r1. Secondly, doesn't this prevent package authors from using the prerelease segment for their own purposes? I wouldn't really mind if we reserved the prerelease segment for our own use, but in that case we would be diverging from semver (if we don't allow authors to use all of the components of semver, we aren't really using semver) so I think we ought to be more upfront about that if that's what we go with.

> @f-f: Great points. I wouldn't like us to diverge from SemVer, but other than that I have no strong opinion on how to do this.
Do you have a concrete idea that we could write down here?

> @hdgarrood: I just looked over the semver spec again and it does say that prerelease identifiers and build metadata are optional so actually I take that back, we are well within our right to only accept what the spec describes as "normal" versions into the registry (i.e. those which don't have prerelease identifiers or build metadata), and since it's quite rare to use prerelease identifiers and build metadata in practice (at least for libraries), I think it may even be a good idea to reject versions which use them. I also think it would be nice to set the registry architecture up so that it's impossible for revisions to affect anything other than the package metadata. What Hackage do is handle revisions of metadata for a version separately from the version tarball itself, so that there's only ever one tarball for a version, and revisions of the metadata are stored separately. It then becomes the job of the registry client to fetch the most recent version of the metadata alongside the package tarball. That approach sounds sensible to me, and I think it suggests that metadata revisions are separate from package versions. So maybe this is just a question for the package index?

> @f-f: I am not quite comfortable with Hackage's approach of storing package sources and metadata revisions separately, as I think it has a couple of problems:
[...]

PR #140, opened 2021-03-08
Implement API pipeline

### LINKED ISSUE #3: Package name constraints

### LINKED ISSUE #16: Support packages with sources not in `src`

### LINKED ISSUE #23: Implement CI checks pipeline
There are a bunch of things that we need to enforce in CI, so that we don't add packages that don't follow the criterias we defined.

Checks we want to run for every package:
- [ ] verify that the package `name` satisfies the constraints (as defined in #3)
- [ ] verify that the `repo` value is the same for all the versions of a package (to prevent packages "stealing" package names)
- [ ] verify that the `license` is a valid SPDX expression (see #13)
- [ ] verify that the `repo` is on GitHub (see #15)
- [x] ~reject packages with ES6 syntax in their FFI (see #24)~
- [ ] ensure that a package has a `lib` Target
- [ ] ensure that a package being added has sources: `["src/**/*.purs"]` in its `lib` Target (see [here](https://github.com/purescript/registry/issues/16#issuecomment-699712095))
- [ ] ensure that no subdir is present to new packages until we are confident that downstream tooling can deal with it (see [here](https://github.com/purescript/registry/issues/16#issuecomment-699712095))

This issue is about:
- tracking that all these checks are implemented
- writing them down in the spec, so that clients can implement them too

### LINKED ISSUE #96: Implement package addition/upgrade
We'd need to implement the CI code that will take care of adding packages to the registry, something like [the flow shown here](https://github.com/purescript/registry/issues/84#issuecomment-712407673)

f-f:
## What does this do?

As per title, this PR implements and deploys a first working version of the main CI pipeline that implements the backend of the Registry API.

Rough summary of what happens in the pipeline:
- someone puts a JSON value in an issue
- a GitHub action kicks off and tries to parse that
- if it conforms to an [API Operation](https://github.com/purescript/registry/blob/master/v1/Operation.dhall) then it starts the publishing pipeline, which:
  - grabs the package from its location
  - if necessary converts it from Bower format
  - runs a bunch of checks on the package (see #23)
  - repackages everything and uploads on our official package backend (which is [up and running!](http://packages.purescript.org/))

## Things not in scope for this PR

- we don't build packages at all (yet at least, we might in the future, see #154), just upload things
- we still need to import all existing packages from the Bower registry (see #95)

## Can I publish packages after we merge this?

No. Because:
- all the packages in the Bower registry need to be uploaded first
- and we need some other improvements to the `Manifest` schema before that (e.g. #124)

## How to review this

- all the `TODO`s left in the code are not supposed to be tackled now, they are either not urgent to get this off the ground, or just placeholders for future code. Some things we are not implementing now:
  - only `Addition` and `Upgrade` operations are implement, no `Unpublish` yet (we'll likely need to tackle #155 first)
  - while we intend to support "any `git` thingy" to pull packages from, for now we only support GitHub
  - support for package sets is not in yet, see #156
  - while we'd like to upload docs to Pursuit too, there are a few blockers before we get there, see #154
- the main file you'll want to look at is `ci/src/Registry/API.purs`, everything else is built to support that
- feel free to ignore the `BowerImport` module, I'll heavily refactor that in the next PR anyways
- refactoring suggestions and improvements on the Registry design are most welcome, while things along the lines of "add more code" are less welcome (the PR is already giant, we just want to get this merged really)

## Ticket bureaucracy

Fixes #96, fixes #3, fixes #16, fixes #23


ci/src/BowerImport.purs
```diff
@@ -69,10 +80,10 @@ main = Aff.launchAff_ do
       -- This is necessary so that we can do the "self-containment" check later.
       -- We keep a temporary cache on disk, so that it's easier to do development
       -- without consuming the GitHub request limit.
-      let allPackages = bowerPackages <> newPackages
+      let (SemigroupMap allPackages) = SemigroupMap bowerPackages <> SemigroupMap newPackages
```
kritzcreek, line 64: I'd just do an explicit `Map.union` here. What about conflicts? Should packages in `bowerPackages` shadow the ones in `newPackages` or should we be reporting an error instead?

ci/src/BowerImport.purs
```diff
@@ -69,10 +80,10 @@ main = Aff.launchAff_ do
...
       releaseIndex <- Map.fromFoldable <$> forWithIndex allPackages \nameWithPrefix repoUrl -> do
-        let name = stripPurescriptPrefix nameWithPrefix
-        let address = unsafePartial $ fromRight $ GitHub.parseRepo repoUrl
+        let name = stripPureScriptPrefix nameWithPrefix
+        let address = fromRight' (\_ -> unsafeCrashWith "Unexpected Left") $ GitHub.parseRepo repoUrl
```
kritzcreek:
```suggestion
        let address = fromRight' (\_ -> unsafeCrashWith $ "Failed to parse the repo url: " <> show repoUrl) $ GitHub.parseRepo repoUrl
```

ci/src/BowerImport.purs
```diff
@@ -56,10 +64,13 @@ main = Aff.launchAff_ do
...
+
+  let
+    parseJsonMap str =
+      Json.jsonParser str
+        >>= (Json.decodeJson >>> lmap Json.printJsonDecodeError)
         >>> map (Map.fromFoldableWithIndex :: Foreign.Object String -> Map String String)
+
   case parseJsonMap bowerPackagesStr, parseJsonMap newPackagesStr of
     Left err, _ -> error $ "Error: couldn't parse bower-packages.json, error: " <> err
     _, Left err -> error $ "Error: couldn't parse new-packages.json, error: " <> err
```
kritzcreek, line 57: Could we have an error case for both failing first? Otherwise you might end up playing ping-pong with this check, where you fix the first only to come back to having to fix the second.

ci/src/BowerImport.purs
```diff
@@ -108,79 +120,86 @@ main = Aff.launchAff_ do
...
+                 ]
             else do
               -- now we should be ready to convert it
               let manifestPath = packageFolder <> "/" <> release.name <> ".json"
-              let manifestStr = stringifyWithIndent 2 $ Json.encodeJson $ toManifest bowerfile release.name address
+              let manifestStr
+                    = stringifyWithIndent 2
+                    $ Json.encodeJson
+                    $ toManifest bowerfile release.name
+                    $ GitHub { owner: address.owner, repo: address.repo, subdir: Nothing }
```
kritzcreek:
```suggestion
              let manifestStr =
                    stringifyWithIndent 2
                      $ Json.encodeJson
                      $ toManifest bowerfile release.name
                      $ GitHub { owner: address.owner, repo: address.repo, subdir: Nothing }
```

More consistent with the rest of this file.

ci/src/PackageUpload.purs
```diff
@@ -20,27 +25,35 @@ type Path = String
 upload :: PackageInfo -> Path -> Effect Unit
 upload { name, version, revision } path = Aff.launchAff_ $ do
   -- first check that the file path exists
-  fileContent <- do
-    fileExists <- FS.exists path
-    if fileExists
-    then FS.readFile path
-    else Aff.throwError $ Aff.error $ "File doesn't exist: " <> show path
+  fileContent <-
+    FS.exists path >>= if _ then
```
kritzcreek: In general you shouldn't "check for existence" before performing a file operation. This way you're making yourself vulnerable to race conditions where the file system changes between the check and the operation. Just run the file operation and handle it potentially failing instead.

kritzcreek (commented):
Just some things I spotted. This looks great in general, I'm excited!

f-f:
@kritzcreek thanks for the comments! :heart:
The code is still unstable and we're still shuffling things around a lot, so we'll get to your suggestions once things have stabilized a bit (also note: some of the code might just disappear because it's not needed anymore)
We'll mark the PR as ready for review once we get there, to signal that it's open for general feedback.

ci/src/PackageUpload.purs
f-f: Sounds good, thanks for spotting this!

ci/src/BowerImport.purs
f-f: [CI checks that the two are not overlapping](https://github.com/purescript/registry/blob/master/verify-registry-unique.sh)

ci/src/BowerImport.purs
f-f: [CI checks that the two files contain sensible JSON](https://github.com/purescript/registry/blob/master/verify-registry-unique.sh), so in principle we shouldn't even need to check for these `Left`s

.github/workflows/API.yaml
```diff
@@ -7,14 +7,44 @@ on:
...
+          path: |
+            .spago
+            output
+
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=channel:nixos-20.09
+
+      - name: "Run API"
+        run: cd ci && npm ci && nix-shell ../default.nix --run "spago run -m Registry.API"
```
thomashoneyman:
```suggestion
        run: cd ci && npm ci && nix-shell ../default.nix --run "spago run -m Registry.API"

```

.github/workflows/main.yml
```diff
@@ -1,25 +1,60 @@
...
+          key: ${{ runner.os }}-spago-${{ hashFiles('./ci/**/*.dhall') }}
+          path: |
+            .spago
+            output
+
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=channel:nixos-20.09
+
+      # Pro-tempore checks, should go away once new registry is in place
```
thomashoneyman, line 44: We'll need to re-enable these and remove the `TestPush` script.

ci/src/FFI/Dhall.purs
```diff
@@ -1,19 +1,16 @@
 module Dhall where
```
thomashoneyman: Can we match module names with locations in the file system? For example, this would be `FFI.Dhall`. Or we could use a `Foreign` directory, so `Foreign.Dhall` as the module name.

If we want a top-level Dhall name then we could just float it out to the top level.

ci/src/FFI/GitHub.purs
```diff
@@ -0,0 +1,75 @@
+module GitHub where
```
thomashoneyman: This module defines some types we might not want to be considered "FFI", like the `Event` type. Perhaps we could float the `GitHub` module out to the top level, include only PureScript-safe code, and move the FFI implementations into a `FFI.GitHub` module?

ci/src/FFI/GitHub.purs
```diff
@@ -0,0 +1,75 @@
...
+
+foreign import createCommentImpl :: EffectFn3 Address Int String (Promise Unit)
+
+createComment :: IssueNumber -> String -> Aff Unit
+createComment issue body = Promise.toAffE $ runEffectFn3 createCommentImpl registryAddress (coerce issue) body
+
+foreign import closeIssueImpl :: EffectFn2 Address Int (Promise Unit)
+
+closeIssue :: IssueNumber -> Aff Unit
+closeIssue issue = Promise.toAffE $ runEffectFn2 closeIssueImpl registryAddress (coerce issue)
```
thomashoneyman:
```suggestion
closeIssue issue = Promise.toAffE $ runEffectFn2 closeIssueImpl registryAddress (coerce issue)

```

ci/src/FFI/SemVer.purs
```diff
@@ -0,0 +1,95 @@
...
+
+instance encodeJsonRange :: EncodeJson Range where
+  encodeJson = encodeJson <<< printRange
+
+foreign import parseRangeImpl :: String -> Nullable String
+parseRange :: String -> Maybe Range
+parseRange = map Range <<< toMaybe <<< parseRangeImpl
+
+printRange :: Range -> String
+printRange (Range r) = r
```
thomashoneyman:
```suggestion
printRange (Range r) = r

```

ci/src/FFI/Tmp.js
```diff
@@ -0,0 +1,13 @@
...
+
+exports.mkTmpDirImpl = function() {
+  return function() {
+    const tmpobj = tmp.dirSync();
+    return tmpobj.name;
+  };
+};
+
+
+// TODO: tmpobj.removeCallback();
```
thomashoneyman: TODO?

ci/src/FFI/Tmp.purs
```diff
@@ -0,0 +1,8 @@
+module Tmp (mkTmpDir) where
+
+import Data.Function.Uncurried (Fn0, runFn0)
+import Effect (Effect)
+
+foreign import mkTmpDirImpl :: Fn0 (Effect String)
+mkTmpDir :: Effect String
+mkTmpDir = runFn0 mkTmpDirImpl
```
thomashoneyman:
```suggestion
mkTmpDir = runFn0 mkTmpDirImpl

```

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+import Text.Parsing.StringParser as Parser
+import Tmp as Tmp
+
+main :: Effect Unit
+main = launchAff_ $ do
+  eventPath <- liftEffect $ Env.lookupEnv "GITHUB_EVENT_PATH"
+  packagesMetadata <- do
+    whenM (not <$> FS.exists metadataDir) do
+      FS.mkdir metadataDir
+    packageList <- FS.readdir metadataDir
```
thomashoneyman: Should we be catching exceptions on things like `readdir` here rather than the `exists` check, as @kritzcreek suggested? For example, we could catch the exception by creating the directory and then move forward with an empty package list.

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+metadataFile packageName = metadataDir <> "/" <> PackageName.print packageName <> ".json"
+
+addOrUpdate :: { fromBower :: Boolean, ref :: String, packageName :: PackageName } -> Metadata -> RegistryM Unit
+addOrUpdate { ref, fromBower, packageName } metadata = do
+  -- let's get a temp folder to do our stuffs
+  tmpDir <- liftEffect $ Tmp.mkTmpDir
+  -- fetch the repo and put it in the tempdir, returning the name of its toplevel dir
+  folderName <- case metadata.location of
+    Git _ -> do
+      -- TODO: remember subdir whenever we implement this
```
```suggestion
      -- TODO: Support non-GitHub packages. Remember subdir when implementing this. (See #15)
```

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+      throwWithComment "Packages are only allowed to come from GitHub for now. See #15"
+    GitHub { owner, repo, subdir } -> do
+      -- Check: subdir should not be there
```
```suggestion
      -- TODO: Support subdir. In the meantime, we verify subdir is not present. (See #16)
```

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+        throwWithComment $ "Error while reading Bowerfile: " <> err
+      Right bowerfile -> case Bower.toManifest bowerfile ref metadata.location of
+        Left err ->
+          throwWithComment $ "Unable to convert Bowerfile to a manifest: " <> err
+        Right manifest -> do
+          let manifestStr = Json.stringifyWithIndent 2 $ Json.encodeJson manifest
+          liftAff $ FS.writeTextFile UTF8 manifestPath manifestStr
+
+  -- Try to read the manifest, typechecking it
+  manifestExists <- liftAff $ FS.exists manifestPath
```
thomashoneyman: Should we just read the text file here, catching the exception by throwing the "Manifest not found" error?

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+    else do
+      manifestStr <- liftAff $ FS.readTextFile UTF8 manifestPath
+      liftAff (Dhall.jsonToDhallManifest manifestStr) >>= case _ of
+        Left err ->
+          throwWithComment $ "Could not type-check Manifest file: " <> err
+        Right _ -> case fromJson manifestStr of
+          Left err -> throwWithComment $ "Could not convert Manifest to JSON: " <> err
+          Right res -> pure res
+
+  -- TODO: pull the maintainers list from the manifest into the metadata?
```
thomashoneyman: Is this TODO necessary for the first cut? Is there an issue that we can associate this TODO with?

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+  liftAff $ FS.rename absoluteFolderPath (tmpDir <> "/" <> newDirname)
+  let tarballPath = tmpDir <> "/" <> newDirname <> ".tar.gz"
+  liftEffect $ Tar.create { cwd: tmpDir, folderName: newDirname, archiveName: tarballPath }
+  log "Hashing the tarball..."
+  hash <- sha256sum tarballPath
+  log $ "Hash: " <> hash
+  log "Uploading package to the storage backend..."
+  let uploadPackageInfo = { name: packageName, version: newVersion }
+  uploadPackage uploadPackageInfo tarballPath
+  -- TODO: handle addToPackageSet
```
thomashoneyman: Should we create an issue about this and associate it with this TODO?

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+  let metadataFilePath = metadataFile packageName
+  liftAff $ FS.writeTextFile UTF8 metadataFilePath (Json.stringifyWithIndent 2 $ Json.encodeJson newMetadata)
+  updatePackagesMetadata manifest.name newMetadata
+  commitToTrunk packageName metadataFilePath >>= case _ of
+    Left err ->
+      comment "Package uploaded, but metadata not synced with the registry repository (cc: @purescript/packaging )"
+    Right _ -> do
+      comment "Package successfully uploaded to the registry! :tada: :rocket:"
+  closeIssue
+  -- TODO: upload docs to pursuit
```
thomashoneyman: Should we create an issue about this and associate it with this TODO?

I think there's some discussion to be had around automatic additions to the package sets and to documentation -- ie. the idea that it should be fully supported as a manual workflow, but if we can pull it off automatically that we should do so.

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+fromJson = Json.jsonParser >=> (lmap Json.printJsonDecodeError <<< Json.decodeJson)
+
+readJsonFile :: forall a. Json.DecodeJson a => String -> Aff a
+readJsonFile path = do
+  strResult <- FS.readTextFile UTF8 path
+  case fromJson strResult of
+    Left err -> Aff.throwError $ Aff.error $ "Error while parsing json from " <> path <> " : " <> err
+    Right r -> pure r
+
+mkNewMetadata :: Repo -> Metadata
```
thomashoneyman: Should this be part of the `RegistrySchema` module?

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+mkNewMetadata location = { location, releases: mempty, unpublished: mempty, maintainers: mempty }
+
+addVersionToMetadata :: SemVer -> VersionMetadata -> Metadata -> Metadata
```
thomashoneyman: Should this be part of the `RegistrySchema` module?

ci/src/Registry/RegistryM.purs
```diff
@@ -0,0 +1,108 @@
...
+
+  where
+    runGit args = ExceptT do
+      result <- Process.spawn { cmd: "git", args, stdin: Nothing } NodeProcess.defaultSpawnOptions
+      case result.exit of
+        NodeProcess.Normally 0 -> do
+          info result.stdout
+          info result.stderr
+          pure $ Right unit
+        _ -> pure $ Left result.stderr
```
thomashoneyman: Should this just go in the `API` module? It doesn't seem specific to `RegistryM`, and other utilities like `wget` (which _is_ in `RegistryM`) are in the API module.

ci/src/Registry/TestPush.purs
```diff
@@ -0,0 +1,59 @@
+module TestPush where
```
thomashoneyman: Should we move modules like this into a dedicated `Scripts` namespace, or even to a different directory? I didn't realize that several of the modules under the `Registry` namespace / directory are actually scripts and not part of the pipeline itself.

```purs
-- ci/src/Scripts/TestPush
module Scripts.TestPush where
```

This way the `Registry` directory is reserved for Registry-specific types and functions, as well as the main pipeline (`API.purs`).

ci/src/Registry/TestPush.purs
```diff
@@ -0,0 +1,59 @@
...
+import Text.Parsing.StringParser as Parser
+
+issueNumber :: IssueNumber
+issueNumber = IssueNumber 149
+
+metadataDir :: FilePath
+metadataDir = "../metadata"
+
+metadataFile :: PackageName -> FilePath
+metadataFile packageName = metadataDir <> "/" <> PackageName.print packageName <> ".json"
```
thomashoneyman: These are the same as the values in `API`; should we just import them here?

ci/test/Main.purs
```diff
@@ -0,0 +1,142 @@
...
+  parseSemVer "v1.2.3"
+  parseSemVer "1.2.3-rc2"
+  parseSemVer "0.1.2+r2"
+
+  let parseRange range expected = Spec.it ("Parse Range " <> show range <> " into " <> show expected) do
+        (SemVer.printRange <$> SemVer.parseRange range) `Assert.shouldSatisfy` case _ of
+          Just parsed -> parsed == expected
+          Nothing -> false
+
+  parseRange "^1.3.4" ">=1.3.4 <2.0.0-0"
```
thomashoneyman:
```suggestion
  parseRange "^1.3.4" ">=1.3.4 <2.0.0-0"

```

default.nix
```diff
@@ -1,22 +1,12 @@
 { pkgs ? import <nixpkgs> {} }:
```
thomashoneyman, line 1: Should we pin this to 20.09, as we did in package-sets?

v1/Metadata.dhall
```diff
@@ -9,16 +9,14 @@ let Map = (./Prelude.dhall).Map.Type
...
-let Release = { ref : Text, hash : Text }
+let VersionMetadata = { ref : Text, hash : Text }
 
 in
   {
   -- The pointer to where the code lives
   , location : Repo
   -- A mapping between versions and info about a release
-  , releases : Map Version Release
+  , releases : Map Text VersionMetadata
```
thomashoneyman: Should this `Text` be a type synonym for a `SemVer`?

v1/Operation.dhall
```diff
@@ -7,7 +7,7 @@ A type describing all the possible operations for the Registry API.
 let Repo = ./Repo.dhall
 
 in
-  < Addition : { packageName : Text, newPackageLocation : Repo, newRef : Text, addToPackageSet: Bool }
-  | Update : { packageName : Text, updateRef : Text }
+  < Addition : { packageName : Text, newPackageLocation : Repo, newRef : Text, addToPackageSet: Bool, fromBower : Bool }
+  | Update : { packageName : Text, updateRef : Text, fromBower : Bool }
   | Unpublish : { packageName : Text, unpublishVersion : Text, unpublishReason: Text }
   >
```
thomashoneyman:
```suggestion
  >

```

v1/backends.dhall
```diff
@@ -20,7 +20,7 @@ let Package = { name : Text, version : Text }
 let Backend = Package -> Text
 
 let backends = toMap
-  { exampleBackend = \(package : Package) -> "https://example.com/purescriptRegistry/${package.name}/${package.version}.tar.gz"
+  { main = \(package : Package) -> "https://packages.purescript.org/${package.name}/${package.version}.tar.gz"
   }
 
 in backends : Map Text Backend
```
thomashoneyman:
```suggestion
in backends : Map Text Backend

```

ci/src/Registry/API.purs
```diff
@@ -0,0 +1,282 @@
...
+
+fromJson :: forall a. Json.DecodeJson a => String -> Either String a
+fromJson = Json.jsonParser >=> (lmap Json.printJsonDecodeError <<< Json.decodeJson)
+
+readJsonFile :: forall a. Json.DecodeJson a => String -> Aff a
+readJsonFile path = do
+  strResult <- FS.readTextFile UTF8 path
+  case fromJson strResult of
+    Left err -> Aff.throwError $ Aff.error $ "Error while parsing json from " <> path <> " : " <> err
+    Right r -> pure r
```
thomashoneyman: My gut sense on these is that they're a bit too magical; we're almost always going to want to provide some more elaborated error messages, right? In which case we'll want to do things like check the JSON parsing and then check decoding in order to throw the right error message.

thomashoneyman: In other words, we'll probably want to inline most of this to where these functions are being used so we can have more control over the errors.

f-f: We use `readJsonFile` in a single place, should we just inline it?

I think `fromJson` is fine, since it doesn't swallow any errors and it's just a helper to reduce verbosity

v1/Metadata.dhall
f-f: Yes!

default.nix
f-f: Great idea

ci/src/Registry/RegistryM.purs
f-f: That was my initial call as well, but then we'd have a circular import, unless we inject the implementation in `mkEnv`, but that doesn't feel right.
A possibly better way to go at this would be to move the `mkEnv` to the `API` module, since it's specific to that execution environment anyways

ci/src/FFI/Tmp.js
f-f: Yeah I wanted to remove the temporary directory in case of failure, but we probably don't need to since this thing is going to run in CI. I'll turn this into a proper TODO

ci/src/FFI/GitHub.purs
f-f: I think that module fragmentation increases indirection and worsens maintainability - this module is already small and I wouldn't like to split it further.

As I mentioned in the call, I think `FFI` is a bad name. My goal was to separate the "business logic" of the Registry from the "general utilities". I went with "FFI" since most of such stuff is indeed coming from JS libraries, but indeed I'd like to find a better name for these modules. Any ideas?

ci/src/FFI/Dhall.purs
f-f: We can match modules with directories but we need a good name for the directory, see https://github.com/purescript/registry/pull/140#discussion_r602153995

ci/src/FFI/Tmp.js
f-f: Done

ci/src/Registry/API.purs
f-f: Done

ci/src/Registry/API.purs
f-f: Done

ci/src/Registry/TestPush.purs
f-f: We'll remove this module, but I moved the other script-like things to the `Scripts` subdirectory

ci/src/Registry/RegistrySchema.purs
```diff
@@ -0,0 +1,157 @@
...
+type Manifest =
+  { name :: PackageName
+  , version :: SemVer
+  , license :: License
+  , repository :: Repo
+  , targets :: Object Target
+  }
+
+type Target =
+  { dependencies :: Object Range
```
f-f: I am not sure about having `Range` here vs just `String`. The reason is that in this way we'll read in the SemVer range and automatically convert it to a "simple range" before this data touches the rest of the app. This means that we are not able to warn the user about "hey you should have used this simple range but you didn't", and/or fail if the range is not simple. I guess a good way to go at this could be to store also the "original" range into the `Range` type, so that we can keep this information. Does this make sense?

ci/src/Registry/API.purs
f-f: Inlined `readJsonFile`

ci/src/Registry/API.purs
f-f: Done

ci/src/Registry/API.purs
f-f: Done

.github/workflows/main.yml
f-f: Done

ci/src/FFI/GitHub.purs
f-f: Renamed to `Foreign`

ci/src/Registry/API.purs
f-f: Opened #154

ci/src/Registry/RegistrySchema.purs
f-f: We went with storing the original string too, so we can eventually error out

ci/src/Registry/API.purs
f-f: Opened #156, but needs to be populated with more info

ci/src/Registry/API.purs
f-f: Opened #155

f-f:
I just realized we should forbid here the use of build metadata in the package version, as we agreed in #80

PR #160, opened 2021-04-15
Adds purescript-event and purescript-behaviors

mikesol:
Making FRP uncool again...

### DIFF
--- a/new-packages.json
+++ b/new-packages.json
@@ -115,5 +115,7 @@
   "purescript-heterogeneous-extrablatt": "https://github.com/sigma-andex/purescript-heterogeneous-extrablatt.git",
   "purescript-pretty-logs": "https://github.com/PureFunctor/purescript-pretty-logs.git",
   "purescript-web-resize-observer": "https://github.com/nsaunders/purescript-web-resize-observer.git",
-  "purescript-homogeneous": "https://github.com/paluh/purescript-homogeneous.git"
+  "purescript-homogeneous": "https://github.com/paluh/purescript-homogeneous.git",
+  "purescript-event": "https://github.com/mikesol/purescript-event.git",
+  "purescript-behaviors": "https://github.com/mikesol/purescript-behaviors.git"
 }


f-f:
It looks like there are already some libraries registered with these names, could you pick new ones?

srghma:
> could you pick new ones

that's the reason why I have created this org https://github.com/purescript-open-community/

we can move these packages there

mikesol:
> It looks like there are already some libraries registered with these names, could you pick new ones?

Definitely doable, although I can also ask the original author if he'll transfer the archived repos to my account. If that's the case, would it be possible to keep the original name? This is to avoid having two libraries that do the same thing - one that works pre-0.14 (the original) and one 0.14 onwards (the fork).

f-f:
> Definitely doable, although I can also ask the original author if he'll transfer the archived repos to my account. If that's the case, would it be possible to keep the original name?

Yes, in that case it would be possible to keep the original names!

thomashoneyman:
@mikesol We've already discussed the situation with the original author and they'd like to keep the archived repos under their account (though they're fine with them being forked under new names).

mikesol:
Got it, makes sense!

I'm wondering what the best move is from a community perspective? Is there a naming convention you feel would adequately communicate that the new package is forked from the old one and is not, at least in the short term, intended to be any different other than versioning updates (ie purescript-event-fork1 or purescript-event-post0.14)?

Mike Solomon
C.E.O. @Meeshkan
https://www.meeshkan.com

Sent via Superhuman ( ***@***.*** )

mikesol:
Btw Serhii I think the project looks cool! In this case I'm not sure it's a good fit as it seems like these packages will gel with the registry after we sort out the details, but I will definitely keep it in mind for future projects.

Mike Solomon
C.E.O. @Meeshkan
https://www.meeshkan.com

Sent via Superhuman ( ***@***.*** )

thomashoneyman:
I think the approach taken by @srghma in creating https://github.com/purescript-open-community makes it clear packages are being maintained despite being archived by their original authors.

PR #170, opened 2021-05-16
typography: curly quotation marks, em dash

toastal:
The purpose of this commit is to make the documentation look nicer and could also be seen as more ‘correct’ since it _is_ the suggested punctuation in most every English style guide. The downside is ANSI compatibility, but I’m pretty sure most everyone is using at least UTF-8 everywhere and these symbols are ASCII extended (codes 128-255) compatible as well.

Technically em dashes, are not surrounded by spaces, but in practice it’s common to see (like in newspapers) and in a monospace setting it helps readability.

- - -

Vim diagraphs (`ctrl+k`)

```
–       -N      2013    8211    EN DASH 
—       -M      2014    8212    EM DASH
‘       '6      2018    8216    LEFT SINGLE QUOTATION MARK
’       '9      2019    8217    RIGHT SINGLE QUOTATION MARK
“       "6      201C    8220    LEFT DOUBLE QUOTATION MARK
”       "9      201D    8221    RIGHT DOUBLE QUOTATION MARK
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,32 +6,32 @@ PureScript needs a way to distribute packages. We used to rely on the Bower regi
 
 ## Goals for a PureScript package registry
 
-Here's a non-comprehensive list of desiderable properties/goals that we'd generally like to
+Here’s a non-comprehensive list of desiderable properties/goals that we’d generally like to
 achieve when talking of a PureScript registry, and that we think this design covers:
-- **independent**: we're coming from a situation when relying on a third party registry has
-  faulted us, so we'd need something that we can control and possibly host ourselves.
-- **immutable**: published packages are immutable - once a version has been published
+- **independent**: we’re coming from a situation when relying on a third party registry has
+  faulted us, so we’d need something that we can control and possibly host ourselves.
+- **immutable**: published packages are immutable — once a version has been published
   then its source code is forever packaged in the tarball uploaded to our storage(s).
   The only exception is that **unpublishing** will be possible for some time after publishing.
   This goal also directly stems from our experience with Bower, where we were not able to
   prevent packages from disappearing and/or being altered, since we were not storing the source
   anywhere, but just pointing to the original location of the source, that we had no control over.  
-  This means that with this Registry if your package is building today then you're guaranteed that
-  the packages you're depending on will not disappear.
-- **with content hashes**, so that we're able to support independent storage backends
-  and be sure that the content they serve is the same, and - perhaps most importantly - that
+  This means that with this Registry if your package is building today then you’re guaranteed that
+  the packages you’re depending on will not disappear.
+- **with content hashes**, so that we’re able to support independent storage backends
+  and be sure that the content they serve is the same, and — perhaps most importantly — that
   its integrity is preserved over time.
 - **with version bounds for packages**, so that package authors can control the amount
-  of support that they want to provide for their packages - i.e. allowing a wider range of
+  of support that they want to provide for their packages — i.e. allowing a wider range of
   versions for upstream dependencies means more support.
 - **with a way for trusted editors to publish new versions**, so that under specified conditions
   faulty bounds/dependencies can be timely corrected if package authors are not around.
 - **ease of publishing a release** is optimized for and entirely automated.
 - package manifests are **declarative**: authors need not to concern themselves with
-  the "how", but just declare properties about their packages. E.g. there's no such
-  thing as NPM's `postinstall` and other similar hooks.
+  the “how” but just declare properties about their packages. E.g. there’s no such
+  thing as NPM’s `postinstall` and other similar hooks.
 - **no webserver**: all the software running the Registry is designed in such a way that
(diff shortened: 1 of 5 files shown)


wclr:
I would also propose using drop-caps and emojis wherever it is possible. 😉

toastal:
> I would also propose using drop-caps and emojis wherever it is possible. wink

If it were to improve readability, like LaTeX for math or `×` for multiplication in a documentation setting, I would agree with the sentiment. `gcc` uses curly quotation marks in there errors, all GPL-related and Mozilla licenses use these symbols, and many editors do “smart quotes” for the user. There is a precedence for this outside of code.

It _is_ a preference but it isn’t purely stylistic; I think it does show professionalism and an attention to details to use the correct punctuation in documentation.

wclr:
> gcc uses curly quotation marks in there errors, all GPL-related and Mozilla licenses use these symbols,

Not sure if it a sign of a better approach to writing technical documentation.

natefaubion:
> It is a preference but it isn’t purely stylistic; I think it does show professionalism and an attention to details to use the correct punctuation in documentation.

Without someway to enforce this automatically, I can guarantee you that this will not survive maintenance over time. I understand the sentiment, but you are essentially saying "here, now it's your job to make sure this is taken care of in the future" which is unlikely to happen.

toastal:
Contact me and I will gladly keep it in sync haha :sweat_smile:. 

Personally, I'd be happy to see even partial support. Since this project _feels_ more like a manifest, something more official and less mutable, it seems a better match (although I can agree that likely non-ANSI would likely degrade over time—even if editors like Vim make it easy with digraphs and OSs like Mac have it behind the option key (which is also why I type a lot of things through `nvim` so I can have access to the extended charsets), also most people never learn or care about typography and they may never notice or care).

wclr:
> also most people never learn or care about typography and they may never notice or care)

I would care about such things in magazine articles and books, to technical docs, where monospace font-face is usually used (at least to create them), I refer almost as to the code. And in the code, for example, I don't like to see Unicode characters where they are not needed really. When people use them to replace straight `->` or `::`, the only thing I do prefer to do though is to use  `∀` as `forall` keyword replacement because it makes code really cleaner. So I'm not saying that it is a matter of taste but of practicality and simplicity.

I may be missing something but, apart from the obvious need to automate such styling, I really don't see the reason why in such tech docs it is needed to use "correct" punctuation symbols from English style guides.

afcondon:
> So I'm not saying that it is a matter of taste but of practicality and simplicity.

Since practicality and simplicity are very much in the eye of the beholder i think they are just as much a matter of taste as aesthetics.

toastal:
I’m pretty sure Pandoc and GitBook automatically switches some symbols inferring context when published from `.md` files (like _not_ doing it in code blocks). If a project manager or tech writer were writing this in Word or on WordPress it would just be the default, but as coders, we tend to shy away from the non-ANSI (which has caused some issues with `$` sigils and say a UK keyboard with its `£` key (`Pd` Vim digraph I had to look up)). If this were the end-result of the say the `README`—to be transformed to markup for consumption on say the package site(?)—then this is a non-issue (closable); if it’s supposed to be read straight from GitHub or via plaintext then that where I think this maintains a stronger argument.

If these ASCII-extended symbols are perceived as a barrier to contribution or maintenance, my feelings won’t be hurt to have it `WONTFIX`’d. It is also not worth spending a lot of a time bike-shedding if there is no perceived value in typography for this situation. It’s just a suggestion, that could set a precedence for PureScript’s documentation as a result, but 🤷. 

- - -

Preference-wise I _love_ that PureScript has built-in `→`, `∷`, `∀` and Dhall’s `⩓`, `⫽`, `λ` and, after I got used to it, I tend to gravitate to these languages.

toastal:
Interesting aside: 
```
``double quoted'', `single quoted' -- ... -> <- => <=
```
 in [AsciiDoc](https://asciidoc.org/) is rendered as 

```
“double quoted”, ‘single quoted’ — …​ → ← ⇒ ⇐
```

 and is supported by GitHub.

f-f:
@toastal I am personally a fan of good typography and of Unicode symbols in modern languages, so I appreciete this!
However, I personally wouldn't like to have pretty formatting for this README. There are multiple reasons for this:
- the idea behind this was to eventually conform to IETF's RFC format, which is traditionally ASCII
- unfortunately, even if I'd really like it to be otherwise, not the whole world is on Unicode yet so this might lower accessibility
- without a CI check this is really doomed to drift away from the nicer typography and back to ASCII

toastal:
@f-f It _is_ technically still ASCII, but I'm splitting hairs. In this case, I would consider keeping an open mind about AsciiDoc over Markdown as it could solve other limitation issues down the line with like good internal linking, ASCII/Unicode symbols, outlines, fancy tables, admonitions, etc.

Close if you see fit.

f-f:
Thanks for the pointer @toastal!
I'll close this for now then

PR #173, opened 2021-05-22
Add description field to manifest

thomashoneyman:
Starts fixing #124 by adding a description field.

Do we want to put any validation in place for description lengths? Tweet-length, perhaps?


f-f:
Thank you for this @thomashoneyman!

> Do we want to put any validation in place for description lengths? Tweet-length, perhaps?

Yeah I think that's a good call.

Note that we'd also need to update the README, as we embed the `Manifest` type there

f-f:
Superseded by #282

PR #180, opened 2021-06-01
Add array-search, assert-multiple, huffman, maps-eager, repr and roman

JoelLefkowitz:


f-f:
@JoelLefkowitz it looks like you didn't use `pulp` to publish these - so they are missing a `bower.json`. Could you add one?

JoelLefkowitz:
I've been using spago as a package manager. I can copy the dependencies over to a bower.json but when I try using bower to install the dependencies and then pulp as a build tool  my build errors:

`Module Prim.Coerce was not found`

I think this is because the version of psc that pulp uses is old.

f-f:
Yeah it looks like you might have several `purs` executables in your path?

f-f:
@JoelLefkowitz any update on this?

JoelLefkowitz:
Fixing now, sorry for the delay

JoelLefkowitz:
Should be good to go now

f-f:
Great, thank you!

f-f (approved):

PR #191, opened 2021-08-06
Remove unreachable repositories from the Bower packages

f-f:
Some of these went down since we first fetched the list of packages from Bower.
Since we'd need all the packages to be reachable (to fetch their Bowerfiles), this PR removes them.

I'm slightly concerned about the fact that removing registered names is not something we generally allow (for security reasons), so I'm thinking that maybe we should keep these names and point them to a fake repo we control.


thomashoneyman:
@f-f I'm not sure I understand in what situations a repository becomes "unreachable." Does this mean that it's been pulled from the Bower registry?

Removing registered names is not something we generally allow, but I'm also open to that rule applying "from the point we start the registry" -- none of these packages see much usage.

f-f:
> I'm not sure I understand in what situations a repository becomes "unreachable."

It means that the repo address returns a 404, so the author deleted the account, or took down the repo. If a repo gets renamed then GitHub sets up redirects, but this is not the case.

> Removing registered names is not something we generally allow, but I'm also open to that rule applying "from the point we start the registry" -- none of these packages see much usage

Yeah, I have the same stance, but my worry is that these names could be recycled from the moment we remove them from here. It's true that these names are not used anywhere right now (i.e. it's only for our bookkeeping) and Bower builds including them would not actually build at all. So I guess I just got convinced that this is fine. I'll merge :smile:

PR #194, opened 2021-08-09
Add support for reading/writing a registry index

colinwahl:
The PR implements support for reading/writing a registry index: a GitHub repo that serves as a cache of the manifests for all packages at all published versions in the registry.

Right now the registry index is an empty GitHub repository:
https://github.com/purescript/registry-index

We have implemented the same directory structure as the Cargo package manager, as described in the Cargo book:
https://doc.rust-lang.org/cargo/reference/registries.html and demonstrated in the Crates index: https://github.com/rust-lang/crates.io-index. This includes using the same tree data structure and jsonl representation for each package.


ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+      fileName <- Array.last $ String.split (Pattern "/") path
+      guard (not (Array.elem fileName exclude))
+      pure $ PackageName fileName
+
+    packages = Array.mapMaybe goPath packagePaths
+
+  parsed <- for packages \package -> Tuple package <$> readPackage package
+  let
+    parsedPackages :: Array (Tuple PackageName (NonEmptyArray Manifest))
+    parsedPackages = Array.mapMaybe (\(Tuple package mbManifests) -> Tuple package <$> mbManifests) parsed
```
thomashoneyman: Here we're ignoring files that don't parse, but possibly we should throw an exception so that it's easier for us to find out when (for some should-not-be-possible reason) a manifest we've written doesn't parse.

ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+  or have a JSONLines (.jsonl) file for every package and a line for every version in there
+
+  Produce the directory containing this package in the registry index,
+  according to the following format:
+
+    * Packages with 1 character names are placed in a directory named 1.
+    * Packages with 2 character names are placed in a directory named 2.
+    * Packages with 3 character names are placed in the directory 3/{first-character} where {first-character} is the first character of the package name.
+    * All other packages are stored in directories named {first-two}/{second-two} where the top directory is the first two characters of the package name, and the next subdirectory is the third and fourth characters of the package name. For example, prelude would be stored in a file named pr/el/prelude.
+-}
```
thomashoneyman: Mind if we rewrite this comment to use `-- |` so it shows up in IDE tooling?

```suggestion
-- | Produce the directory containing this package in the registry index, using the following format:
-- |   * Packages with 1 character names are placed in a directory named 1.
-- |   * Packages with 2 character names are placed in a directory named 2.
-- |   * Packages with 3 character names are placed in the directory 3/{first-character} where {first-character} is the first character of the package name.
-- |   * All other packages are stored in directories named {first-two}/{second-two} where the top directory is the first two characters of the package name, and the next subdirectory is the third and fourth characters of the package name. For example, prelude would be stored in a file named pr/el/prelude.
-- |   * Each package file is a JSON Lines file where each line is a package manifest, stored in sorted order ascending by version.
-- |
-- | Format follows that used by Cargo in crates.io: https://github.com/rust-lang/crates.io-index
-- | As documented in the Cargo book: https://doc.rust-lang.org/cargo/reference/registries.html#index-format
```

ci/src/Registry/Index.purs
thomashoneyman: Thoughts, @f-f?

ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+import Partial.Unsafe (unsafeCrashWith)
+import Registry.PackageName (PackageName(..))
+import Registry.Schema (Manifest)
+
+type RegistryIndex = Map PackageName (Map SemVer Manifest)
+
+-- This function must be run from the root of the registry index.
+-- NOTE: Right now, this assumes that manifest files will parse
+readRegistryIndex :: Aff RegistryIndex
+readRegistryIndex = do
```
f-f: Can we pass a path instead of using the current directory?

```suggestion
readRegistryIndex :: String -> Aff RegistryIndex
readRegistryIndex path = do
```

ci/src/Registry/Index.purs
f-f: Totally. Also I'd go as far as including a CI check that validates all the files against the `Manifest` type, like we [do here already](https://github.com/purescript/registry/blob/master/ci/scripts/verify-dhall.sh). We don't have to include this CI check in this PR though (especially because the index is not there yet)

ci/src/Registry/PackageName.purs
```diff
@@ -1,5 +1,5 @@
 module Registry.PackageName
-  ( PackageName
+  ( PackageName(..)
```
f-f: I'd prefer if we wouldn't export this constructor if possible (i.e. one should always go through `parse`)
```suggestion
  ( PackageName
```

f-f (commented):
This looks good, great work! :clap: 

The main change that I'd like to see is that the various functions that right now assume to be called from the root of the `registry-index` repo should not do that and instead take a `path` parameter, so that we'd be able to call this from anywhere without the code being aware of the current location.

ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+  let
+    -- Exclude certain files that will always be in the root of the registry index.
+    -- These files do not correspond to a package manifest file.
+    exclude :: Array String
+    exclude = [ "config.json" ]
+
+    goPath path = do
+      fileName <- Array.last $ String.split (Pattern "/") path
+      guard (not (Array.elem fileName exclude))
+      pure $ PackageName fileName
```
thomashoneyman: We can use `PackageName.parse` here to keep it a smart constructor as @f-f mentioned

ci/src/Registry/Index.purs
thomashoneyman: Good idea! I'd say skip the CI check in this PR, @colinwahl, but we can switch to throwing an exception here.

ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+getIndexDir :: PackageName -> FilePath
+getIndexDir (PackageName packageName) = case String.length packageName of
+  0 -> unsafeCrashWith "Invalid PackageName"
+  1 -> "1/"
+  2 -> "2/"
+  3 -> "3/" <> String.take 1 packageName <> "/"
+  _ -> String.take 2 packageName <> "/" <> String.take 2 (String.drop 2 packageName) <> "/"
+
+getIndexPath :: PackageName -> FilePath
+getIndexPath (PackageName packageName) = getIndexDir (PackageName packageName) <> packageName
```
thomashoneyman: We can do this instead to keep the smart constructor:

```suggestion
getIndexPath packageName = getIndexDir packageName <> PackageName.print packageName
```

ci/src/Registry/Index.purs
```diff
@@ -0,0 +1,128 @@
...
+      jsonLines <- traverse parseJson $ String.split (Pattern "\n") contents
+      traverse decodeJson jsonLines
+
+  pure case contentsResult of
+    Left _ -> Nothing
+    Right Nothing -> Nothing
+    Right (Just arr) -> NEA.fromArray arr
+
+-- This function must be run from the root of the registry index.
+-- TODO: Should we bail on version collisions?
```
thomashoneyman: Thoughts on this, @f-f? This PR will always write the manifest, even if the manifest already exists in the index file at the given version. We could instead bail out and not write it, or we could throw an exception...I'm not sure if there's some danger to overwriting in the case of version collision.

ci/src/Registry/PackageName.purs
thomashoneyman: I agree.

ci/src/Registry/Index.purs
f-f: Version collisions should be prevented by the fact that we store them in a `Map`, but if we want to rule them out we can check if the Manifest we're about to write is the same as the one that's there, and error out if trying to overwrite.
Since the index is supposed to be just a cache (that you can throw away) I think it's fine to not worry aboout this too much.

thomashoneyman (approved):

thomashoneyman:
@colinwahl One CI failure:

```console
[1/1 UnknownImport] src/Registry/Scripts/BowerImport.purs:30:24

  30  import Registry.Index (RegistryIndex)
```

Also, you may end up with a formatting failure if you haven't formatted with `purs-tidy`.

colinwahl:
Thanks @thomashoneyman for addressing the CI failures :)

f-f (approved):
This looks great, thanks!

PR #196, opened 2021-08-09
Make some progress in parsing all Bowerfiles from the old registry

f-f:
In this PR we start turning the Bower-import machinery from a proof-of-concept that generates Manifests from some Bowerfiles to the "production" machine that will batch-import Bower packages into the Registry.
To do this we have to:
- introduce a more structured approach (a JSON file) in keeping track of what's wrong with the Bowerfiles that we try to convert to Registry `Manifest`s
- have more lenient coercions that come from trying to parse them in an accommodating way


thomashoneyman:
What should happen with https://github.com/purescript/registry/commit/ff1dd9797d2ffc11300bf9ee85180c3c89006696? Should I re-apply it on top of this once merged?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -51,44 +73,33 @@ main = Aff.launchAff_ do
...
       -- as first thing we iterate through all the packages and fetch all the
       -- releases from GitHub, to populate an in-memory "releases index".
       -- This is necessary so that we can do the "self-containment" check later.
       -- We keep a temporary cache on disk, so that it's easier to do development
       -- without consuming the GitHub request limit.
       let (SemigroupMap allPackages) = SemigroupMap bowerPackages <> SemigroupMap newPackages
-      releaseIndex <- Map.fromFoldable <$> forWithIndex allPackages \nameWithPrefix repoUrl -> do
+      releaseIndex :: ReleasesIndex <- withCache ("releaseIndex") (Just $ Hours 1.0) $ Map.fromFoldable <$> forWithIndex allPackages \nameWithPrefix repoUrl -> do
         let name = stripPureScriptPrefix nameWithPrefix
         let address = fromRight' (\_ -> unsafeCrashWith $ "Failed to parse the repo url: " <> show repoUrl) $ GitHub.parseRepo repoUrl
```
thomashoneyman, line 85: Why do we necessarily assume a Bower package is available on GitHub? We use this release index in several places (for example, when checking a package's dependencies are all in the registry by checking the release index). In your experience running this tool are all packages really on GitHub?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -51,44 +73,33 @@ main = Aff.launchAff_ do
...
-            examplePackage = Set.member name (Set.fromFoldable [ "aff", "mysql", "prelude" ])
-            shouldFetch = manifestIsMissing && not shouldSkip && examplePackage
-
-          when shouldFetch do
-            -- if yes, then..
-            log $ "Could not find manifest for version " <> release.name <> " of " <> show address <> ", making it..."
+      bowerRegistry <- forWithIndex releaseIndex \name { address, releases } -> do
+        -- some releases should be skipped because they have issues, see `toSkip`
+        let
+          (workingReleases :: Array GitHub.Tag) = Set.toUnfoldable $
```
thomashoneyman, line 98: Fetching a release simply means grabbing available GitHub tags. Should we parse these tags to make sure they're a valid SemVer tag before continuing? After all, I could have a tag on my project that says "whatever-prep" and not "v1.0.0".

f-f:
@thomashoneyman yeah, you can just apply it on top of trunk and start a new branch (or just rebase `bower-import`)

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -227,123 +250,45 @@ withCache path maybeDuration action = do
...
-  , "facebook" /\ "v0.2.0"
-  , "facebook" /\ "v0.2.1"
-  , "facebook" /\ "v0.2.2"
-  , "facebook" /\ "v0.2.3"
-  , "facebook" /\ "v0.3.0"
-  , "ide-purescript-core" /\ "v0.5.2"
-  , "var" /\ "v0.0.1"
-  , "tree-rose" /\ "v2.0.0"
-  ]
+toSkip = unsafePerformEffect $ do
```
thomashoneyman: Hey, we only use this inside an effectful block! I'm not sure how much we're gaining from this `unsafePerformEffect` vs. something like

```purs
main :: Effect Unit
main = do
  ...
  toSkip <- readSkipped

readSkipped :: Effect (Set (Tuple String String))
readSkipped = do ...
```

I'd only want to introduce an `unsafePerformEffect` if it was bringing a serious ergonomic gain.

ci/src/Registry/Scripts/BowerImport.purs
f-f: Yeah, I had this same curiosity at some point, and all the packages are indeed on GitHub!

ci/src/Registry/Scripts/BowerImport.purs
f-f: This happens further down the road, because converting to a manifest implies that these are parsed as `SemVer` (and if you notice in the JSON there's a section for versions that don't conform to that)

ci/src/Registry/Scripts/BowerImport.purs
f-f: Sure, I totally forgot why I put this here, but it looks like we can get rid of it :smile:

thomashoneyman:
Other than the `unsafePerformEffect` I think this is an improvement!

thomashoneyman (approved):

PR #200, opened 2021-08-13
Improve Bower import by collecting errors and manifests

thomashoneyman:
This PR overhauls the `BowerImport` script so that it produces three files:

- `bower-manifests` contains all manifests that have successfully been converted
- `bower-failures` is a diagnostics report for all package versions that fail, along with why they failed. I've successfully used this to implement a few more cleanup functions that preserve more packages.
- `bower-exclusions` contains package versions that are missing bowerfiles altogether (in the future: for now, it has a few more than that).

You can run this for yourself with:

```
GITHUB_TOKEN=<my-token> spago run -m Registry.Scripts.BowerImport
```

---

There are two significant changes made in this PR.

The first is the introduction of the `Step` and `KeyedStep` newtypes around functions. These functions allow us to implement validation steps at the level of individual packages (`Step`) or individual packages at a specific version (`KeyedStep`). The second is the introduction of the `ImportError` error type, which contains extra information about what sorts of errors can occur when converting a Bower package to a manifest.

For example, if you wanted to write a check that just verifies that the license field in a Bowerfile is correct, then you can do that (without having to understand the entire pipeline) with a `KeyedStep`:

```purs
checkLicense 
  :: forall a
   . KeyedStep 
      a -- any extra data you want to carry around
      { license :: Array String } -- your input in this step
      { license :: SPDX.License } -- your output from this step
checkLicense = KeyedStep \(package :: PackageName) (version :: GitHub.Tag) (extra :: a) { license } -> do
  let
    rewrite = case _ of
      [ "Apache 2" ] -> [ "Apache-2.0" ]

  parsed <- except (lmap ImportError.BadLicense) (SPDX.parse $ rewrite license)
  
  when myCondition do 
    throwError $ ImportError.BadLicense "Bad license!"

  pure { license: parsed }
```

A `Step` will be applied to every package in the registry, collecting failed packages along with their errors into a map of failed packages. A `KeyedStep` will be applied to every package _at every version of that package_ in the registry, again collecting failed packages at particular versions with their failure reason.

When we're all working on this and figuring out how to do these checks in a way that preserves as many repositories as possible, it's nice to think about the process at the level of an individual check and not have to walk deep into some nested `forWithIndex` and `case` statements.

---

As far as next steps go, @colinwahl was going to work on cleaning up the individual steps here to make them a little more single-purpose and structured. We're going to add a few more steps, too, that attempt to recover things like package licenses instead of failing licenses out. We're also going to try and make the diagnostics report a little nicer.

Another next step is to use the work in #194 to actually upload some of these manifests. At the moment, the manifest file is nearly 300,000 lines of packages via `stringifyWithIndent 2`, so there are plenty to add in to an initial index.

---

Here are some excerpts from the various files. First, the `bower-exclusions.json`:

```json
{
  "aws-encryption-sdk": [
    "v0.1.0"
  ],
  "b64": [
    "v0.0.6",
    "v0.0.7"
  ],
}
```

Next, the `bower-failures.json`:

```json
{
 "manifestError.mismatchedName.missingLicense.badVersion": {
    "csv-rk": {
      "v1.2": "Bower file should have name purescript-csv-rk but has name purescript-csv, No 'license' field, Invalid 'version' field: v1.2"
    }
  },
}
```

Finally, a valid manifest produced in `bower-manifests.json`:

```
[
  "aff",
  [
    {
      "version": "0.16.0",
      "targets": {
        "lib": {
          "sources": [
            "src/**/*.purs"
          ],
          "dependencies": {
            "console": ">=0.1.0 <0.2.0-0",
            "exceptions": ">=0.3.0 <0.4.0-0",
            "functions": ">=0.1.0 <0.2.0-0",
            "transformers": ">=0.8.1 <0.9.0-0"
          }
        }
      },
      "repository": {
        "githubOwner": "purescript-contrib",
        "githubRepo": "purescript-aff"
      },
      "name": "aff",
      "license": "Apache-2.0"
    }
  ]
]
```


ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -48,182 +72,334 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+        , dependencies: deps
+        , devDependencies: devDeps
+        , source: url
+        }
+
+    pure normalized
+
+-- | Verify that the dependencies listed in the bower.json files are all
+-- | contained within the registry.
+checkSelfContained :: Map PackageName GitHub.Address -> KeyedStep GitHub.Address Bowerfile Bowerfile
```
thomashoneyman: This is an example of how a broken-out step can be reasoned about in (somewhat) isolation, and make it easier for us to all collaborate on this file and understand what is happening in a specific step. I anticipate this file growing as we find more ways to transform Bower packages to include as many as possible, and I hope this will help keep it understandable as we add more and more conditions.

thomashoneyman:
@f-f when reviewing this code you may find it helpful to compare the old and new implementations for individual steps -- that way you can at least see that the importing code is largely the same, even though the surrounding scaffolding has changed.

I'm all about reducing lines of code where possible, but some of these functions are considerably longer in the new version because we're collecting multiple errors rather than crashing the program at the first error. I've also structured things to preserve similarity with the original implementation, even though we could certainly clean the steps up a bit (for example, we never actually go through and parse all package names as `PackageName`, though we probably should).

Here are the steps side-by-side:

**Read the legacy registry files**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L62-L82

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L509-L529

**Fetch all releases for each package**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L83-L89

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L123-L141

**Skip releases that we want to exclude**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L96-L99

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L143-L150

**Fetch bowerfiles for each release**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L105-L116

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L152-L165

**Ensure all dependencies are in the registry**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L195-L206

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L167-L196

**Convert to a manifest file**

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L129-L136

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L198-L204

These are using a `toManifest` function, which has also changed a bit:

https://github.com/purescript/registry/blob/09f5d667dd853df18f46dee69b6f2edb920bf2d5/ci/src/Registry/Scripts/BowerImport.purs#L152-L193

https://github.com/purescript/registry/blob/41b5e50e9c9637dc1879f9b1d42fffd800280630/ci/src/Registry/Scripts/BowerImport.purs#L206-L317

ci/bower-exclusions.json
```diff
@@ -1,653 +1,433 @@
...
-    ],
-    "csv": [
-      "rm",
-      "v1.2"
-    ],
-    "csv-rk": [
-      "v1.2"
-    ]
-  }
+  "aws-encryption-sdk": [
```
f-f, line 2: We used to have the reason for the exclusion here, but we don't have that anymore. Is this just an oversight or are we doing something different with these?

ci/src/Registry/Prelude.purs
```diff
@@ -36,8 +40,25 @@ import Effect.Ref (Ref) as Extra
...
+import Node.FS.Aff as FS
 import Node.Path (FilePath) as Extra
 
+-- | Partition an array of `Either` values into failure and success  values
+partitionEithers :: forall e a. Array (Either.Either e a) -> { fail :: Array e, success :: Array a }
+partitionEithers = Array.foldMap case _ of
+  Either.Left err -> { fail: [ err ], success: [] }
+  Either.Right res -> { fail: [], success: [ res ] }
+
+unpackErrors :: forall e a. Either.Either e a -> Maybe.Maybe e
```
f-f: I don't find this name fitting as I feel it's trying to carry an application-level meaning even though the function itself is very generic - I'd prefer a more generic name like `unpackLefts`

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -48,182 +72,338 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
-downloadBowerRegistry :: Aff RegistryIndex
-downloadBowerRegistry = do
-  -- Get the lists of packages: Bower packages and new packages
-  -- Assumption: we are running in the `ci` folder or the registry repo
-  bowerPackagesStr <- FS.readTextFile UTF8 "../bower-packages.json"
-  newPackagesStr <- FS.readTextFile UTF8 "../new-packages.json"
+  log "Starting import from legacy registries..."
+  manifests <- downloadLegacyRegistry
+  log "Writing manifests file..."
+  writeJsonFile manifestsFile manifests
```
f-f, line 53: Is there a reason why we write all the manifests to a file instead of printing some of them? Or was this waiting for #194 to go in so that we can just write them to the registry index instead?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -48,182 +72,338 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+  log "Done!"
+
+-- | A map of package names to valid manifests that have been imported from the
+-- | old registries and which can be added to the new registry.
+type PackageManifests = Map String (NonEmptyArray Manifest)
```
f-f: I usually prefer to convey information in types as a form of documentation - I'd suggest we don't use `String` where we could use a more meaningful name (e.g. `RawPackageName` here).
Making `newtype`s for this is not always worth it (e.g. I think it's not worth it here, yet at least), so I think we could just use a `type RawPackageName = String`

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -48,182 +72,338 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+      >>= runKeyedStep (skipExcluded exclusions)
+      >>= runKeyedStep fetchBowerfile
+      >>= runKeyedStep (checkSelfContained allPackages)
+      >>= runKeyedStep convertToManifest
+
+  -- Then, we write failed packages out to files so we can fix the errors or
+  -- simply list packages that won't be included in the new registry.
+  log "Writing exclusions and errors files..."
+  writeFailuresFile failures
+  writeExclusionsFile failures
```
f-f, line 91: Here we write both the `exclusionsFile` and the `failuresFile` using the `failures` collection - why do we need both?
I feel like we could just use the `failuresFile` to exclude packages from being fetched?

It's important to discuss the workflow of "how does it work when we improve the code and we want to try importing packages that were excluded before?" - I suspect that with this code one would delete the `exclusionsFile` so that the script would try to fetch everything once more.
If we would unify them then we would delete the `failuresFile`, but otherwise the workflow would still work?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -48,182 +72,338 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+    manifests = do
+      let toManifestArray { versions } = NEA.fromArray $ Array.fromFoldable $ Map.values versions
+      Map.mapMaybe toManifestArray packages
+
+  log "Finished transforming legacy packages!"
+  pure manifests
+
+-- | Find all released tags for the package, returning the parsed repository
+-- | location and a map of tags.
+getReleases :: Step String { meta :: GitHub.Address, versions :: Map GitHub.Tag Unit }
```
f-f: Same note as above re using `String` instead of a more meaningful name here - what does this string contain?

thomashoneyman:
@f-f In the latest commit, https://github.com/purescript/registry/pull/200/commits/f9aaf7b082a07f6953dd519fc29d06bcf3d5616c, we restored the `fromBower` option for adding and updating a package. With that in place this doesn't reduce any functionality from the existing code, only adds some.

ci/src/Registry/API.purs
```diff
@@ -165,16 +171,26 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
...
+          name =
+            PackageName.print packageName
+
+          printErrors =
+            String.joinWith ", " <<< map printManifestError <<< NEA.toArray
+
+          runManifest =
+            runExceptT <<< mapExceptT (liftAff <<< map (lmap printErrors))
+
+        runManifest (toManifest name metadata.location ref bowerfile) >>= case _ of
```
thomashoneyman: This now collects all errors with the manifest and reports them to the user, instead of only throwing the first error found in the Bowerfile.

f-f: Wonderful!

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -4,40 +4,45 @@ import Registry.Prelude
 
 import Affjax as Http
 import Affjax.ResponseFormat as ResponseFormat
-import Control.Monad.Error.Class (class MonadThrow)
-import Data.Argonaut as Json
+import Control.Monad.Error.Class (throwError)
+import Control.Monad.Except (ExceptT, except, lift, mapExceptT, runExceptT)
+import Control.Monad.State (modify, runStateT)
+import Data.Argonaut (class DecodeJson, class EncodeJson, decodeJson, encodeJson, jsonParser, parseJson, printJsonDecodeError)
```
f-f: I would like to only import things qualified, as it makes it easy to see what's coming from the same module vs what's being imported.
Though common types and classes (`NonEmptyArray`, `DecodeJson`, etc) can be included in our custom `Prelude`

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -47,181 +52,338 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+      targets <- except eitherTargets
+      pure { name, license, repository, targets, version }
+
+    Just err ->
+      throwError err
+
+-- | Perform an effectful transformation on a package, returning the transformed
+-- | package. If a package cannot be transformed, throw an `ImportError`
+-- | exception via `ExceptT` and the error will be collected.
+data Step a b = Step String (String -> a -> ExceptT ImportError Aff b)
```
f-f: I feel like the names `Step` and `KeyedStep` don't quite help to convey the meaning of what's happening.
I'd like more explicit names for these, how about something like `PackageTransform` and `PackageVersionTransform`?

f-f (commented):
Thanks for putting this together @thomashoneyman and @colinwahl!

I left some comments and I think we can merge after we address these.

Unfortunately the PR is quite big and involved so I wasn't able to review it as well as I would have liked, so I skimmed over many parts as I couldn't quite relate them to the existing code.
I would love if going forward we could try to keep PRs smaller and have more gradual refactorings - they require more care when putting them together, but I think it's worth it as reviewing smaller patches is a much faster process while allowing for a deeper understanding of the code.

ci/bower-exclusions.json
thomashoneyman: We used to use this exclusions file to exclude all failing packages from the import tool (if I am understanding it correctly). However, we should exclude as few packages as possible, because as this tool improves our errors become more informative and more granular. If a package is excluded, then it won't benefit from these updates.

Now, I only use this exclusions file to exclude packages that don't have a Bowerfile. I am excluding them because I don't want to try and fetch them each time the tool runs (the fetch will always fail), and we will fetch them each time because there's no file to cache in our bower cache.

I could alternately skip having the exclusions file altogether and cache an empty file when a package does not have a Bowerfile. That way the only output for packages that fail is the bower-failures.json file, which lists packages with their failure reason. Also, if you'd prefer that get renamed back to `bower-exclusions.json` that is also fine by me!

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: Yes, this was waiting for #194, and also because we can improve this pipeline more before we’re ready to push things into the registry officially.

I could change it so that we write the registry index to a file; would you prefer that? Or just not write anything?

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: Originally I’d implemented this as PackageName, but I reverted it to stay close to the original implementation. I can introduce the type synonym, and we could also introduce a step where we parse the package names instead of keeping them raw all the way through. Then we’d have a Map PackageName, and packages that fail would have a MalformedPackageName error (or we could try to fix them).

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: Perhaps we could discuss here? It’s the same subject:

https://github.com/purescript/registry/pull/200#discussion_r693499345

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: Sounds good to me!

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: I've updated this to use `RawPackageName`. I would _rather_ use the `PackageName` type, which has actual meaning (it's parsed and known to be valid) vs. the type synonym over `String`, but I think that ought to be something we do later on, if we do it.

thomashoneyman:
Closing in favor of #210

PR #203, opened 2021-08-20
Add web-page-visibility

toastal:
[`purescript-web-page-visibility`](https://git.sr.ht/~toastal/purescript-web-page-visibility) is a wrapper for the W3C Page Visibility spec. `v2.0.0` tag [build passed](https://builds.sr.ht/~toastal/job/576410) (Nix + Flakes + Cachix, merely builds as there is not much to test).

[builds.sr.ht status](https://builds.sr.ht/~toastal/purescript-web-page-visibility/commits/.build.yml?)

I have a future plans with this library, but one thing I really wanted to test is the Registry handling non-GitHub git repository—this will be the first repository. Sourcehut’s read-only Git cloning URI is `https://git.sr.ht/ownerCanonicalName/repositoryName` _without_ `.git` (in fact, putting `.git` will return a `403`). As Git is distributed, I will be excited to see what will happen in the future as others see examples of packages on multiple platforms.

Pursuit doc upload though will be blocked by [Pursuit #422](https://github.com/purescript/pursuit/issues/422) and whatever happens there. This may be blocked by #204 which I opened for Sourcehut-specific support or #15 which allows arbitrary git repo support.

PR #204, opened 2021-08-21
Initial Sourcehut support

toastal:
Looking for feedback. Could open up #203. I'm adding Sourcehut-specific code in here for the same reason it is in Nixpkgs ([`fetchFromSourcehut`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/fetchsourcehut/default.nix)) and GitHub in this package: access to the tarball downloads which are faster.

---

## Context

```graphql
query {
 repositoryByName(name: "purescript-web-page-visibility") {
    name
    owner { id canonicalName }
    description
  }
  user(username: "toastal") { id username }
}
```

cURLing Sourcehut's API that gets you

```json
{
  "data": {
    "repositoryByName": {
      "name": "purescript-web-page-visibility",
      "owner": {
        "id": 11755,
        "canonicalName": "~toastal"
      },
      "description": "Type definitions and low level interface implementations for W3C Page Visibility API"
    },
    "user": {
      "id": 11755,
      "username": "toastal"
    }
  }
}
```

`v1.0.1` ref: https://git.sr.ht/~toastal/purescript-web-page-visibility/refs/v1.0.1
tarball from that ref: https://git.sr.ht/~toastal/purescript-web-page-visibility/archive/v1.0.1.tar.gz

Canonical name is used for users and groups and such and is separate kinda from username and id as far as repo URLs are concerned.

There is duplicate wget+untar commands going on, but I don't know the proper way to refactor than in the context of this project. Currently there are no tests since nothing has been approved yet for Sourcehut (though #203 is open), but also it _is_ duplicated tarball code.

The Dhall names are longer, not matching `GitHubData` or nixpkgs, to more closely resemble Sourcehut's [schema](https://git.sr.ht/~sircmpwn/git.sr.ht/tree/master/item/api/graph/schema.graphqls) to be more clear to users of that platform (although I don't really agree with how some of the current Dhall is organized having built some bigger Dhall projects myself (e.g. not using `Type` + `default` to get access to `::` operator), I don't want to diverge from what's here already).

I'm willing to refactor anything. I'm very much in full support of at least one GitHub alternative. I could look into GitLab after. I think these may require a little extra care though as self-hosted Sourcehuts and GitLabs and Giteas and cgits will have different URLs despite the same tarball and other APIs.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -139,8 +139,23 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
   -- fetch the repo and put it in the tempdir, returning the name of its toplevel dir
   folderName <- case metadata.location of
     Git _ -> do
-      -- TODO: Support non-GitHub packages. Remember subdir when implementing this. (See #15)
-      throwWithComment "Packages are only allowed to come from GitHub for now. See #15"
+      -- TODO: Support non-GitHub/Sourcehut packages. Remember subdir when implementing this. (See #15)
+      throwWithComment "Packages are only allowed to come from GitHub or Sourcehut for now. See #15"
+    Sourcehut { owner, repo, subdir } -> do
+      when (isJust subdir) $ throwWithComment "`subdir` is not supported for now. See #16"
+      let tarballName = ref <> ".tar.gz"
+      let absoluteTarballPath = tmpDir <> "/" <> tarballName
+      let archiveUrl = "https://git.sr.ht/" <> owner <> "/" <> repo <> "/archive/" <> tarballName
+      log $ "Fetching tarball from Sourcehut: " <> archiveUrl
+      wget archiveUrl absoluteTarballPath
+      log $ "Tarball downloaded in " <> absoluteTarballPath
+      liftEffect (Tar.getToplevelDir absoluteTarballPath) >>= case _ of
+        Nothing ->
+          throwWithComment "Could not find a toplevel dir in the tarball!"
+        Just dir -> do
+          log "Extracting the tarball..."
+          liftEffect $ Tar.extract { cwd: tmpDir, filename: absoluteTarballPath }
+          pure dir
     GitHub { owner, repo, subdir } -> do
       -- TODO: Support subdir. In the meantime, we verify subdir is not present. (See #16)
       when (isJust subdir) $ throwWithComment "`subdir` is not supported for now. See #16"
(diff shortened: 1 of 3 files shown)


f-f:
Thank you for putting this together @toastal!

While we are looking to support more providers (and as you probably realized it's not a big fuss to add a new one), right now the source is quite in flux (e.g. see #200) while we try to fix #95, which should be the last big blocker for a whole set of things (like "downloading packages from the registry", "moving package-sets to this repo", etc)

So I wouldn't like to merge this work right now, as that would mean increasing the scope and the functionality we have to maintain, while we are trying to keep that smaller so that we can get to a full registry pipeline ASAP.
I think we should revisit this once the default publishing pipeline (i.e. GitHub) is fairly settled and up to speed.

toastal:
> we are looking to support more providers

I've noticed the issues all stating that it needs to be platform agnostic. I very much appreciate it <3

> and as you probably realized it's not a big fuss to add a new one

yup

---

I understand. I opened such a small repo mostly to have a minimal open request to get some level of priority out there as well as setting the real-world precedence for support. If things are left as is the, you'll continue to see "and just GitHub because that's all we ever had asked for". I think people less politically-inclined on this topic will just choose GitHub to lower their friction to upload a package for others to consume, but _nah_ I want to increase friction to see as much as possible migrate and decentralize *away* from Microsoft's GitHub in the present and future.

But y'all can only do so much, and one step at a time.

I would like to volunteer anything I _can_ do at this time (though I don't have the same ecosystem experience). As for now, I will continue adding custom packages to my `{spago,packages}.dhall` files from alternatives.

thomashoneyman:
@toastal I just wanted to let you know that we are doing some final refactoring of the registry in preparation for the Spago part of the alpha, at which point I would like to come back and add explicit Sourcehut support. I haven't forgotten this PR!

toastal:
Generic Git is good enough. Seems weird GitHub gets a special case since it’s just Git too, but regardless, I don’t think there is much purpose behind this anymore. Now if we wanted to talk Darcs support… hit me up!

thomashoneyman:
> Seems weird GitHub gets a special case since it’s just Git too

We could probably get away with a generic git for GitHub too; it's just used to fetch repo contents and information a little easier without going via `git clone ...`.

And hey, you wanna talk Darcs support, the goal is to support arbitrary package sources so long as they're reachable somewhere online!

We're really just blocked on Pursuit / `purs publish` now. The registry is designed to support monorepos and non-git sources. It's just disabled until Pursuit and `purs publish` can, too.

PR #210, opened 2021-09-01
Improve Bower import by collecting errors and producing a valid registry index

thomashoneyman:
This PR updates the Bower import script to produce a valid `RegistryIndex` as its result, and to collect detailed errors for all packages that cannot be converted into valid manifests. You can execute the script with:

```console
$ GITHUB_TOKEN=token spago run -m Registry.Scripts.BowerImport --quiet
```

The import errors will be written to the existing `bower-exclusions.json` file on each run. However, packages with the following error types in the exclusions file are excluded when running the tool to avoid hitting GitHub with useless requests:

- `InvalidGitHubRepo` (we can't fetch and cache releases)
- `NoReleases` (we can't cache releases)
- `MissingBowerfile` (we can't cache the bowerfile)
- `MalformedBowerJson` (we _could_, but don't, cache the bowerfile -- we only cache valid ones)

If you'd like to make a fresh run of the tool, delete the `bower-exclusions.json` file.


ci/bower-exclusions.json
```diff
@@ -1,653 +1,3616 @@
...
-      "v1.2.1"
-    ],
-    "cycle": [
-      "v0.0.1",
-      "v0.0.2"
-    ]
+  "malformedPackageName": {
+    "TypeAhead": {
+      "v0.1.0": "Malformed name: Package name should start with a lower case char or a digit; pos = 0"
+    }
```
thomashoneyman: This `bower-exclusions.json` file no longer contains only package versions that we want to exclude when running the tool (because they throw an exception). Instead, it contains _all_ package versions that fail to produce a valid manifest, as well as a detailed error as to why they have failed.

In the future, we may wish to produce a `warnings` file that contains packages that could only be turned into a manifest because we made a change (for example, fixing the license name).

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -0,0 +1,130 @@
...
+import Data.String as String
+import Registry.PackageName (PackageName)
+import Registry.PackageName as PackageName
+import Safe.Coerce (coerce)
+
+-- | An import error printed as a key usable in a map
+newtype ImportErrorKey = ImportErrorKey String
+derive instance Newtype ImportErrorKey _
+derive newtype instance Eq ImportErrorKey
+derive newtype instance Ord ImportErrorKey
```
thomashoneyman: I don't usually add newtypes like this, but I found out (twice!) that a string I _thought_ was a `RawPackageName` was really a `RawVersion`, and so on, and because we do so many map transformations it does help quite a bit to have the newtypes available.

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -0,0 +1,130 @@
...
+  ExcludedPackage -> "excludedPackage"
+  NoReleases -> "noReleases"
+  MalformedPackageName _ -> "malformedPackageName"
+  MissingBowerfile -> "missingBowerfile"
+  MalformedBowerJson _ -> "malformedBowerJson"
+  NonRegistryDependencies _ -> "nonRegistryDependencies"
+  NoManifests -> "noManifests"
+  ManifestError errs ->
+    "manifestError."
+      <> String.joinWith "." (printManifestErrorKey <$> NEA.toArray errs)
```
thomashoneyman: I'm open to other ideas of how to handle this (for example, we could just have a `manifestError` key here that doesn't include the sub-types).

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -0,0 +1,130 @@
...
+
+printImportError :: ImportError -> String
+printImportError = case _ of
```
thomashoneyman: This is somewhat problematic, because we do also _parse_ the exclusions file to produce a `PackageFailures` map so that we can filter out some failing packages from the execution of the tool. In a subsequent PR we probably ought to write `EncodeJson` / `DecodeJson` instances for these, and an instance for `PackageFailures`, so that we can round-trip encoding and decoding a `PackageFailures` map.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -47,161 +48,273 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
-  Debug.traceM registry
--- TODO: upload packages
-
--- | We go through all the legacy packages, and:
--- | - collect all the releases of each package
--- | - download the bower.json file for every release, caching it
--- | - return an index of this "Bower Registry"
-downloadBowerRegistry :: Aff (Map String (Map GitHub.Tag Manifest))
+  log "Starting import from legacy registries..."
+  _registry <- downloadBowerRegistry
```
thomashoneyman, line 51: Here's our valid `RegistryIndex`!

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -210,82 +323,293 @@ withCache
...
+      throwError $ Aff.error $ Json.printJsonDecodeError err
+    Right val ->
+      pure $ fromExclusions val
+  where
+  fromExclusions :: Object (Object (Object String)) -> PackageFailures
+  fromExclusions =
+    fromObject ImportErrorKey
+      <<< map (fromObject RawPackageName)
+      <<< map (map fromErrorObject)
+      <<< filterExclusions
```
thomashoneyman: This should really be a `DecodeJson` instance for `PackageFailures`. As it stands, `readExcludedPackages` will not actually produce a valid `PackageFailures` containing all errors -- it will only include errors that we want to exclude packages for. If we wrote `EncodeJson` and `DecodeJson` instances for `PackageFailures` that round-trip, then the outer function just becomes:

```purs
readExcludedPackages :: FilePath -> Aff PackageFailures
readExcludedPackages = readJsonFile
```

thomashoneyman: See also the `filterExclusions` function below, which does some rudimentary (read: terrible) parsing of strings back into `ImportError`s.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -210,82 +323,293 @@ withCache
...
-      $ f o.missingLicense
-      <> f o.malformedLicense
-      <> f o.missingBowerfile
-      <> f o.malformedBowerfile
-      <> f o.dependOnCommitHashOrBranch
-      <> f o.dependenciesOutsideOfRegistry
-      <> f o.invalidRelease
+-- | Execute the provided transform on every package in the input packages map
+-- | collecting failures into `PackageFailures` and saving transformed packages.
+forPackage
```
thomashoneyman, line 400: This function and the below `forPackageVersions` should be seen as a sort of `foldlTraverseWithIndex` where at any step an exception can be thrown and the failed package / package version will be inserted into the `PackageFailures` map. When processing packages we should always use `forPackage` or `forPackageVersions`.

ci/src/Registry/Scripts/BowerImport/Error.purs
f-f: I think newtypes here are a great idea, I was feeling that we'd need them sooner than later :blush:

thomashoneyman:
@f-f @colinwahl This now works just fine locally. Try running the script, and then running it again, to see that the filtering is preventing failing packages with errors that cause them to miss the cache from being requested again. I'd also check out the `bower-exclusions.json` generated file to see what ya think.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -47,161 +45,279 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+
+    bowerName = stripPureScriptPrefix bowerfile.name
+
+    eitherName = case PackageName.parse $ stripPureScriptPrefix bowerName of
+      Right name | bowerName == PackageName.print package ->
+        Right name
+      Right _ -> do
+        mkError $ MismatchedName { expected: package, received: RawPackageName bowerfile.name }
+      Left _ -> do
+        mkError $ MismatchedName { expected: package, received: RawPackageName bowerfile.name }
```
thomashoneyman: It has turned out that many packages have bogus values as their bowerfile names, and Bower packages are downloaded from their URL anyways. With that in mind, we should perhaps skip parsing the bowerfile.name field altogether and just use the package URL (ie. the existing parsed `PackageName` here).

thomashoneyman: We could collect a warning if the Bowerfile name doesn't match, though, so that it's more obvious why this is changed.

f-f: > With that in mind, we should perhaps skip parsing the bowerfile.name field altogether and just use the package URL (ie. the existing parsed PackageName here)

Yeah, let's just do this. I don't think there's much value in logging a warning either, it's not like these are going to be fixed, and the authoritative name is anyways the one that was used to register to the registry.

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -0,0 +1,119 @@
...
+import Data.Argonaut.Encode.Generic as Json.Encode.Generic
+import Data.Argonaut.Types.Generic as Json.Generic
+import Data.Generic.Rep (class Generic)
+import Registry.PackageName (PackageName)
+import Registry.Prelude as Maybe
+import Safe.Coerce (coerce)
+
+-- | A map of error types to package names to package versions, where failed
+-- | versions contain rich information about why they failed.
+newtype PackageFailures = PackageFailures (Map ImportErrorKey (Map RawPackageName (Map (Maybe RawVersion) ImportError)))
```
f-f: I forgot, what was the case in which this `Maybe RawVersion` goes to `Nothing`?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -47,161 +45,279 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+      -- rate-limited us. Or at least we should collect the errors and print them
+      -- out because it's quite possible the Bowerfile actually does work and
+      -- we could have more granular errors to avoid excluding those packages.
+      Left err -> do
+        log $ "Unable to retrieve bowerfile. Bad request: " <> Http.printError err
+        throwError MissingBowerfile
+      Right { body, status }
+        | status /= StatusCode 200 -> do
+            log $ "Unable to retrieve bowerfile. Bad status: " <> body
+            throwError MissingBowerfile
```
f-f: I think we should check if we get a 404, and throw a `MissingBowerfile` only in that case - this is because every other code is temporary and we want to retry next time we run the script, rather than permanently skip the fetch

ci/src/Registry/Scripts/BowerImport/Error.purs
thomashoneyman: We need a `Map (Maybe RawVersion) ImportError` because:

1. In the case we cannot fetch versions for a package (ie. it has no releases, or it has a malformed GitHub repo), then there is no version but we still want to record an `ImportError` for the package. That's represented with `Map.singleton Nothing NoReleases`, for example.
2. In the case we have versions for a package, then we want to associate `ImportError`s according to particular package versions.

thomashoneyman:
@f-f feedback resolved 💪

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -47,161 +46,273 @@ type PackageReleases = { address :: GitHub.Address, releases :: Set GitHub.Tag }
...
+      Right address -> pure address
+
+    let repoCache = Array.fold [ "releases__", address.owner, "__", address.repo ]
+
+    releases <- withCache repoCache (Just $ Hours 24.0) do
+      log $ "Fetching releases for package " <> un RawPackageName name
+      result <- lift $ try $ GitHub.getReleases address
+      case result of
+        Left err -> logShow err *> throwError NoReleases
+        Right v -> pure v
```
thomashoneyman, line 98: A package has been removed from GitHub entirely (`toastr`), so this fails with a 404 error when fetching releases now. This update doesn't crash but rather collects a `NoReleases` error.

f-f: This sounds like a great success? :smile:

ci/src/Registry/Scripts/BowerImport/Error.purs
f-f: In case (2) we wouldn't have any `Nothing` right? Because `RawVersion` is just a `Text`, so if we manage to fetch releases then we'd always have a release name in there.
If that makes sense then maybe `Either ImportError (Map RawVersion ImportError)` could be a more fitting type?

thomashoneyman: It depends on whether we anticipate ever associating an error with a package generally (not with a version), and _also_ associating errors with the package at particular versions. I don't think so, so I think it's OK to switch to `Either`.

f-f (approved):
:clap: :clap: :clap:

PR #221, opened 2021-09-09
Lenient json

### LINKED ISSUE #211: Decode Bowerfiles as lenient JSON and only parse necessary fields
In #210 we saw a number of packages fail with a `MalformedBowerJson` error, namely because of issues like this:

```json
{
  "name": "purescript-facebook",
  "description": "Idiomatic Purescript bindings for the Facebook SDK",
  "license": "Apache-2.0",
  "keywords": "facebook purescipt",
  ...
}
```

This fails to decode as a `Bower.PackageMeta` because the `keywords` field has to be an _array_ of keywords, not a plain string. We don't even use this field in our manifests, though, so we shouldn't care if it isn't Bower-compliant!

The errors aren't always just that the file can't be decoded as a `Bower.PackageMeta`. Sometimes they aren't valid JSON at all:

```json
{
  "name": "purescript-endpoints-express",
  "version": "0.0.1",
  "authors": {
    "Simon Van Casteren <simon.van.casteren@gmail.com>"
  },
  ...
}
```

This file isn't valid JSON because of the `authors` key, which has an object with no key/value pairs as its value. We can't parse this as `Json` at all.

And yet we _can_ pull off the keys that we know need to be there -- the name, the version, the dependencies, and the devDependencies -- and attempt to decode _their_ contents. To avoid needlessly removing packages from the registry, I suggest that we change how we parse Bowerfiles once #210 merges so that packages like these can be converted into manifests despite their JSON issues.

There are some we maybe still can't parse as `Json`, because even the keys we're looking for have little issues. For example, the next one is even more insidious:

```json
{
    "name": "purescript-uint",
    ...
    "dependencies": {
        "purescript-enums": "^v5.0.0",
        "purescript-gen": "^v3.0.0",
        "purescript-math": "^v3.0.0",
        "purescript-maybe": "^v5.0.0",
        "purescript-prelude": "^v5.0.0",
        "purescript-psci-support": "^v5.0.0",
    }
    "devDependencies": {
        "purescript-effect": "^v3.0.0",
        "purescript-quickcheck": "^v7.1.0",
        "purescript-quickcheck-laws": "^v6.0.1"
    }
}
```

There's a trailing comma after the `"purescript-psci-support`" dependency! It's not valid JSON! It would be nice if we could handle this case too, but perhaps we just pull off the keys we need and if they're not valid JSON then we just drop the package.

YanikCeulemans:
This PR introduces the jsonic package to provide more lenient json parsing for bower.json files.
See https://github.com/purescript/registry/issues/211

### DIFF
--- a/ci/package.json
+++ b/ci/package.json
@@ -15,6 +15,7 @@
     "@octokit/rest": "^18.0.0",
     "aws-sdk": "^2.814.0",
     "fuse.js": "^6.4.6",
+    "jsonic": "^1.0.1",
     "semver": "^7.3.4",
     "spdx-expression-parse": "^3.0.1",
     "tar": "^6.1.9",
--- a/ci/src/Registry/Scripts/BowerImport.purs
+++ b/ci/src/Registry/Scripts/BowerImport.purs
@@ -204,8 +204,8 @@ fetchBowerfile name address tag = do
 -- | Verify that the dependencies listed in the bower.json files are all
 -- | contained within the registry.
 selfContainedDependencies :: Set RawPackageName -> BowerFile -> ExceptT ImportError Aff Unit
-selfContainedDependencies registry (BowerFile {dependencies, devDependencies}) = do
-  let 
+selfContainedDependencies registry (BowerFile { dependencies, devDependencies }) = do
+  let
     allDeps = Object.keys $ dependencies <> devDependencies
   outsideDeps <- for allDeps \packageName -> do
     name <- cleanPackageName $ RawPackageName packageName
@@ -268,8 +268,8 @@ toManifest package repository version (BowerFile bowerfile) = do
             Nothing -> Left { dependency: packageName, failedBounds: versionStr }
             Just range -> Right $ Tuple (PackageName.print packageName) range
 
-      normalizedDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.dependencies
-      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.devDependencies
+      normalizedDeps <- normalizeDeps $ Object.toUnfoldable bowerfile.dependencies
+      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable bowerfile.devDependencies
 
       let
         readDeps = map checkDepPair >>> partitionEithers >>> \{ fail, success } ->
--- a/ci/src/Registry/Scripts/BowerImport.purs
+++ b/ci/src/Registry/Scripts/BowerImport.purs
@@ -33,7 +33,7 @@ import Registry.Index (RegistryIndex)
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Schema (Repo(..), Manifest)
-import Registry.Scripts.BowerImport.BowerFile (BowerFile)
+import Registry.Scripts.BowerImport.BowerFile (BowerFile(..))
 import Registry.Scripts.BowerImport.BowerFile as BowerFile
 import Registry.Scripts.BowerImport.Error (ImportError(..), ImportErrorKey, ManifestError(..), PackageFailures(..), RawPackageName(..), RawVersion(..))
 import Registry.Scripts.BowerImport.Error as BowerImport.Error
@@ -204,9 +204,9 @@ fetchBowerfile name address tag = do
 -- | Verify that the dependencies listed in the bower.json files are all
 -- | contained within the registry.
 selfContainedDependencies :: Set RawPackageName -> BowerFile -> ExceptT ImportError Aff Unit
-selfContainedDependencies registry bowerfile = do
+selfContainedDependencies registry (BowerFile {dependencies, devDependencies}) = do
   let 
-    allDeps = Object.keys $ BowerFile.dependencies bowerfile <> BowerFile.devDependencies bowerfile
+    allDeps = Object.keys $ dependencies <> devDependencies
   outsideDeps <- for allDeps \packageName -> do
     name <- cleanPackageName $ RawPackageName packageName
     pure $ if Set.member name registry then Nothing else Just name
@@ -222,7 +222,7 @@ toManifest
   -> SemVer
   -> BowerFile
   -> ExceptT (NonEmptyArray ManifestError) Aff Manifest
-toManifest package repository version bowerfile = do
+toManifest package repository version (BowerFile bowerfile) = do
   let
     mkError :: forall a. ManifestError -> Either (NonEmptyArray ManifestError) a
     mkError = Left <<< NEA.singleton
@@ -239,7 +239,7 @@ toManifest package repository version bowerfile = do
           [ "3-Clause BSD" ] -> [ "BSD-3-Clause" ]
           other -> other
 
-        { fail, success } = partitionEithers $ SPDX.parse <$> rewrite (BowerFile.license bowerfile)
+        { fail, success } = partitionEithers $ SPDX.parse <$> rewrite bowerfile.license
 
       case fail of
         [] -> case success of
@@ -268,8 +268,8 @@ toManifest package repository version bowerfile = do
             Nothing -> Left { dependency: packageName, failedBounds: versionStr }
             Just range -> Right $ Tuple (PackageName.print packageName) range
 
-      normalizedDeps <- normalizeDeps $ Object.toUnfoldable $ BowerFile.dependencies bowerfile
-      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable $ BowerFile.devDependencies bowerfile
+      normalizedDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.dependencies
+      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.devDependencies
 
       let
         readDeps = map checkDepPair >>> partitionEithers >>> \{ fail, success } ->
--- a/ci/src/Registry/Scripts/BowerImport.purs
+++ b/ci/src/Registry/Scripts/BowerImport.purs
@@ -33,11 +33,12 @@ import Registry.Index (RegistryIndex)
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Schema (Repo(..), Manifest)
+import Registry.Scripts.BowerImport.BowerFile (BowerFile)
+import Registry.Scripts.BowerImport.BowerFile as BowerFile
 import Registry.Scripts.BowerImport.Error (ImportError(..), ImportErrorKey, ManifestError(..), PackageFailures(..), RawPackageName(..), RawVersion(..))
 import Registry.Scripts.BowerImport.Error as BowerImport.Error
 import Safe.Coerce (coerce)
 import Text.Parsing.StringParser as StringParser
-import Web.Bower.PackageMeta as Bower
 
 -- | This main loop uploads legacy packages to the new Registry
 -- | In order to do this, we:
@@ -170,7 +171,7 @@ downloadBowerRegistry = do
 
 -- | Find the bower.json files associated with the package's released tags,
 -- | caching the file to avoid re-fetching each time the tool runs.
-fetchBowerfile :: RawPackageName -> GitHub.Address -> RawVersion -> ExceptT ImportError Aff Bower.PackageMeta
+fetchBowerfile :: RawPackageName -> GitHub.Address -> RawVersion -> ExceptT ImportError Aff BowerFile
 fetchBowerfile name address tag = do
   let
     url = "https://raw.githubusercontent.com/" <> address.owner <> "/" <> address.repo <> "/" <> un RawVersion tag <> "/bower.json"
@@ -192,18 +193,21 @@ fetchBowerfile name address tag = do
         | status /= StatusCode 200 -> do
             log $ "Unable to retrieve bowerfile. Bad status: " <> body
             throwError $ BadStatus $ un StatusCode status
-        | otherwise -> case Json.parseJson body >>= Json.decodeJson of
+        | otherwise -> case BowerFile.parse body of
             Left err -> do
-              log $ "Unable to parse bowerfile: " <> Json.printJsonDecodeError err <> " Malformed body: " <> body <> "."
-              throwError $ MalformedBowerJson { error: Json.printJsonDecodeError err, contents: body }
-            Right (bowerfile :: Bower.PackageMeta) -> pure bowerfile
+              let
+                printedErr = BowerFile.printBowerFileParseError err
+              log $ "Unable to parse bowerfile: " <> printedErr <> " Malformed body: " <> body <> "."
+              throwError $ MalformedBowerJson { error: printedErr, contents: body }
+            Right bowerfile -> pure bowerfile
 
 -- | Verify that the dependencies listed in the bower.json files are all
 -- | contained within the registry.
-selfContainedDependencies :: Set RawPackageName -> Bower.PackageMeta -> ExceptT ImportError Aff Unit
-selfContainedDependencies registry (Bower.PackageMeta bowerfile) = do
-  let Bower.Dependencies allDeps = bowerfile.dependencies <> bowerfile.devDependencies
-  outsideDeps <- for allDeps \{ packageName } -> do
+selfContainedDependencies :: Set RawPackageName -> BowerFile -> ExceptT ImportError Aff Unit
+selfContainedDependencies registry bowerfile = do
+  let 
+    allDeps = Object.keys $ BowerFile.dependencies bowerfile <> BowerFile.devDependencies bowerfile
+  outsideDeps <- for allDeps \packageName -> do
     name <- cleanPackageName $ RawPackageName packageName
     pure $ if Set.member name registry then Nothing else Just name
   for_ (NEA.fromArray $ Array.catMaybes outsideDeps) \outside ->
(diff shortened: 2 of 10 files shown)


ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
+module Registry.Scripts.BowerImport.BowerFile
+  ( BowerFile
```
YanikCeulemans: Do we want to keep this constructor hidden? Considering it is just a container type, it might be more convenient to just export it.

This would give us the benefit of not having to export all the accessor functions below in exchange for loss of guarantees the smart constructor can give us.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+  ) where
+
+import Registry.Prelude
+
+import Data.Argonaut (Json, (.:), (.:?))
+import Data.Argonaut as Json
+import Data.Array as Array
+import Foreign.Jsonic as Jsonic
+
+newtype BowerFile =
```
YanikCeulemans, line 16: Is this a good name considering it is just a container type? Something like "BowerFileDto" might be clearer?

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+parse =
+  Jsonic.parse
+    >>> lmap JsonParseError
+    >=> parseBowerFile
+
+parseBowerFile :: Json -> Either BowerFileParseError BowerFile
+parseBowerFile json = lmap JsonDecodeError do
+  let
+    strings :: Json -> Either Json.JsonDecodeError (Array String)
+    strings = Json.caseJson
```
YanikCeulemans: I'm not sure about whether this is best way to turn either a json string , or json array of strings into the "license" property, meaning an array of strings

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (Array.singleton >>> Right)
+      (traverse Json.decodeJson)
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+  root <- Json.decodeJson json
+  version' <- root .: "version"
+  license' <- strings =<< root .: "license"
```
YanikCeulemans: I'm not sure about the expected behaviour here. The current implementation can handle a json string, or a json array of strings.

However, after searching the internet, I haven't been able to find a spec for a bower file so I'm not sure what the allowed types are for the license property.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: I would rather export the constructor for convenience. I'm not sure it even needs to be newtyped, actually; I think it would be fine as a simple type synonym. It's not a type we'll be using pervasively; it's more of an intermediate type that still needs more parsing. We'll parse it into a `Manifest`, and _that's_ the type we need to make sure is correct.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: I think this is fine. Or we could do something like `BowerFields` or something to represent these are _some_ of the fields from a Bowerfile.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+
+import Data.Argonaut (Json, (.:), (.:?))
+import Data.Argonaut as Json
+import Data.Array as Array
+import Foreign.Jsonic as Jsonic
+
+newtype BowerFile =
+  BowerFile
+    { version :: String
+    , license :: Array String
```
thomashoneyman: We might need to write a JSON decoder manually so as to deal with the fact that the LICENSE field could be a simple string or could be an array of strings (regardless of what's technically correct on the Bower side) and we'd ideally support both. In that case we _would_ need this to remain a `newtype`.

If you do end up writing a manual decoder (we don't need an encoder) then I'd give this a look:
https://github.com/purescript-contrib/purescript-argonaut-codecs/tree/main/docs#1-use-alternative-to-provide-fallback-decoders

In that case we do need to preserve the `newtype`; it can't be a simple type synonym.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+    , dependencies :: Object String
+    , devDependencies :: Object String
+    }
+
+derive newtype instance Show BowerFile
+derive newtype instance Json.DecodeJson BowerFile
+derive newtype instance Json.EncodeJson BowerFile
```
thomashoneyman: I think you could replace the `parseBowerfile` function with a manually-implemented `DecodeJson` instance. Then, your `parse` function would be `Jsonic.parse >>> lmap JsonParseError >=> decodeJson`.

If you would rather keep the `parseBowerfile` function at the top level then I don't think we should keep these instances -- otherwise, we could mistakenly call `decodeJson` to get a `Bowerfile` when you _really_ should use `parseBowerfile`.

Also, while it's not a problem, we don't need `EncodeJson` instances for this type. We don't write it.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: See this comment: https://github.com/purescript/registry/pull/221/files#r705583168

You could either use `<|>` or you can use the `caseJson`, either is good! I would pull this into a variable so it's not repeated so much:

```purs
(const $ Left $ Json.TypeMismatch "expected string or array")
```

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: We should allow either one, regardless of the Bower spec.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+  license' <- strings =<< root .: "license"
+  dependencies' <- fromMaybe mempty <$> root .:? "dependencies"
+  devDependencies' <- fromMaybe mempty <$> root .:? "devDependencies"
+  pure $
+    BowerFile
+      { version: version'
+      , license: license'
+      , dependencies: dependencies'
+      , devDependencies: devDependencies'
+      }
```
thomashoneyman: Any reason not to have your bindings be without apostrophes, like:

```purs
version <- root .: "version"
```

so that you can construct the record at the end like this:

```suggestion
  pure $ BowerFile { version, license, dependencies, devDependencies }
```

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,85 @@
...
+  let
+    strings :: Json -> Either Json.JsonDecodeError (Array String)
+    strings = Json.caseJson
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+      (Array.singleton >>> Right)
+      (traverse Json.decodeJson)
+      (const $ Left $ Json.TypeMismatch "expected string or array")
+  root <- Json.decodeJson json
```
thomashoneyman: You definitely don't have to do this, but common practice at least in this codebase is to name the decoded json after what sort of json it is, ie:

```purs
obj <- Json.decodeJson json`
```

for objects, or

```purs
arr <- Json.decodeJson json
```

for arrays, and so on.

ci/test/Foreign/Jsonic.purs
```diff
@@ -0,0 +1,50 @@
...
+parseTest str json = Spec.it str do
+  parseS str `Assert.shouldContain` Json.stringify json
+
+goodJson :: Spec
+goodJson = do
+  let
+    complexJson = Json.encodeJson { complex: { nested: "json", bool: true } }
+
+  parseTest "[1,2,3]" $ Json.encodeJson [ 1, 2, 3 ]
+  parseTest "{\"name\": \"test\"}" $ Json.encodeJson { name: "test" }
```
thomashoneyman: This test is pretty small, but if you'd rather not deal with the escapes you can use multiline strings (which also work on a single line):

```suggestion
  parseTest """{ "name":  "test" }""" $ Json.encodeJson { name: "test" }
```

thomashoneyman:
Great work so far, @YanikCeulemans! I've left some thoughts and some answers to your questions, but I don't think this is far off being merge-able. Thank you!

ci/package.json
```diff
@@ -15,6 +15,7 @@
     "@octokit/rest": "^18.0.0",
     "aws-sdk": "^2.814.0",
     "fuse.js": "^6.4.6",
+    "jsonic": "^1.0.1",
```
thomashoneyman, line 18: I think you'll need to run `npm install` to get a new lockfile also.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: Because I didn't want the variable shadowing compiler warnings due to the top level functions. Perhaps the warnings aren't a real problem?

ci/package.json
YanikCeulemans: Yes totally, forgot to check in the lock file. It had some many unrelated changes that I just reverted all the changes to it. After like 7 years of professional js development I'm still confused why it changes totally unrelated dependencies. 🤔

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: Oh, right. I agree we should avoid warnings. However, I don’t think we should use the smart constructor so those functions wont be necessary

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: We actually do write it though, it happens in the `withCache` function. That's the reason I had to provide the `EncodeJson` and `DecodeJson` instances. This also means that this needs to remain a `newtype` and that we will provide a manually implemented decoder to properly handle the license field. We can get rid of the `parseBowerFile` function in favour of just `decodeJson` like you mentioned 🙂

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: I got rid of the top level functions and the smart constructor, exposing the constructor instead. I've also implemented the change you suggested here

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: Great I really like that, I've implemented it.

ci/test/Foreign/Jsonic.purs
YanikCeulemans: Great! i've implemented it throughout the tests

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: Ohhh, that’s right. Never mind, we do need it!

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: In that case I'll keep the BowerFile name to avoid having to rename everything if that's ok?

ci/package.json
YanikCeulemans: I ran npm install again and updated the package-lock.json

YanikCeulemans:
I haven't actually tested whether this PR works by running the import script. I'll see if I can get it up and running and report back.

YanikCeulemans:
After testing out the import script I've noticed that a lot of bowerfiles are missing the "version" field. The bowerfile parsing code seems to be working fine though 👍

thomashoneyman:
> After testing out the import script I've noticed that a lot of bowerfiles are missing the "version" field. The bowerfile parsing code seems to be working fine though 👍

You know what -- we don't even use that field, because we use the tag from the associated release for the package. So we actually don't even need a `versions` field in the bower fields that we decode! Just the license, dependencies, and dev dependencies.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
thomashoneyman: Sure, no problem.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -264,8 +268,8 @@ toManifest package repository version (Bower.PackageMeta bowerfile) = do
             Nothing -> Left { dependency: packageName, failedBounds: versionStr }
             Just range -> Right $ Tuple (PackageName.print packageName) range
 
-      normalizedDeps <- normalizeDeps $ un Bower.Dependencies bowerfile.dependencies
-      normalizedDevDeps <- normalizeDeps $ un Bower.Dependencies bowerfile.devDependencies
+      normalizedDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.dependencies
+      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable $ bowerfile.devDependencies
```
thomashoneyman: No need for the trailing `$` here anymore:

```suggestion
      normalizedDeps <- normalizeDeps $ Object.toUnfoldable bowerfile.dependencies
      normalizedDevDeps <- normalizeDeps $ Object.toUnfoldable bowerfile.devDependencies
```

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,63 @@
...
+
+import Control.Alt ((<|>))
+import Data.Argonaut (Json, (.:), (.:?))
+import Data.Argonaut as Json
+import Data.Array as Array
+import Foreign.Jsonic as Jsonic
+
+newtype BowerFile =
+  BowerFile
+    { version :: String
```
thomashoneyman: As mentioned in our discussion we can actually remove this field as well.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,63 @@
...
+    , license :: Array String
+    , dependencies :: Object String
+    , devDependencies :: Object String
+    }
+
+derive newtype instance Show BowerFile
+derive newtype instance Json.EncodeJson BowerFile
```
thomashoneyman, line 25: I'm not entirely sure that this `EncodeJson` instance will correspond with the below manual `DecodeJson` instance. At a glance it looks like it does, but you may need to verify that these files do successfully round-trip encode and decode.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
```diff
@@ -0,0 +1,63 @@
...
+    devDependencies <- fromMaybe mempty <$> obj .:? "devDependencies"
+    pure $ BowerFile { version, license, dependencies, devDependencies }
+
+decodeStringOrStringArray :: Object Json -> String -> Either Json.JsonDecodeError (Array String)
+decodeStringOrStringArray obj fieldName =
+  lmap
+    (const $ Json.AtKey fieldName $ Json.TypeMismatch "String or Array String")
+    do
+      value <- obj .: fieldName
+      (Json.decodeJson value <#> Array.singleton) <|> Json.decodeJson value
```
thomashoneyman: This reads a little oddly for me; would you be open to writing it this way?

```suggestion
decodeStringOrStringArray obj fieldName = do
  let typeError = const $ Json.AtKey fieldName $ Json.TypeMismatch "String or Array String"
  lmap typeError do
    value <- obj .: fieldName
    (Json.decodeJson value <#> Array.singleton) <|> Json.decodeJson value
```

thomashoneyman:
@YanikCeulemans You'll also want to run `purs-tidy` over the files to make sure they're formatted:
https://github.com/purescript/registry/pull/221/checks?check_run_id=3574740092

thomashoneyman:
Once you've removed the `version` field, possibly addressed my two questions / recommendations, and formatted the code so that tests pass, this is a 👍 from me!

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: Yes totally open to it, this is better already.

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: I've added a test for it and it works 👍

ci/src/Registry/Scripts/BowerImport/BowerFile.purs
YanikCeulemans: Okay, I've removed the version field

thomashoneyman (approved):
I'll wait for @f-f to weigh in as well, but 👍 from me. Thanks!

f-f (approved):
This looks excellent, great work! :clap:

Issue #220, opened 2021-09-09
Parse LICENSE file if no license is listed in the Bowerfile


thomashoneyman:
Some packages do not declare a `license` field in their Bowerfiles, despite having a valid SPDX license in the repository root. One example is `aff-coroutines` at its initial released version:

https://github.com/purescript-contrib/purescript-aff-coroutines/tree/v0.1.0

If we don't find a license in the imported Bowerfile (or in the imported `spago.dhall` file if / when #218 is implemented), but there is a `LICENSE` file in the repository root, then we should try to parse that and use it as the package's license in the produced `Manifest`.

thomashoneyman:
This will only work if we can find a suitable library for parsing LICENSE files. Absent that, we should close this issue, because we're certainly not going to expend the effort to write one ourselves!

thomashoneyman:
One potential option is:
https://github.com/juancarlospaco/parselicense

YanikCeulemans:
Looking at the source code for the "parselicense" repo it appears that it actually just runs MD5 on the license content and compares it to a static set of known MD5 hashes. It would be trivial to port it to purescript, i.e. something like the pseudocode `String.trim >>> String.toLowerCase >>> hashWithMD5`.

The downside would be that the known set of MD5 hashes has to be maintained.

YanikCeulemans:
Searching npm, I didn't get a lot of results for a parser for license contents. Most of them just deal with the spdx expression. The few that actually use the license file contents use a regex that matches the start of the license, hardly a sound check.

f-f:
@YanikCeulemans thanks for looking into this! I think an easy approach to this would be:
- implement that MD5 check
- fetch all the licenses that we have to parse, and collect them into bins based on their MD5
- map every bin to its SPDX type
- maintain a set of MD5s here starting from these bins

I think it's acceptable to maintain the list of hashes here, because this import is sort of temporary - hopefully we'll have to run it only a couple of times, and then we can switch over to using the pipeline for new packages.

maxdeviant:
Maybe I'm missing something, but how would we be able to get a consistent MD5 hash across license files if there are variable parts of the license (like the copyright holder's name)?

maxdeviant:
> Maybe I'm missing something, but how would we be able to get a consistent MD5 hash across license files if there are variable parts of the license (like the copyright holder's name)?

This was more a question about the implementation of the aforementioned `parselicense`, because from the implementation I can't tell how it gets consistent hashes.

[...]

PR #227, opened 2021-09-12
Bower Import: Defer license check

thomashoneyman:
In #221, @YanikCeulemans implemented better parsing for Bower JSON files.

One consequence was that packages could fail with a `MalformedBowerJson` error when they're missing a license field. However, we should defer this check until we try to convert a package into a manifest where we already have existing license parsing checks. It's also important to do this to support #220.

This PR updates the `BowerFile` type to produce a `Maybe (NonEmptyArray String)` type for the license key in a bower file instead of the existing `Array String`. It also relaxes the check so that it is acceptable for the license field to be missing.

### DIFF
--- a/ci/test/Main.purs
+++ b/ci/test/Main.purs
@@ -3,6 +3,7 @@ module Test.Main where
 import Registry.Prelude
 
 import Data.Argonaut as Json
+import Data.Array.NonEmpty as NEA
 import Foreign.GitHub (IssueNumber(..))
 import Foreign.Object as Object
 import Foreign.SPDX as SPDX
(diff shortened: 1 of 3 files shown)


ci/test/Main.purs
```diff
@@ -201,16 +202,17 @@ badBowerFiles = do
       BowerFile.parse str `Assert.shouldNotSatisfy` isRight
     failParseBowerFile = failParseBowerFile' <<< Json.stringify
 
-    missingNameFile = Json.encodeJson {}
```
thomashoneyman, line 204: This test isn't necessary, as we don't require that any fields are present now.

thomashoneyman:
With this change, and the existing change in #221, there are now only 4 packages failing with `MalformedBowerJson` errors, and they're all for truly bad JSON reasons.

f-f (approved):
Great work! :clap:

PR #228, opened 2021-09-13
Log failure stats

### LINKED ISSUE #222: Print out statistics on package failures in Bower import
The Bower import tool converts Bowerfiles for packages in the Bower registry into valid `Manifest`s. However, many packages fail along the way and are collected into a `PackageFailures` map:

https://github.com/purescript/registry/blob/3f5c48e518f657ab03c4178df976648c9fb1437f/ci/src/Registry/Scripts/BowerImport/Error.purs#L15

Later, these failures are written out to a file called `bower-exclusions.json`:
https://github.com/purescript/registry/blob/3f5c48e518f657ab03c4178df976648c9fb1437f/ci/src/Registry/Scripts/BowerImport.purs#L167-L169

However, we also would benefit from knowing some statistics about the package failures, which could be written out to a file or just printed to the console. As a first cut, we'd specifically like to know:

* How many packages succeeded in total? How many packages failed in total?
* How many package versions succeeded in total? How many package versions failed in total?
* How many incidences are there of each error (`MissingLicense`, `BadDependencyVersions`, etc.), in sorted order (ie. most failures are at the top of the list, least failures are at the bottom of the list).

These statistics can help us prioritize what issues to fix and report on how many packages are being omitted from the PureScript registry and for what reasons.

ntwilson:
Fixes #222.
You can scroll to the very bottom of the review to see the example output that would be logged to the console.


ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -39,6 +39,7 @@ newtype ImportErrorKey = ImportErrorKey String
 derive instance Newtype ImportErrorKey _
 derive newtype instance Eq ImportErrorKey
 derive newtype instance Ord ImportErrorKey
+instance Show ImportErrorKey where show (ImportErrorKey key) = "(ImportErrorKey " <> key <> ")"
```
ntwilson: I needed this for testing, but I can probably work around it if we don't want this show instance here

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+      in Set.size packages
+
+    countOfVersionFailures = 
+      sum $ map Map.size $ versionFailuresFromPackageFailures pFailures
+
+    countImportErrorsByErrorType = 
+      let 
+        countFailuresForPackage :: Either ImportError (Map RawVersion ImportError) -> Int
+        countFailuresForPackage = either (const 1) Map.size
+      in (sum <<< map countFailuresForPackage) <$> failures
```
ntwilson: The counts for the top level errors (this one included) are the counts for versions that failed _plus_ the count of packages that failed to produce any versions (we have that `Either ImportError (Map RawVersion ImportError)`.  I suppose we could split it into two numbers per type of error.

ci/src/Registry/Scripts/BowerImport/Error.purs
thomashoneyman: If we don't need it anymore, then I'd prefer to drop it. I wish the `Debug` class had become a thing in PureScript, but alas 😬

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -105,8 +109,10 @@ instance Json.EncodeJson ManifestError where
 
 instance Json.DecodeJson ManifestError where
   decodeJson = Json.Decode.Generic.genericDecodeJsonWith encodingOptions
+  
+type ManifestErrorKey = String
```
thomashoneyman: If we're going to introduce a specific type for this, I'd prefer we make it a newtype so that it does have a bit of type safety at least (when comparing it to other strings)

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+  , logStats
+  , Stats
+  , SucceededPackages
+  ) where
+
+import Registry.Prelude
+
+import Control.Monad.Writer (execWriter, tell)
+import Data.Array as Array
+import Data.Foldable (fold, intercalate, sum)
```
thomashoneyman: I think we should add `fold` to the `Registry.Prelude`.

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+import Data.Function (on)
+import Data.Interpolate (i)
+import Data.List as List
+import Data.List.NonEmpty as NonEmptyList
+import Data.Map (SemigroupMap)
+import Data.Map as Map
+import Data.Ordering (invert)
+import Data.Set as Set
```
thomashoneyman: We've been following a policy where we import functions qualified wherever possible; ie. we'd have things like:

```purs
import Control.Monad.Writer as Writer
import Data.Foldable as Foldable
import Data.Function as Function
import Data.Ordering as Ordering
```

I think importing `i` from `Interpolate` is fine, as is the import of the `SemigroupMap` type. However, @f-f might feel differently about these suggestions, and I'll defer to him!

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+  , countOfVersionSuccesses :: Int
+  , countOfVersionFailures :: Int
+  , countImportErrorsByErrorType :: Map ImportErrorKey Int
+  , countManifestErrorsByErrorType :: Map ManifestErrorKey Int
+  }
+
+type SucceededPackages package version a = Map package (Map version a)
+
+errorStats :: forall package version a. SucceededPackages package version a -> PackageFailures -> Stats
+errorStats succeededPackages pFailures@(PackageFailures failures) =
```
thomashoneyman:
```suggestion
errorStats succeededPackages packageFailures@(PackageFailures failures) = 
```

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+  , countOfPackageFailures :: Int
+  , countOfVersionSuccesses :: Int
+  , countOfVersionFailures :: Int
+  , countImportErrorsByErrorType :: Map ImportErrorKey Int
+  , countManifestErrorsByErrorType :: Map ManifestErrorKey Int
+  }
+
+type SucceededPackages package version a = Map package (Map version a)
+
+errorStats :: forall package version a. SucceededPackages package version a -> PackageFailures -> Stats
```
thomashoneyman: I wonder if it would make sense to reuse the `ProcessedPackages` type here?

```purs
type ProcessedPackages k a =
  { failures :: PackageFailures
  , packages :: Map k a
  }

type ProcessedPackageVersions k1 k2 a = ProcessedPackages k1 (Map k2 a)
```

in which case we'd have:

```purs
errorStats :: forall package version a. ProcessedPackageVersions package version a -> Stats
```

If you wanted to do that I'd be OK with just moving those types into this file, and then we can find a more permanent home for the types later on.

thomashoneyman: That way we can call `errorStats` directly on the output of the processing steps in `BowerImport` and it matches with their expected type.

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+      # Array.sortBy (invertCompare `on` snd)
+      <#> (intercalate "\n" <<< execWriter <<< case _ of 
+        (key@(ImportErrorKey keyStr) /\ errCount) 
+          | key == manifestErrorKey -> do
+            tell [ i"  "keyStr": "(show errCount)" (total packages/versions)" ]
+            for_ (Map.toUnfoldable stats.countManifestErrorsByErrorType # Array.sortBy (invertCompare `on` snd)) \(err /\ count) -> do
+              tell [ i"    "err": "(show count) ]
+          | otherwise -> 
+            tell [ i"  "keyStr": "(show errCount) ]
+      )
```
thomashoneyman: I'm finding this a bit difficult to read. Can we break it into distinct functions to avoid mixing operators like `#`, `<#>`, `<<<`, and `case _ of` in the same function?

For example, perhaps we get functions like the below (I've switched some operators, but you don't have to -- this is just an example of making it a little easier for me to read the flow).

```purs
sortedErrors = map writeError $ compareValues stats.countImportErrorsByErrorType

compareValues = Array.sortBy (flip compare `on` snd) <<< Map.toUnfoldable

writeError = intercalate "\n" <<< execWriter <<< case _ of
  Tuple key@(ImportErrorKey keyStr) errorCount
    | key == ...
```

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,125 @@
...
+      Just manifestErrors -> 
+        let 
+          allManifestErrors :: List ManifestError
+          allManifestErrors = 
+            Map.values manifestErrors
+            >>= either List.singleton List.fromFoldable
+            # List.mapMaybe (case _ of 
+              ManifestError errs -> Just $ List.fromFoldable errs
+              _ -> Nothing)
+            # join
```
thomashoneyman: Same comment here as below wrt mixing operators -- I find it kind of hard to read. Could we break these into distinct functions so that things like `>>=` and `#` aren't being mixed?

thomashoneyman:
Thanks for adding this, @ntwilson! It's a great start! I've added a couple of comments that I think will help me understand the code a little better, but on the whole I don't think this will require much tweaking before we go ahead and merge.

You'll need to run `purs-tidy` over your code, by the way, so that CI passes.

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
ntwilson: Yeah that sounds nice to me. It felt a little bad redefining this map structure.

thomashoneyman: You could also specialize it for `errorStats`, if we know we have a specific structure. For example:

```purs
errorStats :: forall a. ProcessedPackageVersions PackageName SemVer a -> Stats
```

We don't actually use the information, so it's totally fine to leave those variables open too.

ci/test/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -15,84 +15,93 @@ import Test.Spec.Assertions as Assert
...
 
-comparableMap :: forall k n v. Newtype k n => Ord n => Show n => Map k v -> Map n v 
-comparableMap originalMap = 
-  let kvPairs = (Map.toUnfoldable originalMap :: Array _) <#> \(key /\ val) -> (Newtype.unwrap key /\ val)
-  in Map.fromFoldable kvPairs
-
-assertMapsEqual :: forall k n v
-  . Show n => Show v => Eq n => Eq v => Newtype k n => Ord n 
-  => Map k v -> Map k v -> Aff Unit
+comparableMap :: forall k n v. Newtype k n => Ord n => Show n => Map k v -> Map n v
```
thomashoneyman: I’m not sure I’m following why this is necessary. Why can’t we compare maps without this function?

ntwilson: We can't assert that Maps are equal without this function because their keys  (`ImportErrorKey` & `ManifestErrorKey`) lack `Show`.  I could just Assert that `==` is `true`, but you'd get no information as to why it failed if there was a problem.  This just alters the keys to plain strings so we can assert them and get reasonable assertion failure output messages.

thomashoneyman: Oh, I see. I misunderstood your comment about the Show instances before — if we’re using the instances in tests then feel free to add them back for those new types; I think that’s better than adding this function in.

ci/src/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,135 @@
...
+import Control.Monad.Writer as Writer
+import Data.Array as Array
+import Data.Foldable as Foldable
+import Data.Function (on)
+import Data.Interpolate (i)
+import Data.List as List
+import Data.List.NonEmpty as NonEmptyList
+import Data.Map (SemigroupMap)
+import Data.Map as Map
+import Data.Ordering as Ordering
```
thomashoneyman: I'm getting an unused import warning:

```suggestion
```

ci/test/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,133 @@
+module Test.Registry.Scripts.BowerImport.Error.Stats where
+
+import Registry.Prelude
+
+import Data.Array.NonEmpty as NonEmptyArray
+import Data.Foldable as Foldable
+import Data.Function (on)
```
thomashoneyman: I'm getting an unused import warning:

```suggestion
```

ci/test/Registry/Scripts/BowerImport/Error/Stats.purs
```diff
@@ -0,0 +1,133 @@
...
+import Data.Map as Map
+import Data.Newtype as Newtype
```
thomashoneyman: I'm getting an unused import warning:

```suggestion
```

thomashoneyman:
I'm seeing these results:

```
Number of successful packages: 1328
Number of failed packages: 484
Number of successful versions: 9279
Number of failed versions: 1719
Failures by error:
  manifestError: 1240 (total packages/versions)
    missingLicense: 928
    badDependencyVersions: 281
    badVersion: 48
    badLicense: 35
    invalidDependencyNames: 7
  nonRegistryDependencies: 242
  missingBowerfile: 231
  noReleases: 33
  malformedBowerJson: 5
  malformedPackageName: 1
```

I searched these results in the `bower-exclusions.json` file and the numbers appear to check out. I'm curious to dig in and see if we can bring some of these numbers down.

thomashoneyman (approved):

ntwilson:
> I'm seeing these results:
> 
> ```
> Number of successful packages: 1328
> Number of failed packages: 484
> Number of successful versions: 9279
> Number of failed versions: 1719
> Failures by error:
>   manifestError: 1240 (total packages/versions)
>     missingLicense: 928
>     badDependencyVersions: 281
>     badVersion: 48
>     badLicense: 35
>     invalidDependencyNames: 7
>   nonRegistryDependencies: 242
>   missingBowerfile: 231
>   noReleases: 33
>   malformedBowerJson: 5
>   malformedPackageName: 1
> ```
> 
> I searched these results in the `bower-exclusions.json` file and the numbers appear to check out. I'm curious to dig in and see if we can bring some of these numbers down.

This is showing the total occurrences of each error.  Is it worth it to show for each error: 
1. Total occurrences
2. How many packages this error affects
3. How many versions this error affects
?  

And would this question have the same answer (or even apply) to both ImportErrors and ManifestErrors (do manifest errors also potentially apply either to an entire package or to just a particular version)?

thomashoneyman:
I think that would definitely be worth showing. Manifest errors only happen for a particular version, and some other errors only happen for packages (malformed name, malformed bower file, no releases), but it’s ok to just show 0 for them for not affecting any versions.

ntwilson:
> I think that would definitely be worth showing. Manifest errors only happen for a particular version, and some other errors only happen for packages (malformed name, malformed bower file, no releases), but it’s ok to just show 0 for them for not affecting any versions.

@thomashoneyman, would you mind reviewing [this](https://github.com/ntwilson/registry/pull/1) to see if you think this is a good idea and worth including in this PR?  It's kind of a lot of changes, so I didn't want to commit it to this branch unless the idea checks out.

thomashoneyman:
> @thomashoneyman, would you mind reviewing [this](https://github.com/ntwilson/registry/pull/1) to see if you think this is a good idea and worth including in this PR? It's kind of a lot of changes, so I didn't want to commit it to this branch unless the idea checks out.

I think that's a great addition! I would prefer to merge this PR first, and then look at improving it further.

f-f (approved):
This looks great, thank you! 👏

thomashoneyman:
@ntwilson Feel free to open a new PR against the default branch here now that this is in!

PR #230, opened 2021-09-15
Flatten BowerImport directories

thomashoneyman:
This PR performs a small cleanup of the `BowerImport` modules to make them a little easier to work with (in my opinion). If y'all disagree, then we can close this -- it doesn't change functionality at all.

Specifically, this PR:

1. Renames `BowerImport.Error.Stats` to `BowerImport.Stats`, because this file reports on successes and on failures, and we may expand on this module in the future to include other statistics on the script without necessarily being anything to do with errors specifically.
2. Creates a new `BowerImport.Process` module to hold our processing-specific functions -- ie. `forPackage`, `withCache`, `forPackageVersions`, etc. -- as well as the related types, ie. `ProcessedPackages` and `ProcessedPackageVersions`.
3. Moves all processing-specific functions out of the main `BowerImport` module and into the `BowerImport.Process` module.
4. Moves the processing-specific types out of the `BowerImport.Stats` module (where they don't really make any sense) and into the `BowerImport.Process` module

No implementations have been touched, things have just been reorganized a little bit to make the names clearer and to make the main `BowerImport` module be completely focused on the processing steps themselves and not the helper functions around them.

### DIFF
--- a/ci/src/Registry/Scripts/BowerImport.purs
+++ b/ci/src/Registry/Scripts/BowerImport.purs
@@ -5,42 +5,34 @@ import Registry.Prelude
 import Affjax as Http
 import Affjax.ResponseFormat as ResponseFormat
 import Affjax.StatusCode (StatusCode(..))
-import Control.Apply (lift2)
 import Control.Monad.Except as Except
-import Control.Monad.State as State
 import Data.Argonaut as Json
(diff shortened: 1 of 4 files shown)


f-f (approved):
This is great!

PR #232, opened 2021-09-16
Add `.env` file support for `BowerImport`

maxdeviant:
This PR adds support for specifying environment variables in a `.env` file to be used by the `BowerImport` script.

As part of this we defer the instantiation of Octokit so that it doesn't happen as soon as the `Foreign.GitHub` module is imported (and thus before `Dotenv.loadFile` can execute).

A side benefit of this is that the `Foreign.GitHub` module can now be imported (for example, for its types) without needing to have a `GITHUB_TOKEN` environment variable set.


ci/src/Foreign/GitHub.purs
```diff
@@ -10,12 +10,16 @@ import Data.Argonaut as Json
...
 import Data.String as String
 import Data.String.CodeUnits (fromCharArray)
-import Effect.Uncurried (EffectFn1, EffectFn2, EffectFn3, runEffectFn1, runEffectFn2, runEffectFn3)
+import Effect.Uncurried (EffectFn2, EffectFn3, EffectFn4, runEffectFn2, runEffectFn3, runEffectFn4)
 import Safe.Coerce (coerce)
 import Text.Parsing.StringParser as Parser
 import Text.Parsing.StringParser.CodePoints as Parse
 import Text.Parsing.StringParser.Combinators as ParseC
 
+foreign import data Octokit :: Type
```
maxdeviant, line 19: I debated naming this something more generic, like `GitHubClient`. Happy to rename, if desired.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -44,6 +45,8 @@ import Text.Parsing.StringParser as StringParser
 -- | - go through this list: if the package is in the registry index then skip, otherwise upload
 main :: Effect Unit
 main = Aff.launchAff_ do
+  _ <- Dotenv.loadFile
```
maxdeviant, line 48: This will load the `.env` file, if it exists. If it doesn't exist, this is a no-op.

thomashoneyman (approved):

thomashoneyman:
I wonder if we should also add like a .env.example that has the env vars we need already set to dummy values, for new contributors?

Or a short section in a CI directory README?

thomashoneyman:
What do you think of adding a file named `.env.example` to the `ci` directory root that has these contents?

```
GITHUB_TOKEN=
SPACES_KEY=""
SPACES_SECRET=""
```

ci/src/Foreign/GitHub.purs
thomashoneyman: I think this is fine 👍

maxdeviant:
> I wonder if we should also add like a .env.example that has the env vars we need already set to dummy values, for new contributors?
> 
> Or a short section in a CI directory README?

Definitely! Both of these would have been useful for me as a newcomer to the project 😄 

I added a short README and a `.env.example` file. I can also add the steps to generate/retrieve the environment variables, but I would need some assistance in knowing what those steps would be.

For reference, when I generated my GitHub token I wasn't sure what permissions it needed, so I just gave it access to the `public_repo` scope. Are there any other scopes it needs?

I also don't know where one would get either of the `SPACES_` environment variables. I see that they correspond to S3 credentials, so I imagine access to those might be limited?

thomashoneyman:
> For reference, when I generated my GitHub token I wasn't sure what permissions it needed, so I just gave it access to the public_repo scope. Are there any other scopes it needs?

It's only for fetching releases from repositories without rate limiting, so I don't know if it needs any scopes at all!

> I also don't know where one would get either of the SPACES_ environment variables. I see that they correspond to S3 credentials, so I imagine access to those might be limited?

Yes, access is limited. Some things require the variables to exist so empty strings is fine.

f-f (approved):
This is great, thank you!

f-f:
I see no "update branch" in here, so maybe this needs to be rebased locally?

maxdeviant:
> I see no "update branch" in here, so maybe this needs to be rebased locally?

Weird, I see the option on my end. Maybe because the branch is on my fork?

Anyways, I just updated, so should be good to merge.

PR #234, opened 2021-09-16
Fall back to spago files if there is no bower file in a repository

### LINKED ISSUE #218: Fall back to spago.dhall files for packages without a bower.json
Some packages are in the PureScript registry and do not have `bower.json` files, but _do_ have `spago.dhall` files sufficient for producing a valid `Manifest`. One example is `halogen-hooks`:

https://github.com/thomashoneyman/purescript-halogen-hooks/tree/v0.5.0

This repository doesn't have a `bower.json` file, but it does have a `spago.dhall` file that is sufficient for producing a manifest:

https://github.com/thomashoneyman/purescript-halogen-hooks/blob/v0.5.0/spago.dhall

I'm proposing that, when we find a package that has a Spago file but not a Bower file, we attempt to either:

* parse the `spago.dhall` file to get the package license and dependencies
* generate a `bower.json` file from the `spago.dhall` file, and then import the Bower file as usual

thomashoneyman:
In #218 we saw that 231 packages are failing with a `MissingBowerfile` error. A large number of these packages do have `spago.dhall` file that can be used to produce a valid Bowerfile, which we can then parse. This PR implements a Spago fallback for these packages.

Before this change:

```
Number of successful packages: 1328
Number of failed packages: 484
Number of successful versions: 9279
Number of failed versions: 1719

Failures by error:
  manifestError: 1240 (total packages/versions)
    missingLicense: 928
    badDependencyVersions: 281
    badVersion: 48
    badLicense: 35
    invalidDependencyNames: 7
  nonRegistryDependencies: 242
  missingBowerfile: 231
  noReleases: 33
  malformedBowerJson: 5
  malformedPackageName: 1
```

After adding the Spago fallback our import results are quite different! The example package `halogen-hooks` now succeeds:

```
Attempting to use spago.dhall file for package: halogen-hooks
Successfully fell back to spago file!
```

Many other packages no longer fail with `MissingBowerfile` errors because we are able to fall back to a Spago file; that Spago file often doesn't have any listed LICENSE, so we're able to get the package dependencies out but we fail with a `BadLicense` error later on instead of a `MissingBowerfile` error. If we were able to parse the LICENSE from elsewhere then we _could_ successfully construct a manifest for this package, so the new error is more appropriate.

```
Number of successful packages: 1406
Number of failed packages: 474
Number of successful versions: 9809
Number of failed versions: 1694
Failures by error:
  manifestError: 1378 (total packages/versions)
    missingLicense: 928
    badDependencyVersions: 293
    badLicense: 179
    badVersion: 48
  nonRegistryDependencies: 246
  missingBowerfile: 64
  noReleases: 32
  malformedBowerJson: 5
  malformedPackageName: 1
```

These numbers don't totally add up, and I haven't checked too deeply why that is. There are many more successful packages, but only 10 fewer failed packages -- I thought they'd move together, but perhaps not! Regardless, it looks like we pick up about 80 new packages and hundreds of new package versions.

### Details

As of 2021-09-15, there are 231 packages failing the Bower import with a `MissingBowerfile` error. However, we know at least _some_ of these packages (`halogen-hooks`, for example) have valid `spago.dhall` files even though they don't have valid `bower.json` files.

When a package has no `bower.json` file, we should check whether it has a `spago.dhall` file. If it doesn't, we should fail as usual. If it does, we should generate a `bower.json` file from the package -- if possible -- and then parse that Bowerfile.

This implementation works by:

1. Look for a `bower.json` file in the repository. If it's there, parse it. If not, continue.
2. Look for a `spago.dhall` and `packages.dhall` file in the repo. If they're there, then attempt to clean up the `spago.dhall` file by inserting the repository field (which we already know) and an empty license field (if there is none). If they are missing, fail with a `MissingBowerfile` error.
3. Use the fixed `spago.dhall` and `packages.dhall` file to generate a new `bower.json` file via the `spago bump-version minor --no-dry-run` command.
4. Return the resulting `bower.json` file, if possible.

#### Packages that depend on `assert`

A handful of packages _possibly_ could be used to generate bower files but don't because of issues parsing `assert` from the package set:

```
Attempting to use spago.dhall file for package: bouzuya-http-status-code
Could not generate bower.json file:
Command failed: spago -x spago-fix-metadata-version.dhall bump-version minor --no-dry-run
spago: 
↳ ./spago-fix-metadata-version.dhall
  ↳ ./spago.dhall
    ↳ ./packages.dhall
      ↳ https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190330/src/packages.dhall sha256:cb0cdde5926cfdff5bd17bb2508a85b5eee794088f253f59f884766828ba722c
        ↳ https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190330/src/groups/purescript.dhall

Error: Invalid input

https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190330/src/groups/purescript.dhall:20:7:
   |
20 |     , assert =
   |       ^
unexpected 'a'
expecting whitespace or }


2│         ./groups/purescript.dhall

https://raw.githubusercontent.com/purescript/package-sets/psc-0.12.3-20190330/src/packages.dhall:2:9
```


ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -195,18 +215,80 @@ fetchBowerfile name address tag = do
...
+              -- empty license for the sake of generating the file. The package
+              -- will fail later with `MissingLicense`.
+              fixupSpagoDhall = fixupRepo <<< fixupLicense
+                where
+                fixupRepo = fixupDhall "repository =" (", repository = \"" <> repoStr <> "\"")
+                fixupLicense = fixupDhall "license =" ", license = \"\""
+
+            write "spago.dhall" (fixupSpagoDhall spagoDhall)
+            write "packages.dhall" packagesDhall
+            write fixMetadataVersionPath "./spago.dhall with metadata.version = \"v0.14.0\""
```
JordanMartinez: Two comments here. First, I made a mistake. When updating the `metadata.version`, the update typically goes in the `packages.dhall` file, not the `spago.dhall` file. I was going to propose we change this expression by prefixing it with `packages`, but it turns out that that doesn't work. Spago assumes there is a `packages.dhall` file, and checks that file, not the `packages` key in the `spago.dhall` file's expression. So, we will actually need to append the following text to the end of the `packages.dhall` file: 
```diff
write "packages.dhall" packagesDhall
+ liftAff $ FS.appendTextFile UTF8 (tmp <> "/" <> path) "    with metadata.version = \"v0.14.0\""
```

Second, if we're already using a custom `spago.dhall` file other than the main one, why go through all the effort above via the `fixDhall` function in editing the `spago.dhall` and why not just override the values in our custom one?

In other words
```purescript
write fixMetadataVersionPath =
  "{ license = \"\", repository = \"" <> repoStr <> "\" } // ./spago.dhall"
```
In other words, guarantee that the spago.dhall file will have a (blank) license key and a repository key but fall back to the real value if `spago.dhall` has one for that key.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -195,18 +215,80 @@ fetchBowerfile name address tag = do
...
+
+            let
+              fixupDhall match val src = fromMaybe src do
+                Alternative.guard $ not String.contains (String.Pattern match) src
+                let lines = String.split (String.Pattern "\n") src
+                ix <- Array.elemIndex "}" lines
+                updated <- Array.insertAt ix val lines
+                pure $ String.joinWith "\n" updated
+
+            tmp <- liftEffect Tmp.mkTmpDir
```
JordanMartinez, line 238: Since we're still running this script locally and not in CI, should we clean up this temporary directory when we're finished using it?

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -195,18 +215,80 @@ fetchBowerfile name address tag = do
...
+                fixupLicense = fixupDhall "license =" ", license = \"\""
+
+            write "spago.dhall" (fixupSpagoDhall spagoDhall)
+            write "packages.dhall" packagesDhall
+            write fixMetadataVersionPath "./spago.dhall with metadata.version = \"v0.14.0\""
+
+            let
+              generateFile = liftEffect $ map (lmap Aff.message) do
+                _ <- ChildProcess.execSync "git init && git add -A && git commit -m 'yeesh'" (ChildProcess.defaultExecSyncOptions { cwd = Just tmp })
+                let bumpCommand = fold [ "spago -x ", fixMetadataVersionPath, " bump-version minor --no-dry-run" ]
```
JordanMartinez: If we used `i` from `interpolate` above, why not use it here, too?

JordanMartinez (changes requested):
I think there's a better way we could handle the license key in the spago file. Also, the metadata.version fix I used doesn't work because it needs to be in the `packages.dhall` file.

The other comments are minor.

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: > So, we will actually need to append the following text to the end of the packages.dhall file:

I'm fine with changing it, but from running the code things appear to work right now -- are you noticing this failing on your end?

> Why not just override the values in our custom one?

Good point, we totally could just override in our custom one. Thanks for the suggestion!

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: We could remove it, just for the sake of being explicit, though the temporary directory _should_ be removed when the program exits:

https://www.npmjs.com/package/tmp#graceful-cleanup

ci/src/Registry/Scripts/BowerImport.purs
thomashoneyman: We certainly could use it here!

thomashoneyman:
@JordanMartinez Thanks for the feedback! I don't have any more time to work on this today, and perhaps not until Monday, so if you have any interest in applying your suggestions directly to the branch I am totally fine with that -- I agree with them. Otherwise, I'll get to this either tomorrow or Monday.

ci/src/Registry/Scripts/BowerImport.purs
JordanMartinez: > are you noticing this failing on your end?

I only noticed it when I ran `spago ls packages | grep 'meta'` and saw that my change here didn't affect that command's output. I think it's safer if we modify the `packages.dhall` file, but I don't think I've seen a "minimum package set version" error like I saw before. So, I guess it's not as needed.

JordanMartinez:
> if you have any interest in applying your suggestions directly to the branch I am totally fine with that -- I agree with them. Otherwise, I'll get to this either tomorrow or Monday.

@thomashoneyman Ok, I'll try to finish this up today then.

JordanMartinez:
After my changes (and with the updates to the stats printer), we now get the following:
```
Number of successful packages: 1330
Number of failed packages: 482
Number of successful versions: 9291
Number of failed versions: 1707
Failures by error:
  manifestError: 1229 occurrences (339 packages / 1229 versions)
    missingLicense: 928 occurrences (241 packages / 928 versions)
    badDependencyVersions: 264 occurrences (102 packages / 264 versions)
    badVersion: 48 occurrences (32 packages / 48 versions)
    badLicense: 35 occurrences (6 packages / 35 versions)
  nonRegistryDependencies: 242 occurrences (45 packages / 242 versions)
  missingBowerfile: 230 occurrences (83 packages / 230 versions)
  noReleases: 33 occurrences (33 packages / 0 versions)
  malformedBowerJson: 5 occurrences (4 packages / 5 versions)
```

thomashoneyman:
That's a little odd -- looks like a _lot_ of versions got dropped in the latest run, like 600?

JordanMartinez:
Which number specifically are you referring to?

Perhaps Spago can't properly handle the custom `spago.dhall` file? I should test out these edits locally and see what happens.

thomashoneyman:
> Which number specifically are you referring to?
> 
> Perhaps Spago can't properly handle the custom `spago.dhall` file? I should test out these edits locally and see what happens.

I’m looking at the total “number of successful versions” from one run to the next

f-f:
As we discussed last week, I think a more reliable solution to convert spago files would be to not use `spago` at all - we are already using `dhall-to-json` in our code, so we could import the `spago.dhall` we want as a remote import to `dhall-to-json`, and that would give us a JSON document from which we could extract dependencies and versions for the package.

thomashoneyman:
Superseded by #240

PR #235, opened 2021-09-17
Tiny fixups to test and env code

thomashoneyman:
This PR makes minor tweaks to the existing test code and environment variables, just to clean up a couple things I noticed. Without these changes you can't make a `.env` file and run `spago test`, for example, without exceptions being thrown.


ci/.env.example
```diff
@@ -1,3 +1,3 @@
 GITHUB_TOKEN=
-SPACES_KEY=""
-SPACES_SECRET=""
+SPACES_KEY=
+SPACES_SECRET=
```
thomashoneyman, line 3: `Dotenv` throws an exception if these are set to empty strings 🤷

ci/src/Foreign/S3.js
```diff
@@ -1,42 +1,41 @@
-const AWS = require('aws-sdk');
+const AWS = require("aws-sdk");
 
-if ("SPACES_KEY" in process.env && "SPACES_SECRET" in process.env) {
-  ;
-} else {
-  console.log('Please set SPACES_KEY and SPACES_SECRET envvars');
-  process.exit(1);
-}
+exports.connectImpl = function (endpoint) {
```
thomashoneyman, line 3: Without this change, any import that ultimately relies on code in this file (even if it isn't executed) will crash.

ci/src/Foreign/S3.js
```diff
@@ -1,42 +1,41 @@
...
       });
     });
   };
 };
 
-exports.putObjectImpl = function(s3, params) {
-  return function() {
-    return new Promise(function(resolve, reject) {
-      s3.putObject(params, function(err, data) {
+exports.putObjectImpl = function (s3, params) {
```
thomashoneyman, line 35: These changes are from the standard Prettier JS configuration. I can try and go back and revert this if desired, but it's what we use throughout the Contrib libraries also.

ci/src/Registry/Scripts/BowerImport/Process.purs
```diff
@@ -63,7 +63,8 @@ forPackageVersion
   -> (k1 -> k2 -> a -> ExceptT ImportError Aff b)
   -> Aff (ProcessedPackageVersions k1 k2 b)
 forPackageVersion input keyToPackageName keyToTag f = do
-  map snd $ State.runStateT iterate { failures: input.failures, packages: Map.empty }
+  Tuple _ processed <- State.runStateT iterate { failures: input.failures, packages: Map.empty }
+  pure $ processed { packages = Map.filter (not Map.isEmpty) processed.packages }
```
thomashoneyman, line 67: This didn't change anything in my test of the code, but it's technically possible for a package to have all its versions removed but still exist in the map. This ensures that packages that no longer have any valid versions get removed.

ci/test/Main.purs
```diff
@@ -15,14 +15,12 @@ import Registry.Scripts.BowerImport.BowerFile (BowerFile(..))
...
 import Test.Fixtures.Manifest as Fixtures
 import Test.Foreign.Jsonic (jsonic)
-import Test.Registry.Scripts.BowerImport.Error.Stats (errorStats)
+import Test.Registry.Scripts.BowerImport.Stats (errorStats)
 import Test.Spec as Spec
 import Test.Spec.Assertions as Assert
 import Test.Spec.Reporter.Console (consoleReporter)
 import Test.Spec.Runner (runSpec)
 
-type Spec = Spec.SpecT Aff Unit Identity Unit
```
thomashoneyman, line 24: I've removed this since it's the same as the `Spec` type from the `Spec` library, just applied to `Unit`.

Without this change, if we want to use the same `Spec` type in all test files, then we're going to have to make a `Test.Types` module to hold this type and avoid circular imports. I thought it would be easier to understand if we just used the `Spec` type provided by the library.

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -1,4 +1,4 @@
-module Test.Registry.Scripts.BowerImport.Error.Stats where
+module Test.Registry.Scripts.BowerImport.Stats where
```
thomashoneyman, line 1: Simple rename of the file, since the same was done to the original file.

f-f (approved):
Looks great!

Issue #237, opened 2021-09-17
Record total packages and versions in stats


thomashoneyman:
Currently, the `Stats` module outputs statistics with this format:

```json
Number of successful packages: 1369
Number of failed packages: 474
Number of successful versions: 9565
Number of failed versions: 1694
Failures by error:
  manifestError: 1378 occurrences (371 packages / 1378 versions)
    missingLicense: 928 occurrences (241 packages / 928 versions)
    badDependencyVersions: 283 occurrences (104 packages / 283 versions)
    badLicense: 179 occurrences (45 packages / 179 versions)
    badVersion: 48 occurrences (32 packages / 48 versions)
  nonRegistryDependencies: 246 occurrences (47 packages / 246 versions)
  missingBowerfile: 64 occurrences (34 packages / 64 versions)
  noReleases: 32 occurrences (32 packages / 0 versions)
  malformedBowerJson: 5 occurrences (4 packages / 5 versions)
  malformedPackageName: 1 occurrences (1 packages / 1 versions)
```

However, it would also be useful to know the _total_ number of packages and versions we attempted to work with. For example, these statistics list 1,369 "successful" packages and 474 "failed" packages, but we only have something like 1500-1600 packages in total; some packages are present in both the success and failure maps because _some_, but not all, of their versions succeeded.

I think the output of the first few lines ought to be more like:

```
Packages: 1500 total (1369 with successes, 474 with failures)
Versions: 11000 total (9565 successful, 1694 failed)
```

where the total packages won't be a sum of its constituents because some packages will be in both maps, but where the total _versions_ should be the sum of its constituents because a version can't be failed _and_ successful.

We could also in the future be more granular and segment packages with all successes, all failures, and a mixture of both, but that's a larger task.

thomashoneyman:
We can get the total number of packages with a `Map.size` of this value:

https://github.com/purescript/registry/blob/28ffb8412ccd09130a81c1aecc27c2a59b7b6674/ci/src/Registry/Scripts/BowerImport.purs#L78

and we have all releases that we're going to try and fetch in the `releaseIndex` -- this corresponds to "versions" we're going to attempt to parse:

https://github.com/purescript/registry/blob/28ffb8412ccd09130a81c1aecc27c2a59b7b6674/ci/src/Registry/Scripts/BowerImport.purs#L94

thomashoneyman:
[...]

PR #239, opened 2021-09-22
Makes application-y functions monomorphic

mikesol:
As per https://github.com/purescript/registry/issues/237.

I implemented this without an understanding of the underlying mechanics of the code, so it may be off. Also, it is an anti-pattern to pass the totals to the stats function only to be spit back out, but it was the quickest way from A to Z.


ci/package.json
```diff
@@ -21,5 +21,10 @@
     "tar": "^6.1.9",
     "tmp": "^0.2.1",
     "xhr2": "^0.2.0"
+  },
+  "devDependencies": {
```
thomashoneyman: These are also made available via the nix shell, just FYI! I don’t know if we want to standardize on one or the other, though licensee is only available via the shell.

thomashoneyman: On second thought, I think we shouldn't add these to the package.json file because they are already included in the Nix shell, as well as other dependencies that can't be made available via NPM (namely `licensee` and `dhall-json`).

mikesol: Sounds good! I'll use nix-shell from here on out.

f-f:
I was wondering: why so many changes to the package-lock?

thomashoneyman:
> I was wondering: why so many changes to the package-lock?

It’s from the added dev dependencies — though I think we should remove them in favor of the Nix shell.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -354,7 +355,7 @@ cleanPackageName (RawPackageName name) = do
     _ -> throwError $ MalformedPackageName name
 
 -- | Read the list of packages in a registry file
-readRegistryFile :: FilePath -> Aff (Map RawPackageName String)
+readRegistryFile :: FilePath -> Aff (Map RawPackageName PackageURL)
```
thomashoneyman: As a quick note -- before you go any further, you might want to merge with the default branch, because these functions have changed and you'll have conflicts.

mikesol: Just saw that, thanks for the heads up!

mikesol:
@thomashoneyman I've changed the scope of this PR to only getting rid of polymorphism in application code.

mikesol:
@thomashoneyman I think this is ready for review & potentially merge as-is given the reduced scope. With the monomorphic functions in there, the mapping you talk about is more clear from the types & it'll be easier to get cracking on the stats.

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -29,24 +31,23 @@ type ProcessedPackages k a =
...
 forPackage
-  :: forall k1 k2 a b
-   . Ord k1
-  => Ord k2
-  => ProcessedPackages k1 a
-  -> (k1 -> RawPackageName)
-  -> (k1 -> a -> ExceptT ImportError Aff (Tuple k2 b))
-  -> Aff (ProcessedPackages k2 b)
-forPackage input keyToPackageName f = do
+  :: ProcessedPackages RawPackageName PackageURL
```
thomashoneyman, line 38: These `for*` functions were originally used in a couple places, which made it so that you couldn't have a concrete type for them. But if they're now this easy to specialize to specific types, then I wonder if we should just inline them in `LegacyImport.purs`? If they're not being reused at all then they don't really need to be in this file.

If you're interested, maybe that's something we could pair on or something I could do in a quick followup PR.

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -29,24 +31,23 @@ type ProcessedPackages k a =
 
 type ProcessedPackageVersions k1 k2 a = ProcessedPackages k1 (Map k2 a)
 
+type NameAndAddress = { address :: { owner :: String, repo :: String }, name :: RawPackageName }
```
thomashoneyman:
```suggestion
type NameAndAddress = { address :: Address, name :: RawPackageName }
```

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -29,24 +31,23 @@ type ProcessedPackages k a =
...
+type NameOriginalAndAddress = { address :: { owner :: String, repo :: String }, name :: PackageName, original :: RawPackageName }
```
thomashoneyman:
```suggestion
type NameOriginalAndAddress = { address :: Address, name :: PackageName, original :: RawPackageName }
```

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -29,24 +31,23 @@ type ProcessedPackages k a =
...
+type SemVerAndOriginal = { semVer :: SemVer, original :: RawVersion }
```
thomashoneyman: I sometimes find these highly-specific type synonym names a little harder to read than just seeing the record type directly, since they're just two or three fields; since they are only used in this file, what do you think of just inlining the types into the functions directly? For example:

```purs
forPackage
  :: ProcessedPackages RawPackageName PackageURL
  -> (RawPackageName -> PackageURL -> ExceptT ImportError Aff (Tuple { name :: RawPackageName, address :: Address } (Map RawVersion Unit))
  -> Aff (ProcessedPackages { name :: RawPackageName, address :: Address } (Map RawVersion Unit)
```

It makes the type signatures a bit long, but I still think they're reasonable to read and the functions are each only used in one place.

ci/src/Registry/Scripts/LegacyImport/Process.purs
thomashoneyman: I understand that widens the scope of this PR way out, so I understand if you'd prefer to punt on that. I'm OK with making the types monomorphic since it compiles and just addressing my other feedback.

ci/src/Registry/Scripts/LegacyImport/Process.purs
mikesol: Makes sense, totally doable

ci/src/Registry/Scripts/LegacyImport/Process.purs
mikesol: I'll do it now.

mikesol:
@thomashoneyman I've done an initial draft of inlining one function. It's not pretty, but it's all in one place. Lemme know if you'd like me to carve it up differently. Once I get feedback on that, I can keep going with the inlining. Another option is to go the other direction: carve out more stuff into `Process.purs` and keep `LegacyImport.purs` smaller.

thomashoneyman:
> @thomashoneyman I've done an initial draft of inlining one function. It's not pretty, but it's all in one place. Lemme know if you'd like me to carve it up differently. Once I get feedback on that, I can keep going with the inlining. Another option is to go the other direction: carve out more stuff into `Process.purs` and keep `LegacyImport.purs` smaller.

Yea, I think this is too much to try and put into this PR. Sorry for the dead end! Are you OK with doing this?

1. Revert https://github.com/purescript/registry/pull/239/commits/cce789a7d7dbefaf35b6033a22508ff472f82276
2. Remove the various `NameAndAddress` etc. type synonyms and just inline the records directly into the types for the `for*` functions (reusing the `Foreign.GitHub.Address` type where you need an `{ owner :: String, repo :: String }`)

That way I can go ahead and merge this PR and you can move on to the stats without getting bogged down in unnecessary refactoring at my request 😬

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -67,6 +69,11 @@ main = Aff.launchAff_ do
   for_ registry \semVer -> for_ semVer (insertManifest indexPath)
   log "Done!"
 
+type Address = { owner :: String, repo :: String }
```
thomashoneyman: This already exists in the `Foreign.GitHub` module, so you should be able to just reuse that.

mikesol:
> That way I can go ahead and merge this PR.

Done! No worries for the back-and-forth on refactorings, they're all minor and it's tough to know what works until we try it out.

thomashoneyman:
I'm personally OK with the long line lengths that were there before https://github.com/purescript/registry/pull/239/commits/710b4852b959192fee419f10cf0af1357a8e591a, and find it a bit easier to read than the line breaks. I know I'm deep into "this is my opinion" territory here, and perhaps you and @f-f feel differently.

thomashoneyman:
That said, I'm going to go ahead and 👍 and see if @f-f has anything to add. I think we'll probably want to tweak this more going forward, but it's not necessary to have the back-and-forth go on for too long.

thomashoneyman (dismissed):

mikesol:
@thomashoneyman @f-f I added the `stats` stuff to this PR just cuz I have time now & won't necessarily in a bit. The monomorphic stuff made it super easy to write as I just let the compiler guide me until I got the types right for the new fields in `Stats`. I think this corresponds to the spec you wanted, Thomas, but if not lemme know & I'll tweak it!

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -73,7 +80,51 @@ exampleFailures = PackageFailures $ Map.fromFoldable
   errsByVersion = Right
 
 exampleStats :: Stats.Stats
-exampleStats = Stats.errorStats examplePackageResults
+exampleStats = Stats.errorStats mockStats
+  where
+  mockLicense :: License
+  mockLicense = unsafeCoerce "None"
```
thomashoneyman: I'd rather not use `unsafeCoerce` here because it admits so many types -- for example, if we change `License` to not just be a simple string internally then this will silently break.

Instead, we should use the provided constructor functions for these various newtypes -- ie. `License.parse` or `PackageName.parse` or `SemVer.parseSemVer`. Then, depending on whether they return a `Maybe` or an `Either` we can unwrap them with either:

```purs
-- on `Maybe` values
unsafePartial $ Maybe.fromJust ...

-- on `Either` values
Either.fromRight' (\_-> unsafeCrashWith "Unexpected Left") ...
```

That way we get a compiler error if the constructor function for these newtypes changes to accept a different type.

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -73,7 +80,51 @@ exampleFailures = PackageFailures $ Map.fromFoldable
...
+
+  mockPackageName :: PackageName
+  mockPackageName = unsafeCoerce "foo"
+
+  mockSemVer :: SemVer
+  mockSemVer = unsafeCoerce "0.0.0"
```
thomashoneyman: This isn't correct -- a `SemVer` isn't just the raw string itself, it's a bigger structure:

https://github.com/purescript/registry/blob/b1f4b5e0052ab1f9c01c081e0d83a9f2ec6f92a7/ci/src/Foreign/SemVer.purs#L21-L30

You'll want to use `SemVer.parseSemVer` here:

https://github.com/purescript/registry/blob/b1f4b5e0052ab1f9c01c081e0d83a9f2ec6f92a7/ci/src/Foreign/SemVer.purs#L60-L61

and do the same thing described in [my prior comment](https://github.com/purescript/registry/pull/239/files?authenticity_token=kII5YqA1fsfSTJ1yXpArH55gg%2BZIzDd88oD4dZcJNiE%2F0c4e8aXNAR5cj%2FjeUVksJWVRQqnPrtyD6WR9PVmcRw%3D%3D&file-filters%5B%5D=.purs#r723988926)

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -73,7 +80,51 @@ exampleFailures = PackageFailures $ Map.fromFoldable
...
+                { address:
+                    { owner: "purescript"
+                    , repo: "a-repo"
+                    }
+                , name: mockPackageName
+                , original: raw
+                }
+            )
+        $ (identity :: Array ~> Array)
+        $ Map.toUnfoldable examplePackageResults.packages
```
thomashoneyman:
```suggestion
        $ (Map.toUnfoldable examplePackageResults.packages :: _ -> Array _)
```

ci/test/Registry/Scripts/BowerImport/Stats.purs
mikesol: One thing to check out is https://github.com/purescript/registry/issues/246. I think there may be an issue with the FFI here (I could be wrong). When I tried using the type suggested by the constructor, it crashed in the Js layer.

ci/test/Registry/Scripts/BowerImport/Stats.purs
mikesol: Totally makes sense. Would you be open to exporting a function ie `unsafePackageName` for the constructor in the file itself. The danger with this stuff is that people will wind up using it, which defeats the purpose of not exporting the constructor, but usually stuff that starts with `unsafe` is known to be use-at-your-own-risk.

ci/test/Registry/Scripts/BowerImport/Stats.purs
thomashoneyman: If you weren't using `parseSemVer` then I think you would fail the `instanceof` check in the JS layer

ci/test/Registry/Scripts/BowerImport/Stats.purs
thomashoneyman: I'd rather just use the `unsafe` in the places where we really need to use it, and not export an unsafe constructor; we should only really be using unsafe things in the test code and I don't mind the extra verbosity there.

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -122,8 +122,7 @@ exampleStats = Stats.errorStats mockStats
                 , original: raw
                 }
             )
-        $ (identity :: Array ~> Array)
-        $ Map.toUnfoldable examplePackageResults.packages
+        $ (Map.toUnfoldable examplePackageResults.packages :: _ -> Array _)
```
thomashoneyman: Whoops, must have missed a paren. Maybe this?

```purs
$ ((Map.toUnfoldable examplePackageResults.packages) :: _ -> Array _))
```

mikesol:
@thomashoneyman I incorporated your suggestions - everything should be in there now.

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -116,10 +167,18 @@ errorStats = do
...
+              <> show exampleStats.countOfPackageSuccessesWithoutFailures
+              <> " with successes, "
+              <> show exampleStats.countOfPackageFailuresWithoutSuccesses
+              <> " with failures)"
+          , "Versions: " <> show exampleStats.totalVersions
+              <> " total ("
+              <> show exampleStats.countOfVersionSuccessesWithoutFailures
+              <> " successful, "
+              <> show exampleStats.countOfVersionFailuresWithoutSuccesses
+              <> " failed)"
```
thomashoneyman: This still results in output I find a little difficult to parse (pasted at the bottom of this comment). Specifically, it's not clear what packages fully succeeded (no versions dropped), partially succeeded (some versions dropped), or fully failed (all versions dropped). Could we try adjusting to:

```
Packages: 1575 total (1175 succeeded, 200 partially succeeded, 200 failed)
Versions: 10,000 total (8,000 succeeded, 2,000 failed)
```

as the top line? To do this, we'd have to go through package names and look them up.

* If a package is present in the `RegistryIndex` and not present in the `PackageFailures` then it succeeded
* If a package is present in the `RegistryIndex` _and_ the `RegistryIndex` then it partially succeeded
* If a package is not present in the `RegistryIndex` and is present in the `PackageFailures` then it failed

thomashoneyman: Current output on this PR from running `spago run -m Registry.Scripts.LegacyImport':

```
Packages: 1575 total (1200 with successes, 127 with failures)
Versions: 774 total (486 successful, 76 failed)
Failures by error:
  manifestError: 1215 occurrences (314 packages / 1215 versions)
    badLicense: 481 occurrences (76 packages / 481 versions)
    missingLicense: 452 occurrences (141 packages / 452 versions)
    badDependencyVersions: 307 occurrences (114 packages / 307 versions)
    badVersion: 50 occurrences (34 packages / 50 versions)
  noDependencyFiles: 66 occurrences (34 packages / 66 versions)
  resourceError: 32 occurrences (32 packages / 0 versions)

Comment (thomashoneyman) [reply ci/test/Registry/Scripts/BowerImport/Stats.purs]: While you're here, could we replace the

```
X occurrences (X packages / X versions)
```

lines with the below instead, since occurrences == versions?

```
# perhaps this?
X versions across X packages

# or this?
X versions (across X packages)
```

(or another format of your choice -- we just don't need to duplicate occurrences / versions since they're the same)

**Comment by thomashoneyman** [comment]
@mikesol Just a couple quick comments about the output format and then we can get this in. Thanks!

**Review by f-f — APPROVED** [review]
This looks good, thanks @mikesol! I think this is good to go once we address Thomas' suggestions

**Comment by mikesol** [comment]
@thomashoneyman Done :+1:

**Comment by thomashoneyman** [comment]
@mikesol — in this output, what is the difference between “partially failed” and “partially succeeded”? Shouldn’t they be the same thing?

I think we still need to bucket the successes / failures according to the three criteria set out in https://github.com/purescript/registry/pull/239#discussion_r724891862

For example, this test output:

```
Packages: 7 total (2 totally succeeded, 2 partially succeeded, 5 totally failed, 5 partially failed)
Versions: 8 total (2 totally succeeded, 5 partially succeeded, 4 totally failed, 7 partially failed)
```

...doesn't really make sense:

* If there are 7 packages, then succeeded + partially succeeded + failed should add up to 7, but this adds up to 14.
* Packages can "partially succeed" if some versions in the package parsed, but other versions didn't. But while there are four categories represented in this output, there are really only three possible categories: "fully succeeded" (no versions failed), "partially succeeded" (some, but not all, versions failed), or "failed" (all versions failed).
* Versions can't "partially succeed"; a version either parses or doesn't. So versions should only have "succeeded" and "failed" categories.

File: ci/test/Registry/Scripts/BowerImport/Stats.purs [hunk 13]
```diff
@@ -3,8 +3,15 @@ module Test.Registry.Scripts.LegacyImport.Stats where
 import Registry.Prelude
```
thomashoneyman, line 3: I'm not totally sure what's going on here, but it looks like the directory for this is incorrect (it should be `ci/test/Registry/Scripts/LegacyImport/Stats.purs`). Can we rename it?

thomashoneyman (changes requested):
I think a few things need to be fixed about this, as per my latest comments

mikesol:
> I think a few things need to be fixed about this, as per my latest comments

Done & ready for another review.

thomashoneyman:
The output looks better:

```
Packages: 1575 total (1200 totally succeeded, 248 partially succeeded, 127 totally failed)
Versions: 11521 total (10239 totally succeeded, 0 partially succeeded, 1282 totally failed)
Failures by error:
  manifestError: 1215 versions across 314 packages
    badLicense: 481 versions across 76 packages
    missingLicense: 452 versions across 141 packages
    badDependencyVersions: 307 versions across 114 packages
    badVersion: 50 versions across 34 packages
  noDependencyFiles: 66 versions across 34 packages
```

thomashoneyman (approved):

PR #240, opened 2021-09-27
Collect manifest information from multiple files, not just Bowerfile

### LINKED ISSUE #218: Fall back to spago.dhall files for packages without a bower.json

### LINKED ISSUE #220: Parse LICENSE file if no license is listed in the Bowerfile

### LINKED ISSUE #233: Attempt to parse licenses and dependencies even without a bower.json
This issue is related to #220, #218, and #211, all of which discuss how to import packages that have a malformed Bowerfile, or a Bowerfile missing a `license` key, or have no Bowerfile at all (but have other files, such as a `spago.dhall` file).

I propose that we view this problem from a new angle: how do we get the information necessary to produce a `Manifest` for a package, regardless of the presence of a Bower or other files?

https://github.com/purescript/registry/blob/c28f98584e9858d3185a3842dd2fca31514574f4/v1/Manifest.dhall#L13-L24

To construct this, we need:

1. The package name (we get this from the `new-packages.json` and `bower-packages.json` files)
2. The package repository (we get this from the same two files)
3. The package version (we get this from GitHub tags)
4. The package license (we get this from the Bowerfile)
5. The package targets, ie. dependencies and dev dependencies (we get this from the Bowerfile)

We already handle the first three items about as well as we possibly can, but the work in #220, #218, and #211 all have to do with getting the package license and targets. We've started with the idea that we'll just pull everything out of the Bowerfile, but a lot of packages have missing Bowerfiles or Bowerfiles with no license key.

My proposal is that we no longer consider ourselves to be parsing a `BowerFile`, but rather a license and targets, and it's not necessary that we get that information from a Bowerfile. For example:

1. We can get the license and/or targets directly from a Bowerfile.
2. We can get the license and/or targets directly from a Spago file (either by generating a `bower.json` file or using say `spago ls deps`).
3. We can possibly get targets from `psc-package` files, though I've never used it myself and I'm not sure if it's worth bothering with this.
4. We can get the license from a package.json file (but not dependencies).
5. We can get the license from the repository root in the LICENSE file.

So I'm proposing that we attempt to get the license and targets by moving through these steps: read as much information as we can from the Bowerfile, then the Spago file, then the package.json, and finally the LICENSE in the repo root. Only if we can't get the license and targets from _any_ of those sources, or a combination of them, do we fail the package.

thomashoneyman:
Closes #233, #220, #218. Supersedes #234.

This PR implements a `LegacyImport` script instead of a `BowerImport` script, because we now look at multiple files to try and import a manifest instead of just the Bowerfile. Those files include the `LICENSE`, `package.json`, `bower.json`, and `spago.dhall` and `packages.dhall` files.

Implementing the pipeline using multiple files to produce the manifest required shuffling around quite a bit, but the majority of the code moved into the `constructManifestFields` function which produces the license and dependencies for a package from a combination of the available files.

There are two other notable changes:

1. The `withCache` function works a little differently. First, it no longer requires `EncodeJson` and `DecodeJson` instances, because you can be storing something like a `LICENSE` file; it now uses `String`. Second, the `withCache` function now takes an `ImportError -> Boolean` function that allows you to cache files _even if_ producing the file throws an exception. The exception is stored instead of the file contents, and when the file is read the exception is thrown. With this in place we can remove all the `filterFailedPackages` stuff, which is a bit error-prone, and we no longer need to read in the excluded packages.
2. The `Error` type has been changed around a little to accommodate the new files we're fetching. There's now a `ResourceError` that represents that files related to the package couldn't be fetched. `MissingBowerfile` has been replaced since it's not really an error any more, and `NoDependencyFiles` took its place to represent that both a spago _and_ bower file are missing.


ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -177,47 +158,23 @@ downloadBowerRegistry = do
...
-            throwError $ BadStatus $ un StatusCode status
-        | otherwise -> case BowerFile.parse body of
-            Left err -> do
-              let printedErr = BowerFile.printBowerFileParseError err
-              log $ "Unable to parse bowerfile: " <> printedErr <> " Malformed body: " <> body <> "."
-              throwError $ MalformedBowerJson { error: printedErr, contents: body }
-            Right bowerfile -> pure bowerfile
-
+-- | TODO: This check can't be done on a per-package basis without ordering.
+-- | It requires that everything is first topographically-sorted.
```
f-f:
```suggestion
-- | It requires that all versions are topologically-sorted first.
```

thomashoneyman:
These are the results, working from a `.cache` with only releases cached:

```
Number of successful packages: 1448
Number of failed packages: 375
Number of successful versions: 10234
Number of failed versions: 1281
Failures by error:
  manifestError: 1214 occurrences (314 packages / 1214 versions)
    badLicense: 480 occurrences (76 packages / 480 versions)
    missingLicense: 452 occurrences (141 packages / 452 versions)
    badDependencyVersions: 307 occurrences (114 packages / 307 versions)
    badVersion: 50 occurrences (34 packages / 50 versions)
  noDependencyFiles: 66 occurrences (34 packages / 66 versions)
  resourceError: 32 occurrences (32 packages / 0 versions)
```

It's a significant improvement! Compare that to the current main branch:

```
Number of successful packages: 1369
Number of failed packages: 474
Number of successful versions: 9565
Number of failed versions: 1694
Failures by error:
  manifestError: 1378 occurrences (371 packages / 1378 versions)
    missingLicense: 928 occurrences (241 packages / 928 versions)
    badDependencyVersions: 283 occurrences (104 packages / 283 versions)
    badLicense: 179 occurrences (45 packages / 179 versions)
    badVersion: 48 occurrences (32 packages / 48 versions)
  nonRegistryDependencies: 246 occurrences (47 packages / 246 versions)
  missingBowerfile: 64 occurrences (34 packages / 64 versions)
  noReleases: 32 occurrences (32 packages / 0 versions)
  malformedBowerJson: 5 occurrences (4 packages / 5 versions)
```

The number of missing licenses has been cut in half, and about 700 more versions are included in the final registry. I would like to keep looking at the exclusions file to see if any packages are suspiciously failing, but on the whole this is a nice improvement over the current stats.

thomashoneyman:
The tests are failing because I added a Licensee test; we’re going to have to add that to CI as a CLI tool.

ci/test/Registry/Scripts/BowerImport/Stats.purs
```diff
@@ -26,20 +26,20 @@ examplePackageResults =
 
 exampleFailures :: PackageFailures
 exampleFailures = PackageFailures $ Map.fromFoldable
-  [ printImportErrorKey MissingBowerfile /\ missingBowerfileErrors
-  , printImportErrorKey NoReleases /\ noReleasesErrors
+  [ printImportErrorKey NoDependencyFiles /\ noDependencyFileErrors
```
thomashoneyman, line 29: These test files are just changed because some of the errors were renamed or removed.

ci/test/Foreign/Licensee.purs
```diff
@@ -0,0 +1,47 @@
...
+
+import Registry.Prelude
+
+import Foreign.Licensee as Licensee
+import Node.Path as Node.Path
+import Test.Fixtures.ManifestFiles as ManifestFiles
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+licensee :: Spec.Spec Unit
```
thomashoneyman, line 11: We'll need to add `licensee` to the tooling available in CI so that tests pass. These tests pass locally and help show in what situations `licensee` works and what its output looks like, as well as what it _doesn't_ work on.

ci/src/Registry/Scripts/BowerImport/SpagoJson.purs
```diff
@@ -0,0 +1,62 @@
...
+packageDependencies :: SpagoJson -> Object String
+packageDependencies (SpagoJson { dependencies, packages }) = do
+  let
+    foldFn m name = fromMaybe m do
+      version <- Map.lookup name packages
+      pure $ Object.insert (un RawPackageName name) (un RawVersion version) m
+
+  Array.foldl foldFn Object.empty dependencies
+
+newtype SpagoJson = SpagoJson
```
thomashoneyman, line 34: This is the output of calling `dhall-to-json` on a `spago.dhall` file.

ci/src/Registry/Scripts/BowerImport/Process.purs
```diff
@@ -208,11 +202,49 @@ withCache path maybeDuration action = do
...
+--
+-- Inspired by the work originally done in:
+-- https://github.com/natefaubion/purescript-dodo-printer/blob/540dba0442abe686c0b211868d6f423e5df81b69/test/Snapshot.purs#L54-L61
+parBounded
+  :: forall k t a b
+   . TraversableWithIndex k t
+  => (t a)
+  -> (k -> a -> Aff b)
+  -> Aff Unit
+parBounded t f = do
```
thomashoneyman, line 230: I was finding that the existing processing code runs super slowly on the new code, because we're fetching several more files and running `licensee` and reading from `stdout` for each package version; doing this sequentially makes each package version take half a second or so.

This function makes it easy to parallelize a traversal over something like a map. I found that traversing all package versions in parallel will fail because of resource contention over things like the network and stdout, we're OK if we uses a bounded queue and only run a few processing steps in parallel at any given time.

This speeds things up dramatically, and it replaces the use of `forWithIndex_` when you're able to do something in parallel.

ci/src/Registry/Scripts/BowerImport/Process.purs
```diff
@@ -124,74 +137,55 @@ insertFailure key value failures = do
...
-  => FilePath
+  :: forall e a
+   . Serialize e a
+  -> (ImportError -> Boolean)
+  -> FilePath
   -> Maybe Hours
   -> ExceptT ImportError Aff a
   -> ExceptT ImportError Aff a
-withCache path maybeDuration action = do
+withCache { encode, decode } cacheFailure path maybeDuration action = do
```
thomashoneyman: The major changes here are that:

1. `withCache` no longer uses `EncodeJson` and `DecodeJson` and rather accepts a serializer that can work with `String` as the raw source. That's necessary because we have `.dhall` files and `LICENSE` files and so on to cache now, and they aren't JSON.
2. `withCache` now accepts a `cacheFailure` argument that allows you to still cache something if the provided action throws an exception. For example, we catch and then cache `404` errors in requests, so that we don't make those requests again in the future. This works because all exceptions we throw are serializable. If `withCache` reads back an exception from disk then it will just re-throw it -- as far as the calling code is concerned, it's identical to the original request failing and throwing an exception.

The second point is why we can remove the `filterFailedPackage*` functions altogether -- the filtering is essentially happening at the cache level now.

ci/src/Registry/Scripts/BowerImport/Error.purs
```diff
@@ -86,16 +87,29 @@ manifestErrorKey = ImportErrorKey "manifestError"
...
-  MalformedBowerJson _ -> ImportErrorKey "malformedBowerJson"
+  NoDependencyFiles -> ImportErrorKey "noDependencyFiles"
   NonRegistryDependencies _ -> ImportErrorKey "nonRegistryDependencies"
   NoManifests -> ImportErrorKey "noManifests"
   ManifestError _ -> manifestErrorKey
 
--- | An error representing why a Bowerfile cannot be migrated into a manifest.
+-- | An error fetching a resource necessary to produce a Manifest for a
+-- | given package.
+type ResourceError = { resource :: RemoteResource, error :: RequestError }
```
thomashoneyman, line 99: These more granular errors are important for being able to selectively cache failures in the `withCache` function. We can, for example, cache request errors that result in a 404, for specific resources.

ci/src/Foreign/Jsonic.purs
```diff
@@ -1,11 +1,19 @@
...
+import Data.Argonaut as Json
+import Data.Bifunctor (lmap)
+import Data.Either (Either(..))
 import Data.Function.Uncurried (Fn3, runFn3)
 
-parse :: String -> Either String Json
-parse = runFn3 parseJsonicImpl Left Right
+-- | A drop-in replacement for `Data.Argonaut.parseJson` that uses a lenient
+-- | JSON parser.
+parseJson :: String -> Either Json.JsonDecodeError Json
```
thomashoneyman, line 11: I ended up wanting to use the `Jsonic` parser more places than just parsing the Bowerfile, and so I modified this module to produce a drop-in replacement for `Argonaut.parseJson`. You can see it used in a few places and I think we should use it in general as the replacement.

ci/src/Foreign/Licensee.purs
```diff
@@ -0,0 +1,52 @@
...
+    -- Licensee will exit with `Normally 1` if it didn't parse any licenses,
+    -- but we consider this valid Licensee output.
+    NodeProcess.Normally n | n == 0 || n == 1 -> do
+      let
+        parse :: String -> Either Json.JsonDecodeError (Array String)
+        parse str = Json.parseJson (String.trim str) >>= \json -> do
+          obj <- Json.decodeJson json
+          licenses <- obj `Json.getField` "licenses"
+          spdxIds <- traverse (_ `Json.getField` "spdx_id") licenses
+          pure spdxIds
```
thomashoneyman, line 39: `licensee` returns a _lot_ of information, including the full body of the license, but we only really want to get out the `spdx_id`s, which we can parse into valid SPDX licenses.

ci/src/Registry/API.purs
```diff
@@ -318,13 +322,3 @@ pushToMaster packageName path = Except.runExceptT do
         info result.stderr
         pure $ Right unit
       _ -> pure $ Left result.stderr
-
-readBowerfile :: String -> Aff (Either String BowerFile)
```
thomashoneyman, line 322: I just realized this is basically the same as `try <<< readJsonFile`, so I replaced this with that on line 171.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -58,81 +67,47 @@ main = Aff.launchAff_ do
...
   newPackages <- readRegistryFile "new-packages.json"
 
-  let
-    handleError _ = do
-      log "bower-exclusions.json does not exist, writing..."
-      let failures = PackageFailures Map.empty
-      writeJsonFile "bower-exclusions.json" failures
-      pure (Right failures)
-
-  excludedPackages <- Except.catchError (readJsonFile "bower-exclusions.json") handleError >>= case _ of
```
thomashoneyman, line 73: It's no longer necessary to use the excluded packages or do any filtering. The filtering only existed to avoid making requests for packages that we know will fail, and that is now handled automatically in the cache layer.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -151,14 +126,21 @@ downloadBowerRegistry = do
 
     pure $ Tuple outerKey innerKey
 
-  log "Converting bowerfiles to manifests..."
+  log "Converting to manifests..."
   let forPackageRegistry = Process.forPackageVersion packageRegistry _.original _.original
-  manifestRegistry <- forPackageRegistry \{ name, address } { semVer } bowerfile -> do
+  manifestRegistry <- forPackageRegistry \{ name, original: originalName, address } tag _ -> do
+    manifestFields <- constructManifestFields originalName tag.original address
```
thomashoneyman, line 126: The `constructManifestFields` function is the replacement for `fetchBowerfile`, but it fetches _all_ relevant files for a package and then parses things from each of them to produce a partial `Manifest`. That can get fed directly back into `toManifest`.

ci/src/Registry/Scripts/BowerImport.purs
```diff
@@ -360,3 +318,167 @@ readRegistryFile source = do
...
+      packagesDhall <- parallel $ fileRequest PackagesDhall Process.stringSerializer
+      in { spagoDhall, packagesDhall }
+
+    tmp <- liftEffect Tmp.mkTmpDir
+    liftAff $ FS.writeTextFile UTF8 (tmp <> "/packages.dhall") files.packagesDhall
+
+    spagoJson <- do
+      let
+        mkError = ResourceError <<< { resource: FileResource SpagoDhall, error: _ } <<< DecodeError
+        runDhallJson = Dhall.dhallToJson { dhall: files.spagoDhall, cwd: Just tmp }
```
thomashoneyman, line 428: This follows your advice, @f-f, from #234 where you suggested using `dhall-to-json` instead of converting the Spago file into a Bowerfile. You're right: it's much simpler (and faster!).

thomashoneyman:
Whew, sorry, had to do a little CI fixup (case sensitivity is a bear!), but this is now _actually_ good to go.

ci/spago.dhall
```diff
@@ -1,4 +1,5 @@
 { name = "registry-ci"
+, license = "MIT"
```
f-f: Interestingly enough, there's no license at all in the repo, so I think so far it has been unlicensed? Not sure about the implications of that, but if we're picking a license for this code then I'd like to use BSD-3, which is a little less permissive than MIT

ci/src/Foreign/Jsonic.purs
f-f: Agreed, this is great!

ci/src/Foreign/Licensee.purs
```diff
@@ -0,0 +1,52 @@
...
+import Node.ChildProcess as NodeProcess
+import Node.FS.Aff as FS
+import Sunde as Process
+
+-- | Attempt to detect the license associated with a set of provided files
+detectFiles :: Array { name :: FilePath, contents :: String } -> Aff (Either String (Array String))
+detectFiles files = do
+  tmp <- liftEffect Tmp.mkTmpDir
+  files # Parallel.parTraverse_ \{ name, contents } ->
+    FS.writeTextFile UTF8 (tmp <> "/" <> name) contents
```
f-f, line 18: Could we add a note here about why we are copying the files over to a temp folder instead of checking them in place?

ci/src/Registry/API.purs
```diff
@@ -169,22 +168,27 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
...
   when fromBower do
-    liftAff (readBowerfile (absoluteFolderPath <> "/bower.json")) >>= case _ of
+    liftAff (try (readJsonFile (absoluteFolderPath <> "/bower.json"))) >>= case _ of
       Left err ->
-        throwWithComment $ "Error while reading Bowerfile: " <> err
-      Right bowerfile -> do
+        throwWithComment $ "Error while reading Bowerfile: " <> Aff.message err
+      Right (Left err) ->
+        throwWithComment $ "Could not decode Bowerfile: " <> Json.printJsonDecodeError err
+      Right (Right bowerfile) -> do
```
f-f, line 176: I'm not sure I understand why we are having a nested Either here?

ci/src/Registry/Scripts/BowerImport/Bowerfile.purs
```diff
@@ -0,0 +1,44 @@
+module Registry.Scripts.BowerImport.Bowerfile
```
f-f: The slight rename of this file makes it so that git doesn't show the diff. I appreciate these renames, but I'd like to see the diff of the changes in this PR, and batch all the renames in a subsequent PR

ci/test/Main.purs
```diff
@@ -142,7 +145,7 @@ decodeEventsToOps = do
         , fromBower: false
         }
 
-    res <- API.readOperation "test/fixtures/issue_comment.json"
+    res <- API.readOperation "test/Fixtures/issue_comment.json"
```
f-f: Is there a reason for changing this folder name?

ci/test/Foreign/Licensee.purs
f-f: I think this is solved now that we are using the right nix file?

ci/src/Registry/Scripts/BowerImport/SpagoJson.purs
f-f: Let's add the comment above as a docstring for this type?

ci/src/Registry/Scripts/BowerImport/Process.purs
```diff
@@ -35,20 +39,25 @@ forPackage
...
+    Except.runExceptT (f key value) >>= case _ of
       Left err -> do
         let
           errorType = BowerImport.Error.printImportErrorKey err
           name = keyToPackageName key
           failure = Map.singleton name (Left err)
-        State.modify \state -> state { failures = insertFailure errorType failure state.failures }
+          modify state = state { failures = insertFailure errorType failure state.failures }
+        state <- AVar.take var
+        AVar.put (modify state) var
```
f-f: Is this atomic?

ci/src/Registry/Scripts/BowerImport/Process.purs
```diff
@@ -95,25 +106,27 @@ forPackageVersionKeys
...
-              failure = Map.singleton name $ Right $ Map.singleton tag err
-            State.modify \state -> state { failures = insertFailure errorType failure state.failures }
-          Right (Tuple k3 k4) -> do
-            let
-              newPackage = Map.singleton k4 value
-              insertPackage = Map.insertWith Map.union k3 newPackage
-            State.modify \state -> state { packages = insertPackage state.packages }
+  var <- AVar.new { failures: input.failures, packages: Map.empty }
+  parBounded input.packages \k1 inner ->
+    parBounded inner \k2 value -> do
```
f-f, line 102: Would nested `parBounded` multiply the amount of threads running? Is this fine?

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: This is cool - a couple of questions:
- I might have missed where `Serialize` is defined, do you have a pointer?
- Not sure what's the point of that `ImportError -> Boolean` function. Why do we have to inject it? Is there a case where we don't want to use this functionality?

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: This is great! :clap:

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: Oh, I guess it is since `take` should be blocking?

ci/src/Registry/Scripts/BowerImport.purs
f-f: Wonderful!

ci/src/Registry/Scripts/BowerImport.purs
f-f: Great news! :rocket:

f-f (commented):
Great work @thomashoneyman! I left a bunch of small comments, and this is almost ready to go

ci/spago.dhall
thomashoneyman: Totally fine with me. And we don't even have to license it. I actually added this while testing out `licensee` 😬 Whatever you'd like!

ci/src/Foreign/Licensee.purs
thomashoneyman: I can add one, yes. `licensee` doesn't have an option for reading from STDIN for these files, as far as I'm aware; `detectFiles` is just `detect`, where you write the files first instead of already having them ready in a directory.

ci/test/Main.purs
thomashoneyman: It was already changed in #229, but due to case-sensitivity on Macs we ended up with both a `fixtures` and a `Fixtures` directory and this makes it so we just have a `Fixtures` directory.

ci/src/Registry/API.purs
thomashoneyman: The first `Either` comes from reading the file, which can throw an exception if the file doesn't exist. The second comes from decoding the file contents, which can fail with a JSON decoding error. Previously we had a specialized `readBowerfile` function that checked for the file's existence, but now that we're just using `try $ readJsonFile` we end up with the nested Either. We can't flatten it because the error types are different (`Error` vs. `JsonDecodingError`).

ci/src/Registry/Scripts/BowerImport/Process.purs
thomashoneyman: Yes, nesting does increase the number of threads running, but in my tests this seems to be doing just fine. If we run into any issues then we could replace one of these with a `forWithIndex_`.

ci/src/Registry/Scripts/BowerImport/Process.purs
thomashoneyman: Yea, once the `AVar` is empty via a `take` then everything blocks it until we `put` again.

ci/src/Registry/Scripts/BowerImport/Process.purs
thomashoneyman: > I might have missed where Serialize is defined, do you have a pointer?

`Serialize` is defined just a few lines above this:
https://github.com/purescript/registry/pull/240/files#diff-45cac7fbb3e235f610383520c4e8fafb270a3c7e88777b1e937cbd59262458beR140-R143

> Not sure what's the point of that `ImportError -> Boolean` function. Why do we have to inject it? Is there a case where we don't want to use this functionality?

We could inline it from its use elsewhere, namely that we cache resource errors pertaining to the GitHub API and 404 errors fetching files, and that _should_ be fine. I originally had this a little more generic, letting you supply whatever exception type you want along with a `cacheFailure` function for that exception type, but since it's back to being a concrete `ImportError` then inlining it seems fine if you'd like to do that.

ci/src/Registry/Scripts/BowerImport/SpagoJson.purs
thomashoneyman:
```suggestion
-- | The output of calling `dhall-to-json` on a `spago.dhall` file
newtype SpagoJson = SpagoJson
```

ci/test/Foreign/Licensee.purs
thomashoneyman: Yep! This comment is no longer relevant.

thomashoneyman:
Some of my responses to your comments are only showing up in the review, and not in this main PR view, so you might have to read them in the review tab. Thanks for your comments! I believe I've addressed everything.

ci/src/Registry/API.purs
f-f: I see, thanks!

ci/spago.dhall
f-f:
```suggestion
, license = "BSD-3-Clause"
```

ci/src/Foreign/Licensee.purs
f-f: I see - so IIUC `licensee` works on a directory (and not on single files) and sometimes we don't have that ready and we prepare one with the files that we have?

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: Great! Then since we repeat this pattern many times I was wondering if it'd be worth making a helper that `take`s, does stuff, and then `put`s, so that we reduce clutter. No strong opinion on this though, feel free to ignore

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: Wonderful, thanks for checking!

ci/src/Registry/Scripts/BowerImport/Process.purs
f-f: Thanks for the explanations! Since we have concrete types inlining seems like a good idea, but no strong opinion on this so you can skip this suggestion if you prefer the current shape

ci/test/Main.purs
f-f: Oh, I didn't realize then - thanks for the pointer. I have a slight preference for the lowercase version (I am definitely traumatized by macOS' filesystem shenanigans with that) so I'd appreciate if you can switch to that, but no worries if you wouldn't like to

f-f (approved):
Thanks for addressing the feedback! I left a few minor comments, and I'll approve this since it's basically good to go :slightly_smiling_face:

ci/src/Foreign/Licensee.purs
thomashoneyman: That’s right.

ci/test/Main.purs
thomashoneyman: I went with upper-case because we have PureScript files in the folder, so they’re say Test.Fixtures.Manifest — but we could also lower case and possibly move those files elsewhere.

PR #242, opened 2021-10-04
Add tests for Registry.Index module

### LINKED ISSUE #212: Write tests for the Registry.Index module
The [`Registry.Index` module](https://github.com/purescript/registry/blob/master/ci/src/Registry/Index.purs) supports reading and writing to the registry index. (For context, the registry index is a listing of all packages in the registry along with their manifest files.) This code has not yet been tested, and we need to make sure that once run it does produce a valid registry index which can be read.

colinwahl:
Fixes #212

Adds tests for the `Registry.Index` module. In particular, we keep a version of the registry index in memory as well as on the file system, and we insert new packages checking our in memory version matches the version stored in files.

In particular, it covers the following cases:

* Two/three/four character `PackageName`, since these cover the different directory structures
* Version collision, where the last inserted version is preferred
* Checking the index directory matches the directory structure we expect 
https://github.com/purescript/registry/blob/b1f4b5e0052ab1f9c01c081e0d83a9f2ec6f92a7/ci/src/Registry/Index.purs#L68-L77

Right now this test is in `Aff`, where it fails if we throw an exception. I'd like to move this into `Spec` somehow, but each test requires the state from the last step, and I wasn't sure how to do that. I'm happy to clean it up if it is wanted, though.


thomashoneyman:
@f-f We're going to do a little more code cleanup on this on Thursday, but it's pretty much there -- this is the test output:

```
Registry Index » Registry index writes to and reads back from disk
  ✓︎ Reads an empty index from disk
  ✓︎ Inserts ab version 1.0.0
  ✓︎ Inserts ab version 1.0.0
  ✓︎ Inserts ab version 2.0.0
  ✓︎ Inserts abc version 1.0.0
  ✓︎ Inserts abc version 2.0.0
  ✓︎ Inserts abcd version 1.0.0
  ✓︎ Inserts abcd version 2.0.0
  ✓︎ Matches the expected directory structure
```

and with @colinwahl's fixes to the main `Registry.Index` file we can now confirm that it _does_ work as expected. If you'd like to review, please go for it; otherwise, I've verified the work and once we clean it up a tiny bit I'll merge on Thursday.

ci/test/Registry/Index.purs
```diff
@@ -0,0 +1,94 @@
...
+import Registry.Index as Index
+import Registry.PackageName as PackageName
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+import Test.Support.Manifest as Support.Manifest
+
+type TestIndexEnv =
+  { tmp :: FilePath
+  , readMemory :: Aff RegistryIndex
+  , writeMemory :: RegistryIndex -> Aff Unit
```
thomashoneyman: There's not much reason to type these as `Aff` here, because we lift them into `ReaderT` later on anyway. And since we process the environment when we write handler for `Spec.before`, we might as well pull these functions out at that point. So I think we should just write this as:

```suggestion
  , indexRef :: Ref RegistryIndex
```

ci/test/Registry/Index.purs
```diff
@@ -0,0 +1,94 @@
...
+mkTestIndexEnv :: Aff TestIndexEnv
+mkTestIndexEnv = liftEffect do
+  tmp <- Tmp.mkTmpDir
+  ref <- Ref.new (Map.empty :: RegistryIndex)
+
+  let
+    writeMemory index = liftEffect $ Ref.write index ref
+    readMemory = liftEffect $ Ref.read ref
+
+  pure { tmp, writeMemory, readMemory }
```
thomashoneyman: If we follow the above comment (https://github.com/purescript/registry/pull/242/files#r721715445) then this would just become:

```suggestion
  registryRef <- Ref.new (Map.empty :: RegistryIndex)
  pure { tmp, registryRef }
```

ci/test/Registry/Index.purs
```diff
@@ -0,0 +1,94 @@
...
+          , [ "3", "a", "abc" ]
+          , [ "ab", "cd", "abcd" ]
+          ]
+
+      actualPaths `Assert.shouldEqual` expectedPaths
+  where
+  runBefore = do
+    { tmp, readMemory, writeMemory } <- Reader.ask
+    index <- lift $ readMemory
+    pure { tmp, index, writeMemory: lift <<< writeMemory }
```
thomashoneyman: If we wait to create the read / write functions until here, then this code block can become:

```suggestion
  runBefore = do
    { tmp, registryRef } <- Reader.ask
    index <- lift $ Ref.read registryRef
    let writeMemory new = lift $ Ref.write new registryRef
    pure { tmp, index, writeMemory }
```

thomashoneyman (approved):

Issue #214, opened 2021-09-08
Produce a graph of package versions and their dependencies from the registry index


thomashoneyman:
The registry contains a registry index, which is a list of all packages in the registry along with their manifests (which themselves include information about dependencies, licenses, and so on):

https://github.com/purescript/registry/blob/3f5c48e518f657ab03c4178df976648c9fb1437f/ci/src/Registry/Index.purs#L24

In order to upload a new registry index as part of our pipeline, we need to be able to order packages at specific versions by their dependencies, so that we only upload packages that already have all their dependencies present (at correct versions) in the registry. That way the registry index is always correct.

I'm proposing we implement a function:

```purs
toVersionsGraph :: RegistryIndex -> PackageVersionsGraph
```

where a `PackageVersionsGraph` contains every package version in the registry, and we can order it topologically:

```purs
toOrderedArray :: PackageVersionsGraph -> Array Manifest
```

...so as to walk through each package version in order and upload it to the registry index.

thomashoneyman:
I've assigned this to you, @f-f, since you said you had some code locally for it. If I'm misremembering then please let me know and I will unassign it!

f-f:
This is all right thanks, I will sort out the code that I have locally and push up a PR!

colinwahl:
I've made good progress on this on my branch `package-version-graph` - I am hoping to have a PR up soon, just finishing up some last cleanup.

PR #243, opened 2021-10-05
Add graphlib to do topological sorting

f-f:
This is a small step towards #214


thomashoneyman:
@f-f Any reason not to use https://github.com/purescript/purescript-graphs, which has a toposort? The graph representation there seems like it would map well to our use case (cc: @colinwahl). The ordering isn't specified for graphs with cycles, but packages shouldn't have cycles anyway and that doesn't necessarily mean it won't work, just that it doesn't have a specified order. I'm not necessarily against using `graphlib`, but if there's a PureScript solution it's always nice to use it :)

f-f:
@thomashoneyman sure, somehow I didn't find this when searching for a PS solution!
We can use that library and close this PR

thomashoneyman:
No worries 😆 somehow literally no one knows about that library, and I probably wouldn't either if I hadn't had to do 0.14 updates.

Issue #246, opened 2021-10-07
Implementation of SemVer does not line up with expectation in semver.js


mikesol:
I believe that semver.js is expecting just a version (a string) when it does stuff like comparing, but in the FFI we are passing it a record. I think the fix here would be to pass semver.js the `version` field of the record in `SemVer`. As I'm not exactly sure if this is right, I'll file it as an issue, and if folks confirm this is indeed the case, I can make a PR.

thomashoneyman:
It's a bit confusing -- the `compare` function from `node-semver` is implemented like this:

https://github.com/npm/node-semver/blob/f1e4e293de1ad73fa250336bfcfb9b4142d21d1b/functions/compare.js#L2-L3
```js
const compare = (a, b, loose) =>
  new SemVer(a, loose).compare(new SemVer(b, loose))
```

which implies that `a` and `b` are versions, since they get constructed into `SemVer` classes. But the `SemVer` constructor itself checks whether it's been passed an existing `SemVer`:

https://github.com/npm/node-semver/blob/f1e4e293de1ad73fa250336bfcfb9b4142d21d1b/classes/semver.js#L7-L20
```js
class SemVer {
  constructor (version, options) {
    options = parseOptions(options)


    if (version instanceof SemVer) {
      if (version.loose === !!options.loose &&
          version.includePrerelease === !!options.includePrerelease) {
        return version
      } else {
        version = version.version
      }
    } else if (typeof version !== 'string') {
      throw new TypeError(`Invalid Version: ${version}`)
    }
```

and so, from my reading, passing in a full `SemVer` object is fine.

That said, it's not always clear when some manipulation in PureScript might cause a JS `instanceof` check to no longer be true, and so I would be a little happier passing in just the version string itself. (cc: @f-f)

f-f:
I'm a little confused here - did we uncover a bug? It we're able to write a test that shows funny behaviour then we can do something about it, but if things work fine for all possible testcases then I think the current implementation is alright?

thomashoneyman:
> but if things work fine for all possible testcases

I don't know if I'm bold enough to say _that_, but I do think things are working correctly right now.

@mikesol encountered crashes in the JS layer when constructing the `SemVerJS` record directly (ie. not via `parseSemVer`), but presumably that's because it's failing the `instanceof` check. So first the constructor checks if we passed a `SemVer`, which we didn't (we just passed a bare record), and then it checks if we passed a string (again, no, we passed a bare record), and then crashes.

[...]

PR #247, opened 2021-10-11
Treats semver as opaque type

mikesol:
This addresses https://github.com/purescript/registry/issues/246. By treating semver as an opaque type, there is no representation on the PS layer but rather all interactions happen over the FFI.


ci/src/Foreign/SemVer.purs
```diff
@@ -18,14 +18,7 @@ import Data.Argonaut as Json
...
-  { major :: Int
-  , minor :: Int
-  , patch :: Int
-  , prerelease :: Array String
-  , build :: Array String
-  , version :: String
-  }
+data SemVerJS
 
 newtype SemVer = SemVer SemVerJS
```
thomashoneyman: If we make this change (which I'm on board with), then there's no reason for the newtype. We can just `foreign import SemVer` and then implement the methods on it.

ci/src/Foreign/SemVer.purs
```diff
@@ -60,23 +53,29 @@ foreign import parseSemVerImpl :: String -> Nullable SemVer
 parseSemVer :: String -> Maybe SemVer
 parseSemVer = toMaybe <<< parseSemVerImpl
 
+foreign import majorImpl :: SemVerJS -> Int
 major :: SemVer -> Int
-major (SemVer v) = v.major
+major (SemVer v) = majorImpl v
```
thomashoneyman: These are simple enough that I'd say we just implement them directly in the FFI, ie.

```purs
foreign import major :: SemVer -> Int
```

without the indirection via `majorImpl`

thomashoneyman:
@mikesol just a couple comments for you, and then remember to run `purs-tidy`!

ci/src/Foreign/SemVer.purs
```diff
@@ -32,17 +30,18 @@ instance showSemVer :: Show SemVer where
   show = printSemVer
 
 instance decodeJsonSemver :: Json.DecodeJson SemVer where
-  decodeJson json = do
-    version <- Json.decodeJson json
-    note (Json.TypeMismatch $ "Expected version: " <> version) (parseSemVer version)
+  decodeJson = Json.decodeJson
+    >=> note
+    <$> (Json.TypeMismatch <<< append "Expected version: ")
+    <*> parseSemVer
```
thomashoneyman: Can we leave this the way it was? Thanks!

thomashoneyman (approved):

PR #253, opened 2021-10-17
Removes unsafe operation from main pipeline

mikesol:


f-f:
What's the reason for this change?

mikesol:
@f-f I should have asked before - what's the project's policy wrt stuff like `unsafePartial`? Usually, the way I've done it is that unsafe stuff is kept in library code and sketches/experiments but removed from production application code. If it's ok to be here given the project's style/guidelines, that's fine :+1:

f-f:
Oh I see what's happening and why now! Yes, removing `unsafePartial` and pushing it earlier in the pipeline sounds good to me. Thanks for doing this!

f-f (approved):

PR #257, opened 2021-10-23
Adds license parsing to main pipeline

mikesol:


ci/test/Registry/Scripts/LegacyImport/Stats.purs
```diff
@@ -12,6 +12,7 @@ import Foreign.SemVer (SemVer, parseSemVer)
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Schema (Repo(..))
+import Registry.License as RLicense
```
thomashoneyman, line 15:
```suggestion
import Registry.License as License
```

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -397,7 +368,7 @@ constructManifestFields package version address = do
     -- We can detect the license for the project using a combination of `licensee`
     -- and reading the license directly out of the Spago and Bower files (the
     -- CLI tool will not read from either file).
-    licenseeOutput <- detectLicense files
+    licenseeOutput <- detectLicense files ------------------ pull this out
```
thomashoneyman: When pulling this out, keep in mid that we still need to do the work below -- where we pull the licenses out of the Spago manifest and Bowerfile -- because we detect the licenses using a combination of `licensee` (which checks the LICENSE file and package.json file) and the licenses we pull out of the other manifest files. So the entire thing will need to be factored out.

mikesol:
@thomashoneyman when you get a chance, could you take a peek here and let me know if I'm going in the right direction? Basically, I created `Manifest.purs` to work with manifests from a project, and I'll probably tweak it for usage in the actual pipeline. That required importing a bunch of stuff from the legacy script, and I noticed that in general there was bit of cruft where stuff from the legacy script was imported into `API.purs`. I'm assuming modules should flow the other way - the legacy script and current pipeline should pull from the same files which "belong" to neither and are sorty of lib-ish/util-ish in nature.

If you sign off on this leg of work, the next step will be using `constructManifestFields` in the main pipeline followed by the license-checking stuff on the resulting manifest.

f-f:
I'll close this in favor of #288

Issue #251, opened 2021-10-13
Add license parsing code to the Registry pipeline


thomashoneyman:
The Legacy Import script contains code that checks a package's various manifest files to verify their licenses. In the registry we only accept packages with valid SPDX licenses.

When publishing or updating a package in the Registry pipeline we should verify its license. In other words, we should:

1. Verify that the manifest contains a valid SPDX license string, AND
2. If the repository contains LICENSE files or other manifest types (bower.json, package.json), we should check them as well and ensure they are the same as the manifest SPDX license string

I'm not completely sold on step 2 there, but it seems like we ought to try and do _some_ verification that the manifest license does match with the other license files in the repository, so that our pledge that "packages in the registry have valid SPDX licenses" is true, and that the SPDX license in the package is actually the one in the repository more generally.

...on the other hand, I really don't know the legal ramifications of all this. It brings to mind other questions:

1. What happens when the LICENSE file says BSD-3-Clause but the manifest says Apache-2.0? Should we reject this?
2. What happens if there are other manifest files in the repository -- for example, alternate backends like `pureerl`?

mikesol:
Hey! Some opinions:

1. Yes, we should reject packages with a mismatch between the actual license and the license as reported in the manifest (with a courtesy note to the maintainer about the rejection).
2. We should include in the documentation a list of manifest files we will parse, where we will look, and steps that will entail rejection.

Writing this, I'm actually not sure where the CI pipeline lives in this repo. I see SPDX license parsing happening in the legacy bower import, but where is the CI pipeline that will be used for new submissions? Has it been started yet?

Thanks!

thomashoneyman:
@mikesol the "registry pipeline" is the `Registry.API` module (it's a bit messy!)

mikesol:
@thomashoneyman if no one has started on this yet, I can do it.

f-f:
@mikesol the `Manifest` type [requires a SPDX identifier](https://github.com/purescript/registry/blob/763fe2757d26c39d64c20feada2f9c17323e2f10/ci/src/Registry/Schema.purs#L20), so we parse the license when reading the package manifest in (which includes the case when we are not converting from a Bowerfile)

mikesol:
I can take this on if no one else has grabbed it, just let me know 👍

thomashoneyman:
No one has grabbed it -- perhaps we can talk about it on the registry call today?

thomashoneyman:
[...]

PR #258, opened 2021-10-24
Adds basic license check

mikesol:
@thomashoneyman I have [a draft PR](https://github.com/purescript/registry/pull/257) with something along the lines of what we talked about in the registry meeting, but as I was putting it together I realized that the solution may be more simple than what we had discussed. This PR is my attempt to do that simple thing. If it is missing something, then let me know and I can add more logic.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -250,6 +250,7 @@ runChecks { metadata, manifest, absoluteFolderPath } = do
   case licenseFromLicensee of
     Left err -> throwWithComment $ "Could not find a license in the package: " <> err
     Right li -> when (Array.notElem (SPDX.print manifest.license) li) (throwWithComment $ "License from the manifest does not match the license detected by licensee. In manifest: " <> SPDX.print manifest.license <> ". Deteceted in package: " <> show li <> ".")
+
   log "Checking that the Manifest includes the `lib` target"
   libTarget <- case Object.lookup "lib" manifest.targets of
     Nothing -> throwWithComment "Didn't find `lib` target in the Manifest!"
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -13,7 +13,9 @@ import Effect.Ref as Ref
 import Foreign.Dhall as Dhall
 import Foreign.GitHub (IssueNumber)
 import Foreign.GitHub as GitHub
+import Foreign.Licensee as Licensee
 import Foreign.Object as Object
+import Foreign.SPDX as SPDX
 import Foreign.SemVer as SemVer
 import Foreign.Tar as Tar
 import Foreign.Tmp as Tmp
@@ -205,7 +207,7 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
           Left err -> throwWithComment $ "Could not convert Manifest to JSON: " <> err
           Right res -> pure res
 
-  runChecks metadata manifest
+  runChecks { metadata, manifest, absoluteFolderPath }
 
   -- After we pass all the checks it's time to do side effects and register the package
   log "Packaging the tarball to upload..."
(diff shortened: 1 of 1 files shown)


ci/src/Registry/API.purs
```diff
@@ -238,11 +240,16 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
...
+runChecks :: { metadata :: Metadata, manifest :: Manifest, absoluteFolderPath :: String } -> RegistryM Unit
+runChecks { metadata, manifest, absoluteFolderPath } = do
   -- TODO: collect all errors and return them at once. Note: some of the checks
   -- are going to fail while parsing from JSON, so we should move them here if we
   -- want to handle everything together
+  log "Checking that the SPDX license in the manifest corresponds to the one in the package"
+  licenseFromLicensee <- liftAff $ Licensee.detect absoluteFolderPath
+  case licenseFromLicensee of
+    Left err -> throwWithComment $ "Could not find a license in the package: " <> err
+    Right li -> when (Array.notElem (SPDX.print manifest.license) li) (throwWithComment $ "License from the manifest does not match the license detected by licensee. In manifest: " <> SPDX.print manifest.license <> ". Deteceted in package: " <> show li <> ".")
```
thomashoneyman, line 252: I added a comment in the other PR to the same effect, but this isn't quite enough -- `licensee` will detect from the LICENSE file, but we want to also pull the licenses out of the manifest files present in the repository in general, all of which is included in the `toManifestFields` function in the `LegacyImport` module.

We want to check the Bowerfile, Spago file, package.json file, and the LICENSE file and make sure they all match with the `purs.json` (manifest) file, but here we're only checking the LICENSE file.

The work to check all those files is done in the legacy import tool already so it's just a matter of pulling it out.

mikesol: I think I understand. Lemme write it up here and you can tell me if it is correct or not.

Currently, the `API.purs` file is looking for a file called `purs.json` with various bits of information including a license. In addition to this file, we want to parse other manifests which may also contain a license. Currently, the two additional manifest files we will parse are `bower.json` and `spago.dhall + packages.dhall`. So, for example, if we were writing it up in a README, it'd look something like this:

***
# License

Packages added to the registry must have a valid SPDX license, and that license must be indicated in a manifest file. We determine license validity using the following steps:

1. We look for, and if present parse, the following three manifest files to determine a package's license:
  - `purs.json`
  - `spago.dhall`
  - `bower.json`
2. We then use [`licensee`](https://www.npmjs.com/package/licensee) to find license files in the package.
3. We cross-reference the licenses found by `licensee` with the licenses specified in a package's manifest file(s). If there is a mismatch (ie too many licenses, too few licenses or an incorrect license) we will reject the package.

mikesol: @thomashoneyman ping regarding this just to make sure I'm on the right track - thanks!

thomashoneyman: Yes! Two clarifications:

1. We check the manifest file for the license for the project (the purs.json file). We then cross-reference against the other manifest files. In practice this is basically the same as what you’ve written, but until we fully decide what to do about mismatches I think it’s worth calling out that in our minds the purs.json manifest is the “true” one
2. Licensee checks the package.json file (if I remember correctly), so we’re checking that too.

mikesol: Would it be ok if I added the following?

> In the future, we will be using `purs.json` as the single source of truth for licenses for PureScript packages and will be deprecating `bower.json`, `package.json`, and `spago.dhall` support. We thus encourage all package authors to include their licenses in a `purs.json` file so as not to be dropped from a future edition of the registry.

thomashoneyman: I’m not sure about that — we may always check the other manifests, but in the future you will have to have a purs.json file to publish to the registry. We’re only auto-importing legacy packages

f-f:
One of the recommendations in #251 is the following:

> Finally, he recommended that -- even though it isn't strictly necessary -- when new packages are submitted to the registry in the future, we should reject packages that specify multiple licenses (MIT in package.json, BSD-3-Clause in purs.json) and ask them to join those licenses with a proper SPDX conjunctive like AND or OR.

..which means that we'd still need something like this check, right?

thomashoneyman:
Yea, we still want to have a check in the package upload pipeline that just verifies that the SPDX identifier you've asserted in your package manifest checks out against other manifest files / LICENSE files present. We won't fix it for you, but we'll reject packages that have an identifier in their `purs.json` that doesn't accurately represent the licenses in their repository.

ci/src/Registry/API.purs
toastal: > Packages added to the registry must have a valid SPDX license

Scenario: I have a lawyer and I have a new open source license type (not "UNLICENSED"). I have it documented in a `LICENSE` file so consumers understand how to comply. Would this make my package unregisterable? Or do I need to petition to get it in SPDX to be valid? Is creating my own license a good practice or idea? Probably not, but should it straight up not be able to be handled? I could see an argument where it should be a `Either SPDX CustomAndIKnowWhatImDoing`.

thomashoneyman: The issue is that we (the registry trustees) also would need to review the license to be sure that it is usable for open source — for example, open enough that we can host the code in the registry backend. Given that our bandwidth is already stretched it’s not something we’d like to take on if we can defer to a widely used standard (SPDX).

If a package does have this requirement in the future, then I’d encourage them to reach out to the registry trustees to see if this restriction can be changed at that time, or just not register the package. Package managers will still let you use the package even if it isn’t registered.

thomashoneyman:
Superseded by #719

PR #248, opened 2021-10-11
Disallow empty licenses and rewrite licenses one at a time

thomashoneyman:
I noticed two issues with our previous license handling; this PR fixes both:

1. We were rewriting licenses by pattern matching on arrays, ie. `[ "Apache 2.0" ]`, but since we now aggregate licenses from more places it's common to have several licenses in the array. Rewriting shouldn't fail; we should have been mapping over the array going license-by-license in the first place.
2. In license parsing, we'd pull in empty licenses as empty strings, which would cause them to fail SPDX parsing. Now, we require licenses to be non-empty strings while parsing, and we'll fail later on with a `MissingLicense` error (instead of failing on the empty string with a `BadLicense` error).

Current results:

```
Packages: 1,575 total (1,214 totally succeeded, 236 partially succeeded, 125 totally failed)
Versions: 11,521 total (10,350 totally succeeded, 1,171 totally failed)
Failures by error:
  manifestError: 1104 versions across 300 packages
    missingLicense: 455 versions across 143 packages
    badLicense: 362 versions across 53 packages
```

Previous results:

```
Packages: 1,575 total (1,200 totally succeeded, 248 partially succeeded, 127 totally failed)
Versions: 11,521 total (10,239 totally succeeded, 1,282 totally failed)
Failures by error:
  manifestError: 1,215 versions across 314 packages
    badLicense: 481 versions across 76 packages
    missingLicense: 452 versions across 141 packages
```

PR #263, opened 2021-11-02
Handle pagination in `getReleasesImpl`

### LINKED ISSUE #262: `getReleases` octokit binding doesn't handle pagination
While working on #226, I noticed that ~2000 packages were not self-contained within the registry index, as some of their dependencies didn't exist after `LegacyImport`. Digging in to it, @thomashoneyman and I realized that a lot of the unfulfilled dependencies did exist by manually checking GitHub repos.

Then we realized, we were only getting 30 results back for `purescript-dom`, even though there were 74 releases on GitHub.... Some digging and we found https://octokit.github.io/rest.js/v18#pagination, and realized that we weren't handling pagination correctly!

`getReleasesImpl` needs to be updated to properly handle fetching all releases.

colinwahl:
Fixes #262 by properly handling pagination in the `getReleasesImpl` FFI definition. See https://octokit.github.io/rest.js/v18#pagination for docs on pagination handling, this change is taken basically directly from there.

In https://github.com/purescript/registry/pull/248, this was the result of running `LegacyImport`: 

```
Packages: 1,575 total (1,214 totally succeeded, 236 partially succeeded, 125 totally failed)
Versions: 11,521 total (10,350 totally succeeded, 1,171 totally failed)
Failures by error:
  manifestError: 1104 versions across 300 packages
    missingLicense: 455 versions across 143 packages
    badLicense: 362 versions across 53 packages
```

Now, here is the result:

```
Packages: 1582 total (1126 totally succeeded, 307 partially succeeded, 149 totally failed)
Versions: 12479 total (10820 totally succeeded, 0 partially succeeded, 1659 totally failed)
Failures by error:
  manifestError: 1591 versions across 396 packages
    missingLicense: 1273 versions across 297 packages
    badDependencyVersions: 328 versions across 116 packages
  noDependencyFiles: 67 versions across 35 packages
  resourceError: 0 versions across 33 packages
```

Where we now check ~1000 more versions than before.

Note: for this to take affect, you'll need to clear out your `.cache` directory, or at least clear out the `releases` entries within the cache.

Note: I believe the difference in package counts here is because of packages that were recently added to the registry - I didn't run it on `master` because it takes a long time, but I could if needed.

### DIFF
--- a/ci/src/Foreign/GitHub.js
+++ b/ci/src/Foreign/GitHub.js
@@ -11,12 +11,12 @@ exports.mkOctokit = function () {
 };
 
 exports.getReleasesImpl = function (octokit, { owner, repo }) {
-  return octokit.repos
-    .listTags({
+  return octokit
+    .paginate(octokit.repos.listTags, {
       owner,
       repo,
     })
-    .then(({ data }) => {
+    .then((data) => {
       return data.map((element) => {
         return {
           name: element.name,


thomashoneyman (approved):

thomashoneyman:
Great catch and thanks for fixing this!

thomashoneyman:
I’m a bit surprised by so many new missingLicense errors, but it wouldn’t be related to the change you’ve made here.

f-f:
Great catch and thanks for fixing this!!

colinwahl:
I was a bit surprised by those `missingLicense` errors as well. 

I went and manually verified a few versions across a few different packages, and indeed they didn't have a license, so maybe it is correct.

PR #264, opened 2021-11-03
`RegistryIndex` graph utilities: Compute maximal self-contained `RegistryIndex`, topologically sort a self-contained `RegistryIndex`

### LINKED ISSUE #214: Produce a graph of package versions and their dependencies from the registry index

### LINKED ISSUE #226: Consider updating check that dependencies are self-contained
Currently, we check that all package dependencies are part of the registry right after fetching the bowerfile:

https://github.com/purescript/registry/blob/daf1399c458b8159deb473358a707c660af2fc27/ci/src/Registry/Scripts/BowerImport.purs#L133

However, this means that package versions could be dropped from the registry later (for example, when being converted to `Manifest`s), and if all package versions for a package are dropped then the package isn't in the registry at all.

For this reason, we could end up considering a package valid because all of its dependencies are in the registry, only to later on drop one of its dependencies from the registry.

We could avoid this problem by moving the check for self-contained dependencies for packages to be the final check.

colinwahl:
Fixes #214 and #226

For #214, this PR implements the following functions:
- `toPackageGraph :: RegistryIndex -> PackageGraph`
- `inOrder :: RegistryIndex -> Array Manifest`

These can to process the `Manifest`s contained `RegistryIndex` in topological order.
Note: Both of these functions will crash when there are unsatisfied dependencies within the `lib` target of a `Manifest`. They are meant to be used on a self-contained `RegistryIndex` - see the `checkRegistryIndex` function.

For #226, this PR implements the `checkRegistryIndex` function, which produces a maximally self-contained  `RegistryIndex`, along with recording which `Manifest`s had unsatisfied dependencies. I've tried my best to make this quick, and it currently takes 6-8seconds to solve all of the constraints resulting from the legacy import script.
The `checkRegistryIndex` function replaces the `selfContainedDependencies` check that existed in the `LegacyImport` script. It is executed at the very end of `LegacyImport`, and ensures that the `RegistryIndex` that is written to `registry-index` is self-contained.

Here is the new output from running the legacy import script:

```
Writing exclusions file...
Packages: 1578 total (1130 totally succeeded, 299 partially succeeded, 149 totally failed)
Versions: 11538 total (10022 totally succeeded, 0 partially succeeded, 1516 totally failed)
Failures by error:
  manifestError: 1449 versions across 389 packages
    missingLicense: 1156 versions across 292 packages
    badDependencyVersions: 308 versions across 114 packages
    badVersion: 50 versions across 34 packages
  noDependencyFiles: 66 versions across 34 packages
Constructing self-contained registry index...
Writing unsatisfied dependencies file...
1958 manifest entries with unsatisfied dependencies
Writing registry index to registry-index
Done!
```

in particular, it now includes the lines

```
Constructing self-contained registry index...
Writing unsatisfied dependencies file...
1958 manifest entries with unsatisfied dependencies
```

All unsatisfied dependencies are written to `unsatisfied-dependencies.json`.


colinwahl:
1958 seems like a lot of manifests to be dropped due to unsatisfied dependencies :O 

However, I've verified quite a few random ones by hand, and they seem to be correctly dropped. Here's how I did it:

- Check `unsatisfied-dependencies.json` and follow the graph of unresolved dependencies until you can't find the root in that file
- Check `bower-exclusions.json` for why that root dependency failed

The most common reason I am seeing for the ones I checked is `MissingLicense`.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,200 @@
...
+        We need to construct the RegistryIndex from legacy files, and then take all of those
+        entries and prune them to only include valid entries at the end.
+        We do this after the rest of LegacyImport because we want to avoid including a package
+        but then drop one of it's dependencies for some reason later.
+
+Complications:
+Package dependencies specify a `Range`, not a concrete `SemVer`.
+If we have to drop a (PackageName, SemVer), we need to make sure we don't drop entries
+which relied on it until we have checked that there are no other valid versions which satisfy the range.
+For any given dependency, we want to use the `maxSatisfying` valid version that satisfies the range.
```
f-f: I think we generally cannot trust version bounds (they are either plain wrong, or too strict or too lax), so here we should avoid checking if a package version satisfies a certain range. This is also because a version bound that might not include any version today might be lax enough to be able to include a version of a dependency that will be published in the future.
So we should just check that every dependency of a package has at least one version published, regardless if it satisfies the range or not.
I suspect this will also simplify (and possibly speed-up) the implementation.

thomashoneyman: I agree that it's OK to just verify that all dependencies of a package are present in the registry at any version. 

I also think that the version bounds check here is valuable because it would be nice to emit a warning when uploading a package if we can easily detect that the version bounds declared don't actually exist in the registry. (Without failing the package, because there are ways to use it without using the declared bounds, such as the package sets.)

My mistake on this one, @colinwahl, as I indicated to you we should check the ranges. We should pull that back and only check a dependency by verifying it exists at any version in the index.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,200 @@
...
+      go { satisfied, constraints }
+
+  isSolved :: Map PackageName (Array SemVer) -> PackageWithRange -> Boolean
+  isSolved solved { package, range } = fromMaybe false do
+    versions <- Map.lookup package solved
+    Just $ Array.any (flip SemVer.satisfies range) versions
+
+type PackageGraph = Graph (Tuple PackageName SemVer) Manifest
+
+-- We will store `maxSatisfying` for all dependencies.
```
thomashoneyman: If we don't care about ensuring correct versions, then I suppose we only need to store the package dependencies without bothering with the particular version solved for its dependencies.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,200 @@
...
+    -- If there was progress, then we need to iterate through constraints again.
+    if Array.null progress then do
+      { satisfied, unsatisfied: constraints }
+    else
+      go { satisfied, constraints }
+
+  isSolved :: Map PackageName (Array SemVer) -> PackageWithRange -> Boolean
+  isSolved solved { package, range } = fromMaybe false do
+    versions <- Map.lookup package solved
+    Just $ Array.any (flip SemVer.satisfies range) versions
```
thomashoneyman: If we're only looking at package dependencies and not versions, then does this simplify to the below?

```suggestion
  isSolved solved { package, range } = fromMaybe false $ Map.lookup package solved
```

ci/src/Registry/PackageGraph.purs
colinwahl: Sure, I can change it up a bit. @f-f is right that it would make this check quicker :)

For my clarification - why are dependency version bounds ofter wrong? Is it because most packages are published to be used in a package set, that doesn't care about those bounds?

My only concern is that it may be more likely to get a cycle in the graph if we are just collecting package dependencies in the `lib` target, but I'm not sure if it happens in practice. I'll see!

f-f: Version bounds are often wrong because they are typed by humans that might not test all the version that the bound admits, or because they allow for future versions that could break the build, or because they don't include package versions that indeed work fine.
As a reference, very many package sets that exist would not be admitted by the version bounds of every package.

The concern about cycles in the graph is very relevant though! I think in principle there's no reason why we shouldn't allow circular dependencies (your build would fail anyways if you'd try to build such packages), so it looks like this is only a concern for our next step which is the topological sorting for uploading packages to Bower? If that's the case then we can take care of detecting cycles that would make us fail the self-containment check (which is a subset of all the possible cycles) when sorting all the versions for upload. We can take care of this either in this PR or in a subsequent one.

colinwahl: @thomashoneyman and I updated `checkRegistryIndex` to only check that each dependency of a package exists, without checking that versions bound match.

To make `toPackageGraph` without checking any version bounds, we edited `toPackageGraph` to take each `(PackageName, SemVer)` and construct edges to each `(PackageName, SemVer)` for each `PackageName` listed in the `dependencies` array.

I'm not sure if this will cause issues with cycles but a quick glance at the output seemed fine. We would like to think more deeply on this & add tests soon.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,188 @@
...
+
+    -- When there is no progress, we have satisfied all of the constraints possible.
+    -- If there was progress, then we need to iterate through constraints again.
+    if Array.null progress then do
+      { satisfied, unsatisfied: constraints }
+    else
+      go { satisfied, constraints }
+
+  isSolved :: Map PackageName (Array SemVer) -> PackageName -> Boolean
+  isSolved solved package = Map.member package solved
```
colinwahl, line 120: Now we consider a dependency solved if the `PackageName` has at least one satisfied version.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,188 @@
...
+          $ flip bind resolveDependency
+          $ map (lmap unsafeParsePackageName)
+          $ Object.toUnfoldable
+          $ _.dependencies
+          $ fromJust' (\_ -> unsafeCrashWith "Manifest missing lib target")
+          $ Object.lookup "lib" manifest.targets
+    Tuple manifest deps
+
+  resolveDependency :: Tuple PackageName Range -> Array (Tuple PackageName SemVer)
+  resolveDependency (Tuple dependency _) = do
```
colinwahl, line 166: Instead of resolving to a particular `SemVer`, we construct an each to each version of the dependency.

I'm not sure if this is correct or will cause issues with cycles, but I need to do more thinking on it.

colinwahl:
After changing the `checkRegistryIndex` function, here is the output from the `LegacyImport` script:

```
Writing exclusions file...
Packages: 1582 total (1133 totally succeeded, 299 partially succeeded, 150 totally failed)
Versions: 11544 total (10028 totally succeeded, 0 partially succeeded, 1516 totally failed)
Failures by error:
  manifestError: 1449 versions across 389 packages
    missingLicense: 1156 versions across 292 packages
    badDependencyVersions: 308 versions across 114 packages
    badVersion: 50 versions across 34 packages
  noDependencyFiles: 66 versions across 34 packages
  resourceError: 0 versions across 33 packages
Constructing self-contained registry index...
Writing unsatisfied dependencies file...
121 manifest entries with unsatisfied dependencies
Writing registry index to registry-index
Done!
```

Now there are only 121 manifest entries with unsatisfied dependencies - checking the `unsatisfied-dependencies.json` file seems correct in that all of the missing dependencies I checked not in the registry index, and they appeared in the `bower-exclusions.json` file for missing licenses.

thomashoneyman:
Also, we're considering writing a test that will iterate through the resulting ordered list to verify it really is sorted. It'd work by walking through the array and for each entry:

1. Add the entry to a set of already-visited package names
2. If the entry has dependencies, verify that all dependencies already exist in the set of visited package names

If any entry has dependencies that haven't already shown up in the list, then the sorting is incorrect. We could have this test function run in the test files but also be available as a check done during the `LegacyImport` if we aren't 100% sure about the sorting being correct in the presence of cycles in the package graph.

colinwahl:
We just added a quick sanity check to the legacy import script to make sure our topological sort actually worked, and it seems to be working!

Here's the latest output:

```
Packages: 1582 total (1213 totally succeeded, 243 partially succeeded, 126 totally failed)
Versions: 12493 total (11192 totally succeeded, 0 partially succeeded, 1301 totally failed)

Failures by error:
  manifestError: 1233 versions across 309 packages
    missingLicense: 503 versions across 148 packages
    badLicense: 415 versions across 55 packages
    badDependencyVersions: 330 versions across 118 packages
  noDependencyFiles: 67 versions across 35 packages
  resourceError: 0 versions across 31 packages

466 manifest entries with unsatisfied dependencies
```

ci/src/Foreign/SemVer.purs
```diff
@@ -31,8 +33,8 @@ instance showSemVer :: Show SemVer where
 
 instance decodeJsonSemver :: Json.DecodeJson SemVer where
   decodeJson json = do
-    version <- Json.decodeJson json
-    note (Json.TypeMismatch $ "Expected version: " <> version) (parseSemVer version)
+    version' <- Json.decodeJson json
```
colinwahl, line 34: There was a shadowed name here, I took care of it here.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,202 @@
...
+      map (\{ package: { package, version }, dependencies } -> { package, version, dependencies })
+        allUnsatisfied
+
+  { index, unsatisfied }
+  where
+  go :: ConstraintArgs -> ConstraintResult
+  go args = do
+    let
+      -- Check if the given constraint is satisfied
+      -- If so, add package/version to satisfied & record progress
```
thomashoneyman, line 92: Could we expand a little on the use of `progress` here -- what it's tracking and why it's necessary? It appears to be a way to terminate in the case we are no longer _making_ progress (ie. we've reached a fix point), but that wasn't immediately obvious to me until I'd read through everything. It'd be nice to clarify that a little more right up front in these comments.

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,202 @@
...
+    unsatisfied =
+      map (\{ package: { package, version }, dependencies } -> { package, version, dependencies })
+        allUnsatisfied
+
+  { index, unsatisfied }
+  where
+  go :: ConstraintArgs -> ConstraintResult
+  go args = do
+    let
+      -- Check if the given constraint is satisfied
```
thomashoneyman, line 91: Would it be correct to say a "constraint" is a set of a package's dependencies that still need to be satisfied? And that we consider the registry index 'checked' once we have no more constraints to satisfy?

thomashoneyman (approved):
I have a couple of minor questions around comments you've left in the code, but the implementation looks fantastic and it's easy to read. Thank you for putting this together!

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,202 @@
...
+
+  resolveDependency :: Tuple PackageName Range -> Array (Tuple PackageName SemVer)
+  resolveDependency (Tuple dependency _) = do
+    depVersions <- maybe [] Array.singleton (Map.lookup dependency allVersions)
+    version <- depVersions
+    pure $ Tuple dependency version
+
+-- | Produces a topologically sorted list of Manifests.
+inOrder :: RegistryIndex -> Array Manifest
+inOrder index = do
```
f-f: Apologies for the nameshedding :smile: 

```suggestion
sort :: RegistryIndex -> Array Manifest
sort index = do
```

ci/src/Registry/PackageGraph.purs
```diff
@@ -0,0 +1,202 @@
...
+  getDeps :: Manifest -> Array PackageName
+  getDeps = map unsafeParsePackageName <<< Object.keys <<< getLibDependencies
+
+  foldFn :: Tuple Boolean (Set PackageName) -> Manifest -> Tuple Boolean (Set PackageName)
+  foldFn (Tuple valid visited) manifest = do
+    let newSet = Set.insert manifest.name visited
+    if all (flip Set.member visited) (getDeps manifest) then
+      Tuple (valid && true) newSet
+    else
+      Tuple false newSet
```
f-f: It seems that we're only introducing this code as a sanity check and/or for tests?
I wonder if we could skip all of this by just wrapping the `Array Manifest` that we get from the toposort into a opaque newtype?
I.e.
```purescript
newtype SortedRegistry = SortedRegistry (Array Manifest)
```

f-f (approved):
I left a couple of small comments, but otherwise this looks great! :clap:

ci/test/Registry/Index.purs
```diff
@@ -89,9 +93,34 @@ testRegistryIndex = Spec.before runBefore do
...
+
+  getLibDependencies :: Manifest -> Object Range
+  getLibDependencies manifest =
+    ( fromJust'
+        (\_ -> unsafeCrashWith "Manifest has no 'lib' target in 'isOrdered'")
+        (Object.lookup "lib" manifest.targets)
+    ).dependencies
+
+  unsafeParsePackageName :: String -> PackageName
+  unsafeParsePackageName = fromRight' (\_ -> unsafeCrashWith "PackageName must parse") <<< PackageName.parse
```
thomashoneyman: These are now duplicated with the same functions in `PackageGraph`, but I think it's OK since these are simple, internal-only, unsafe utilities.

Issue #215, opened 2021-09-08
Kickoff the API pipeline in batch mode for Bower packages


f-f:
Once we produce a graph in #214 we should be able to take these packages in topological order and then uploads them one after the other, running locally and using the current `Registry.API` code.

f-f:
As of #288 we are now able to successfully import all the core packages, but I'll leave this open until we have a PR that adds all the Metadata files here.

thomashoneyman:
> until we have a PR that adds all the Metadata files here.

Do you mind clarifying this? I'm not sure what this is referencing.

f-f:
If you run the legacy-import script locally, it will:
1. Upload all the packages
2. Update a local copy of the `registry-index` repo with the new info
3. Update a metadata file for each package, where we store the hashes and stuff. These files are our source of truth! I.e. if a package version is there THEN it's in the registry and (should be) available for download.

What I'm referencing above is the product of step 3, which is a bunch of files in the `metadata` folder, which right now we don't have since no package has been uploaded yet 🙂 
(Well, a bunch of tarballs are on DO, but we should nuke them until we commit the metadata here. However that's not done yet because things are in flux and running the import takes a long while)

thomashoneyman:
This has been done, and we’ll continue to run uploads regularly (#425), so I’m going to close this.

PR #279, opened 2021-12-23
Finish implementation for the mass-import of Legacy packages

f-f:
This PR implements the last bits that we were missing in order to have a mass-importer of "legacy" packages into the new Registry.
I had to shuffle a few things around (e.g. adjust what's returned by the topological sort, or how we print SemVer versions), but now it's possible to run
```bash
spago run -m Registry.Scripts.LegacyImport
```
...and the script will:
1. read all the legacy packages in
2. upload all the versions of the core packages to the storage backend
3. write out the `Metadata` (with the correct hashes) and the `registry-index` files

(the files produced by the last step have to be committed manually, the auto-commit is only for the CI)

Right now we limit this to only work on core packages (until we sort out the license issues for the rest of the packages), but extending this to all packages is only a matter of removing the filtering in the `LegacyImport` module.

We could consider this a fix for #215, but I guess we can leave that open until we actually push all the packages there?


ci/src/Registry/PackageGraph.purs
```diff
@@ -23,6 +23,12 @@ import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Schema (Manifest)
 
+type Node =
+  { manifest :: Manifest
+  , packageName :: PackageName
+  , version :: SemVer
+  }
```
thomashoneyman: Do you mind explaining why we need to introduce this new `Node` type? After all, a `Manifest` already includes the package name and version with the types used in `Node`:

https://github.com/purescript/registry/blob/05fa80c2bd1b95b9f470f9db8f223ba89577865a/ci/src/Registry/Schema.purs#L17-L23

From what I can tell, we don't need for a `Node` and the included `Manifest` to ever have different package names or versions, so it seems like it's just duplicated information.

ci/src/Registry/Scripts/LegacyImport/Manifest.purs
```diff
@@ -0,0 +1,136 @@
...
+      let
+        rewrite = case _ of
+          "Apache 2" -> "Apache-2.0"
+          "Apache-2" -> "Apache-2.0"
+          "Apache 2.0" -> "Apache-2.0"
+          "BSD" -> "BSD-3-Clause"
+          "BSD3" -> "BSD-3-Clause"
+          "BSD-3" -> "BSD-3-Clause"
+          "3-Clause BSD" -> "BSD-3-Clause"
+          other -> other
```
thomashoneyman, line 45: Unrelated to this PR, but I think we should disable any modification of package information when uploading packages, at least until we have the legal side of things confirmed.

thomashoneyman:
Also unrelated to this PR: are the hashes we're using compatible with what Nix requires? I can't remember if Nix has a specific hash format. If so, it would be _really_ nice to be compatible with Nix out of the box.

thomashoneyman (changes requested):
Everything looks good except for the introduction of the `Node` type. Once we chat about whether & why that's necessary to add, or revert it if it isn't necessary, then 👍 from me.

ci/src/Registry/Scripts/LegacyImport/Manifest.purs
f-f: We are limiting the upload to core packages for now, so we probably don't really have to worry about this yet:
https://github.com/purescript/registry/blob/e55de07f6f5c91390d77d0d4d07edad64b7d6fcb/ci/src/Registry/Scripts/LegacyImport.purs#L64-L70

ci/src/Registry/PackageGraph.purs
f-f: I am actually not sure why I went on this trip of adding this type - you're absolutely right, it's useless and I removed it :smile:

f-f:
Thanks for the review @thomashoneyman! Everything should be addressed now.

> Also unrelated to this PR: are the hashes we're using compatible with what Nix requires?

When implementing the hashing I made sure that our code computes the same result as the `sha256sum` command (but don't trust me on this, we actually need to add some tests :smile:).
It looks like that's compatible with nix hashes:
```bash
$ nix-hash --flat --type sha256 package.json
d6d115122eba9d81410cd6cd7b88205f8e6515ac88d30aa29a5bb17ef4908580
$ sha256sum package.json
d6d115122eba9d81410cd6cd7b88205f8e6515ac88d30aa29a5bb17ef4908580  package.json
```

thomashoneyman (approved):

PR #282, opened 2021-12-26
Add description

### LINKED ISSUE #124: Manifest description

f-f:
In this PR we add an optional `description` field for package Manifests - fixing #124 - and some code to import any description present in the Bowerfiles when processing legacy packages.

Note: this is based on #279 so that should be merged first


thomashoneyman (approved):
I switched the base on this because it contains a number of commits also present in your other PR. 👍 from me!

ci/src/Registry/Schema.purs
```diff
@@ -19,6 +19,7 @@ type Manifest =
   , license :: License
   , repository :: Repo
   , targets :: Object Target
+  , description :: Maybe String
```
thomashoneyman, line 22: Actually -- now that I think of it, we may not want to rely on the `EncodeJson` instance here, because it's going to encode this field as `null` when we more likely would like to omit the field altogether.

It's way more common to set a field to `null` by not including it rather than explicitly setting `null`, and any test that reads in a manifest missing this field and then writes this type is going to write out `null`.

Overall we really ought to switch to something like `codec-argonaut` and have explicit codecs for registry types.

f-f: Yeah, I came to the same realization once I saw the CI failure :smile:

f-f:
Update:
- the `Manifest` is now a `newtype`, which doesn't make it less convenient for the rest of the application, but allows for a custom `EncodeJson` instance, which is now able to omit the `description` key
- I moved the "roundtrip example manifests" test from "a custom script we only run in CI" to being integrated in our test suite

ci/src/Registry/Schema.purs
JordanMartinez: I'm currently working on this.

PR #285, opened 2022-01-04
Upgrade registry dependencies

thomashoneyman:
This small PR upgrades dependencies across the board to up-to-date versions as a housekeeping chore.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -280,8 +280,8 @@ sha256sum :: String -> Aff String
 sha256sum filepath = do
   fileBuffer <- FS.readFile filepath
   liftEffect do
-    newHash <- Hash.createHash Hash.SHA256
-    fileHash <- Hash.update newHash fileBuffer
+    newHash <- Hash.createHash "sha256"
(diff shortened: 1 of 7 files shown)


ci/src/Registry/API.purs
```diff
@@ -280,8 +280,8 @@ sha256sum :: String -> Aff String
 sha256sum filepath = do
   fileBuffer <- FS.readFile filepath
   liftEffect do
-    newHash <- Hash.createHash Hash.SHA256
-    fileHash <- Hash.update newHash fileBuffer
+    newHash <- Hash.createHash "sha256"
```
thomashoneyman, line 283: This is due to recent changes in the `node-crypto` library.

f-f: We don't have tests for this hashing code, and even with this mechanical change I wonder: is this still going to work?
It might be a good moment to add a simple test ensuring that our `sha256sum` function returns the same hash as the `sha256sum` command

f-f: We decided to tackle this in a followup

f-f (approved):

PR #286, opened 2022-01-08
Make the legacy import script resumable and more verbose

f-f:
As per title, this PR:
- adds some more logging to the legacy import script, to ease debugging
- makes it resumable, by not trying to upload packages that are already present in the Metadata and only considering new ones


ci/src/Foreign/SPDX.purs
```diff
@@ -24,6 +24,9 @@ instance decodeJsonSPDXLicense :: DecodeJson License where
 instance encodeJsonSPDXLicense :: EncodeJson License where
   encodeJson = encodeJson <<< print
 
+instance Show License where
+  show = print
```
f-f, line 28: I had to add a bunch of these `Show` instances, and I was thinking that we could remove all the various `print` functions, since these instances have the same purpose

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -65,12 +65,27 @@ main = Aff.launchAff_ do
...
+      GitHub { repo: "purescript-text-encoding" } -> Just manifest
+      GitHub { repo: "purescript-typelevel" } -> Just manifest
+      GitHub { repo: "purescript-sized-vectors" } -> Just manifest
+      GitHub { repo: "purescript-nonempty-array" } -> Just manifest
+      GitHub { repo: "purescript-colors" } -> Just manifest
+      GitHub { repo: "purescript-eff-functions" } -> Just manifest
+      GitHub { repo: "purescript-node-events" } -> Just manifest
+      GitHub { repo: "purescript-nonempty-array" } -> Just manifest
+      GitHub { repo: "purescript-aff-promise" } -> Just manifest
+      GitHub { repo: "purescript-naturals" } -> Just manifest
```
thomashoneyman, line 88: Is this a running list of already-uploaded packages?

thomashoneyman: Each of these are here because they were required for `contrib`?

ci/src/Foreign/SPDX.purs
thomashoneyman: Are we using this for debugging, or for display purposes? Traditionally speaking, `Show` instances meant for debugging would include the constructor, ie.

```purs
instance Show License where
  show (License license) = "(License " <> license <> ")"
```

If we need it for display purposes, then maybe we could have our own `Display` class for that purpose? I feel like Show gets overused for different purposes 😬

ci/src/Foreign/SemVer.purs
```diff
@@ -96,3 +96,6 @@ parseRange original = do
 
 printRange :: Range -> String
 printRange (Range r) = r
+
+instance Show Range where
+  show = printRange
```
thomashoneyman:
```suggestion
  show = printRange

```

ci/src/Registry/Schema.purs
```diff
@@ -170,3 +170,10 @@ mkNewMetadata location = { location, releases: mempty, unpublished: mempty }
 addVersionToMetadata :: SemVer -> VersionMetadata -> Metadata -> Metadata
 addVersionToMetadata version versionMeta metadata =
   metadata { releases = Object.insert (SemVer.version version) versionMeta metadata.releases }
+
+isVersionInMetadata :: SemVer -> Metadata -> Boolean
+isVersionInMetadata version metadata = versionPublished || versionUnpublished
+  where
+  versionStr = SemVer.version version
+  versionPublished = isJust $ Object.lookup versionStr metadata.releases
+  versionUnpublished = isJust $ Object.lookup versionStr metadata.unpublished
```
thomashoneyman:
```suggestion
  versionUnpublished = isJust $ Object.lookup versionStr metadata.unpublished

```

thomashoneyman (approved):

ci/src/Registry/Scripts/LegacyImport.purs
f-f: Yes! Contrib packages are now self-contained, but it was not always like that, so we need to pull in this list of other packages to upload older versions.

This is all temporary though, until we fire the big import!

ci/src/Foreign/SPDX.purs
f-f: We are using this in logs, yes. I'd not like to add the constructor, since that adds so much noise that hinders readability while looking at logs - but I of course feel the pull in different directions with `Show`. I think in such cases we might want to have `Debug` instances for debugging.
Though I think this is good enough for now, my complaint was more about all the differently-named `print` functions :smile:

PR #288, opened 2022-01-09
Reuse the legacy package import process in the API pipeline

### LINKED ISSUE #251: Add license parsing code to the Registry pipeline

f-f:
Fixes #251 

Supersedes #257

Note: this is based on #286, so that should be merged first.


ci/src/Registry/API.purs
```diff
@@ -159,34 +158,36 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
...
+
+      runManifest =
+        Except.runExceptT <<< Except.mapExceptT (liftAff <<< map (lmap printErrors))
+
+      gatherManifest :: ExceptT ImportError Aff Manifest
+      gatherManifest = do
+        manifestFields <- Manifest.constructManifestFields (RawPackageName $ show packageName) (RawVersion ref) address
+        Except.mapExceptT liftError $ Manifest.toManifest packageName metadata.location semVer manifestFields
+
+    runManifest (gatherManifest) >>= case _ of
```
```suggestion
    runManifest gatherManifest >>= case _ of
```

thomashoneyman (approved):

Issue #79, opened 2020-10-07
Malicious code and legal policies


f-f:
In https://github.com/purescript/registry/pull/76#discussion_r495607360 we figured that we probably want to be able to remove "malicious code" from the registry. We also figured that [NPM has already a policy on this that we might want to adopt](https://www.npmjs.com/policies/open-source-terms#acceptable-content), together with [other legal policies](https://www.npmjs.com/policies/)

So we should:
- read through these policies
- port inside this repo (e.g. in a `legal` folder) the policies we like
- patch the README to point to the legal documents where relevant

thomashoneyman:
Some additional legal things we need to make sure we include are described here:
https://github.com/purescript/registry/issues/251#issuecomment-1007839570

thomashoneyman:
Our legal advisor especially emphasized the need for DMCA protections.

PR #289, opened 2022-01-10
Use SPDX.And when parsing manifests

thomashoneyman:
In #79 we've discussed the need to join package licenses with the SPDX `AND` conjunctive when producing new manifest files during the legacy import process. During an unrelated code review I noticed we are currently using the much less restrictive `OR`. This PR updates our code to use `AND` instead.


f-f (approved):
Thank you for spotting this!

thomashoneyman:
Do we need to re-run the import tool now that this is in place? Were any packages hitting this code path?

f-f:
I am not sure if any package was affected, but I'll rerun the import tool with _all_ the packages (i.e. not only core) once we fix #290

PR #291, opened 2022-01-11
Switch to using value-based Json codecs

JordanMartinez:
Updates the codebase to use `purescript-codec-argonaut`.

In ce1c8bdcbc66d168821680af86e06f605b193e50, there are a few codecs I wasn't sure how to port properly. The others are likely correct, but a few migration may still have issues due to the usage of optional fields.

I haven't tested this out yet, but it should be good enough for an initial review.

### DIFF
--- a/ci/spago.dhall
+++ b/ci/spago.dhall
@@ -10,6 +10,8 @@
   , "arrays"
   , "avar"
   , "bifunctors"
+  , "codec"
+  , "codec-argonaut"
   , "console"
   , "control"
   , "crypto"
@@ -42,6 +44,7 @@
   , "parallel"
   , "partial"
   , "prelude"
+  , "profunctor"
   , "profunctor-lenses"
   , "psci-support"
   , "refs"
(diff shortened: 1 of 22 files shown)


f-f:
We discussed about moving to these codecs in the weekly call last week - I realized that I wouldn't like to move to codecs because:
- I find them ultimately harder to read than handwritten encode/decode instances
- we might want asymmetric instances, e.g. right now we have one for `Manifest`

JordanMartinez:
I talked about this with Fabrizio and Thomas via the Registry call today. In short, this isn't worth doing _now_ for a few reasons:
1. while a subjective reason, "not being familiar" with codecs means one is less likely to want to work on the codebase because there's friction with using the code (and with having to add new codecs if needed). Since this work is already hard enough, we don't want to make it harder.
1. there are cases where we don't want 'bidirectional' codecs. In some situations, we will parse something in one format and output it as a different format. Writing the equivalent would require writing two codecs, which could get confusing as to which one to use
1. things are still in alpha-quality, so the types themselves may change. Needing to update codecs at the same time isn't ideal. (For example, types using the `Generic`-based `DecodeJson`/`EncodeJson` codecs)
1. there are other higher priority items, so spending time on this isn't worth it now

Once things settle down and a spec is written that is unlikely to change frequently (or that much), then this idea could be re-examined.

I'm in agreement with them on all of this.

PR #294, opened 2022-01-13
Introduce a Json class for the registry

thomashoneyman:
This PR replaces `argonaut`, `argonaut-codecs`, and `argonaut-generic` with a `RegistryJson` class that provides more sensible behavior for how we handle JSON encoding and decoding. There are several reasons to introduce this class:

1. We can provide better encoding and decoding for how we handle commonly-used data types like `Map` and `Maybe`.
   For example, we encode `Map`s as JSON objects (versus Argonaut's approach of encoding them as arrays of tuples); this saves us from plenty of tedious conversions to and from maps when encoding and decoding, and helps us prefer `Map` throughout the application instead of `Foreign.Object`. As another example, we by default do not insert fields into objects when their value is `Nothing`, rather than fill an object with all its fields set to `null`.
2. We can provide instances for data types we did not write, like `BigInt` or `RFC3339String`
   This hasn't been an issue so far, but it's an advantage that relates to (1): when we adopt and use a type defined elsewhere, we can write the instance for it directly rather than having to wrap it in a newtype first and write instances for that.

Some of this value can also be provided by switching to value-based codecs, but the drawbacks of those were noted in #291 and we decided against that direction at the moment. Right now we're basically preserving the feel of using Argonaut, but with instances that match better with the way we write our code.

I have not provided a `Generic` interface for this class; that was only used in the non-user-facing `Error` module for the legacy import tool, and I find that generics (while convenient) tend to bloat things up significantly in the generated code and it'd be a finicky module to write. I can add it though, if we want it!


ci/src/Registry/Json.purs
```diff
@@ -0,0 +1,267 @@
...
+    where
+    toTupleArray :: Map String a -> Array (Tuple String a)
+    toTupleArray = Map.toUnfoldable
+
+  decode = map (Map.fromFoldable <<< toTupleArray) <=< decode
+    where
+    toTupleArray :: Object a -> Array (Tuple String a)
+    toTupleArray = Object.toAscUnfoldable
+
+else instance (Ord k, Newtype k String, RegistryJson v) => RegistryJson (Map k v) where
```
natefaubion: Could you get rid of the instance chain altogether by using `Coercible k String` instead of `Newtype` (since String is trivially coercible to itself)?

ci/src/Registry/Schema.purs
```diff
@@ -24,21 +24,35 @@ newtype Manifest = Manifest
...
   { dependencies :: Object Range
   , sources :: Array String
   }
 
+derive instance Newtype Target _
+derive newtype instance Eq Target
+derive newtype instance Show Target
+
+-- We write a manual instance here to control the ordering of the resulting
+-- object, which we don't want to be alphabetical
```
f-f, line 49: Why no alphabetical ordering here?

f-f (approved):
This is lovely! Great work :clap:

ci/src/Registry/Schema.purs
thomashoneyman: Just because the `dependencies` list can be quite long, and it's nice to see the (much shorter) `sources` list right up front. But we can remove the newtype to switch that back.

f-f: I don't mind either, I'll just note that these manifests are not really meant to be read/written manually by humans :)

ci/src/Registry/Json.purs
thomashoneyman: I can get rid of it, but because `Coercible` requires the newtype constructor to be in scope I had to shuffle around a few types so that the constructors can be imported as needed.

ci/src/Registry/Schema.purs
thomashoneyman: I'm not sure I agree on that part -- humans will read them in their projects, no? Same as Bowerfiles? I'd still like them to be as readable as possible.

f-f: They don't even have to be versioned - we only require them to be present in tarballs, but a package manager could (and Spago probably will) generate them on the fly when publishing the package

f-f: They are versioned alongside the code, so having them readable is good but I wouldn't stress too much on their UX, since e.g. when using Spago these will be autogenerated

thomashoneyman:
@f-f I made some tweaks resulting from the `Coercible` change, namely pulling the `RawPackageName` and `RawVersion` out into their own module and then re-exporting them from the `Registry.Prelude`, if you'd like to review again.

f-f:
Looks great!

ci/src/Registry/Schema.purs
```diff
@@ -1,14 +1,14 @@
 module Registry.Schema where
 
-import Registry.Prelude
+import Registry.Prelude (class Eq, class Newtype, class Show, Maybe, Object, bind, discard, fromMaybe, genericShow, isJust, mempty, pure, show, ($), (<$>), (<<<), (<>), (<|>), (||))
```
JordanMartinez: Was switching from an open import to this explicit import intentional or accidental?

thomashoneyman: Oops! Not intentional!

ci/src/Registry/Scripts/LegacyImport/Error.purs
```diff
@@ -144,35 +194,33 @@ printManifestErrorKey = ManifestErrorKey <<< case _ of
...
 
 derive instance Eq APIResource
-derive instance Generic APIResource _
-
-instance Json.EncodeJson APIResource where
-  encodeJson = Json.Encode.Generic.genericEncodeJsonWith encodingOptions
 
-instance Json.DecodeJson APIResource where
-  decodeJson = Json.Decode.Generic.genericDecodeJsonWith encodingOptions
+instance RegistryJson APIResource where
```
JordanMartinez, line 217: Shouldn't this be encoded similar to the 'tag' approach you used above? I assume that `APIResource` is a sum type that only _currently_ has one data constructor. But, if we add more in the future, it'll break this codec, correct?

JordanMartinez:
LGTM otherwise!

ci/src/Registry/Scripts/LegacyImport/Error.purs
thomashoneyman: I'm encoding nullary constructors as strings, and constructors with a single argument as `{ tag, value }`, so this is in line with the other codecs. If we add more constructors in the future we'll need to update this codec but the output will be the same (`GitHubReleases` will still encode as `"GitHubReleases"`)

JordanMartinez: Ah, I didn't realize that this was just an enum and not something where data constructors could be given arguments.

PR #295, opened 2022-01-16
Reintroduce Address type for package sets

### LINKED ISSUE #245: Reintroduce `Address` type for package sets
An earlier revision of the registry spec included an `Address` type that described where you could find the sources for a particular package:

https://github.com/purescript/registry/blob/14a8fb7c6c711ae2bcee54c4753010d57af11721/v1/Address.dhall

```dhall
let Address = 
  < Registry : { name : Text, version : Text }
  | Repo : Repo
  | Local : Prelude.Location.Type 
  >
```

It could be used to override a package in the package set (or add new ones to the package set), for example:

https://github.com/purescript/registry/tree/14a8fb7c6c711ae2bcee54c4753010d57af11721#overriding-a-package

However, this type was later removed, which means that a package set _always_ points to packages from the Registry. That would mean that you can create your own custom packages sets, but they would solely be comprised of packages from the registry and you just choose the packages and versions. If you wanted anything outside the registry (for example, a private package), then you're out of luck. With this model, it's up to the package managers to implement a way to use packages outside of the package set.

However, I think that it's considerably nicer to be able to create your own package set or modify the existing package set to produce the list of available dependencies for your project -- including packages outside the registry, like your own internal packages. This way the package set is always the list of available packages to install, and you don't have to learn a new way to install packages depending on whether they're in the registry or not.

To do that, I propose we:

1. Reintroduce the `Address.dhall` file / type, and use it to implement the official package set
2. Use `assert` or some other validation step to require the official package set to _only_ use packages with the `Registry` constructor (only packages in the registry are allowed in the official set)
3. Reintroduce documentation on how you can add new packages to your custom package set.

colinwahl:
Intends to close #245 

This PR reintroduces the `Address` type, which is used for overriding a package in a package set. It allows users to replace individual packages within a package set, or add their own, without that package being changed in the upstream package set.

So far, the following is done:
- [x] Reintroduce `Address.dhall`, with improvements to the `Repo` constructor
- [x] Reintroduce documentation for overriding a package in a package set
- [x] Migrate example package set to use `Address`
- [x] Add a `PackageSet.dhall` type specification
- [x] Add an `assert` function to ensure official package sets only rely on packages within the registry

Notably the `assert` check to ensure our official package set is only made up of `Registry` entries has not been implemented. I understand how the check works/what it is for - but I am not sure the best place to put it. 

@f-f - two questions for you. First, is having the `PackageSet.dhall` specification useful? I noticed I couldn't type the example package set as that type, presumably because the keys of the map are unknown, but my dhall-fu isn't quite there yet. Second - where do you think the `assert` should live? I was thinking it could go in `PackageSet.dhall` and be imported to the package sets when we construct them. I suggest this because I don't think it'd be ideal to have to that implement in every package set.


v1/sets/20200911.dhall
```diff
@@ -414,4 +412,8 @@ let packages =
   , zeta-extra = Address.Registry { name = "zeta-extra", version = "v0.0.1" }
   }
 
+let isPackageSetInRegistry = ../isPackageSetInRegistry.dhall
+
+let test = assert : isPackageSetInRegistry (toMap packages) === True
```
thomashoneyman, line 417: Should this test be included in every package set we publish?

f-f: Yes. The `assert` operator can't be packaged inside a function (because you can't do assertions about abstract values, only concrete ones), so this line goes in every set. Everything else is self-contained in that function we import

v1/isPackageSetInRegistry.dhall
```diff
@@ -0,0 +1,23 @@
...
+      , Local = \(x : Prelude.Location.Type) -> False
+      } pkg.mapValue
+
+-- here we map all the packages in the set with the function defined above,
+-- then we say "all the elements of this list are `True`"
+let isPackageSetInRegistry = \(pkgs : Pkgs) ->
+      Prelude.List.all Bool (\(x : Bool) -> x == True)
+        (Prelude.List.map Pkg Bool isRegistryPkg pkgs)
+
+in isPackageSetInRegistry
```
thomashoneyman:
```suggestion
in isPackageSetInRegistry

```

README.md
```diff
@@ -228,6 +228,39 @@ but it's allowed only for a set period of time because of the `leftpad` problem
...
+          , packages = toMap (upstream.packages // overrides)
+          }
+    , targets =
+        toMap
+          { src = Registry.Target::{
+            , sources = [ "src/**/*.purs" ]
+            , dependencies = toMap { effect = "^2.0.0" }
+            }
+          }
+    }
```
thomashoneyman: @f-f I wonder if there's an easier way to specify an override, as this involves a bit of specific knowledge (`toMap`, the type annotations) that I'd rather not force users to learn about. Is there a way to abstract any of this out?

f-f: A new Dhall operator has been proposed to deal with these cases (changing a record without using `toMap`): https://github.com/dhall-lang/dhall-lang/issues/979
The status there is that there's a proposal, there's pretty much agreement that it's good to have, there's availability to put a bounty on it, and it needs someone to implement it.

In any case, I would not add this documentation over here, since it doesn't really concern the registry?

f-f:
@colinwahl thanks for doing this!

A couple of notes:
- I just pushed a commit with the implementation of the `assert` check - we can abstract away most of it, but we'll have to copy one line in every package set we have
- as you realized it's not possible to give a proper type to the package set because it's a record. We might want to make sure that the set has the right type - in that case we could run `toMap` (as we do for the assertion) and then conform to the `PackageSet` type that you added (we'll just need a type annotation) if we don't need the `PackageSet`

README.md
colinwahl: Oh hah, I had meant this to be a comment on how to override a package within a `PackageSet`, but misread the old documentation when I moved it over. I'll change this to be an example on how to override a specific package within a package set

f-f:
Oh, I forgot to bring this up: IIRC the reason we have the form
```dhall
prelude = Address.Registry { name = "prelude", version = "v6.0.0" }
```
instead of 
```dhall
prelude = Address.Registry { version = "v6.0.0" }
```
..is so that the pointer to the package is "self-contained", i.e. we have all the information to get to the package in the record, instead of having to rely on the key.

Now, the question I have is: is this useful? Self-contained is nice, but the second form is much nicer to look at and to deal with? (less accidental typos, etc.) Should we switch to the second form? cc @thomashoneyman

README.md
f-f: Yeah, I think we only need documentation about how to author such package sets

thomashoneyman:
Since users also use the `Address` type in their code, I would greatly prefer the second form:

```dhall
prelude = Address.Registry { version = "v6.0.0" }
```

or even

```dhall
prelude = Address.Registry "v6.0.0"
```

thomashoneyman:
I'm struggling to think of a time when we would need the self-contained version and wouldn't be able to produce it from the not-self-contained version. In PureScript we can easily do this processing step, and I'd assume that even in Dhall it should be possible to turn a `Map Package Version` into a `Map Package { package, version }`

colinwahl:
Got it - I'll make the change to the `Address.Registry` constructor.

colinwahl:
Alrighty - I've updated the `Address.Registry` construct to not include the `name`, and I've corrected the documentation.

I looked a bit into adding the `PackageSet` type back in - but I decided it wasn't worth it at this point. We can loop back and change that later if we want!

v1/Address.dhall
```diff
@@ -0,0 +1,19 @@
...
+- some local path, for local packages. Note: the path must point to the spago.dhall
+- a GitHub/Git repo
+-}
+
+let Prelude = ./Prelude.dhall
+
+let Repo = ./Repo.dhall
+
+let Address =
+    < Registry : { version : Text }
```
f-f: Should we change this to
```suggestion
    < Registry : Text
```
as @thomashoneyman suggested?

colinwahl: As discussed in the Registry call - we will introduce a `SemVer = Text` type alias and use that directly in the `Registry` constructor.

thomashoneyman: Once that's done this looks good to me

v1/isPackageSetInRegistry.dhall
```diff
@@ -9,7 +9,7 @@ let Repo = ./Repo.dhall
 let Pkg = Prelude.Map.Entry Text Address
 let Pkgs = Prelude.Map.Type Text Address
 let isRegistryPkg = \(pkg : Pkg) -> merge
-      { Registry = \(x : { version : Text }) -> True
+      { Registry = \(x : Text) -> True
```
thomashoneyman, line 12: Should this be SemVer?

f-f: The SemVer synonym is not defined in this file, so we'd have to define it again or pull it out as a separate definition, but I don't think any of these are worth doing (as one can always have a look at where `Address` is defined)

f-f (approved):
Thanks @colinwahl!

PR #296, opened 2022-01-19
Use base64-encoded SRI sha256 hashes in metadata

### LINKED ISSUE #290: Rename `hash` in Metadata?
Right now we just include a `hash` field in the Metadata for every package:
https://github.com/purescript/registry/blob/85503f8fc24ac8358911cd5b4c4292a9803dd0e1/v1/Metadata.dhall#L14

I was wondering: should this be clearer, and/or more future-proof? What if we want to change the hashing function at some point?

So I'd like to discuss if we should do anything about this. Some ideas:
- we could rename the `hash` field to `sha256`, so that we can add new fields without trying to reuse this one
- or we could change the type to a record like `hash: { sha256: "57ba32c54958..." }`, so that we can have a "variant" in there if we need to add new hashing functions
- or we could add information about the hashing used in the string itself, e.g. "sha256-57ba32c54958.." (does this scheme have a name? I think so but I couldn't find it)

cc @thomashoneyman @cdepillabout

thomashoneyman:
Fixes #290 by introducing a new `Sha256` type along with functions for hashing a file and parsing a hash from a string (to ensure validity).

cc: @cdepillabout


ci/test/Registry/Hash.purs
```diff
@@ -0,0 +1,68 @@
...
+-- openssl dgst -sha256 -binary < test/fixtures/halogen-hooks-0.5.0.tar.gz | openssl base64 -A
+hooksTarballHash :: Hash.Sha256
+hooksTarballHash = Hash.unsafeSha256 "sha256-3nz2p8KZYRMbHFTuSk4kCVO47k0rZqn3qbtes/ebp9M="
+
+hooksTarball :: FilePath
+hooksTarball = Node.Path.concat [ "test", "fixtures", "halogen-hooks-0.5.0.tar.gz" ]
+
+sha256Nix :: FilePath -> ExceptT String Aff Hash.Sha256
+sha256Nix path = ExceptT do
+  let args = [ "hash-file", "--sri", path ]
```
cdepillabout, line 62: Might want to add a small comment here that this command will eventually change to be `hash file` instead of `hash-file`, as per https://github.com/purescript/registry/issues/290#issuecomment-1010635767.

cdepillabout (approved):
This LGTM.

cc @ddellacosta as well

ci/src/Registry/Hash.purs
```diff
@@ -0,0 +1,62 @@
...
+  ( Sha256
+  , sha256File
+  , parseSha256
+  , unsafeSha256
+  ) where
+
+import Registry.Prelude
+
+import Data.Array as Array
+import Data.List.Lazy as LL
```
f-f: Could we expand these import abbreviations? It's fine if there are only a couple, but so many of them make the code more confusing to read as it's not immediately clear where a function is from and why it's there

ci/test/Registry/Hash.purs
```diff
@@ -0,0 +1,68 @@
...
+-- Test hash produced by `openssl`:
+-- openssl dgst -sha256 -binary < test/fixtures/halogen-hooks/spago.dhall | openssl base64 -A
+hooksSpagoHash :: Hash.Sha256
+hooksSpagoHash = Hash.unsafeSha256 "sha256-fN9RUAzN21ZY4Y0UwqUSxwUPVz1g7/pcqoDvbJZoT04="
+
+hooksSpago :: FilePath
+hooksSpago = Node.Path.concat [ "test", "fixtures", "halogen-hooks", "spago.dhall" ]
+
+-- Test hash produced by `openssl`:
+-- openssl dgst -sha256 -binary < test/fixtures/halogen-hooks-0.5.0.tar.gz | openssl base64 -A
```
f-f: Can we hash something that we already have in the repo (e.g. some fixture or the `LICENSE` file) instead of pulling in a binary blob just for this test?

ci/src/Registry/Hash.purs
thomashoneyman: Done!

ci/test/Registry/Hash.purs
thomashoneyman: I've added a short comment, thanks!

thomashoneyman:
@f-f, @colinwahl, this is ready for another look

f-f (approved):
Looks great! :100:

PR #297, opened 2022-01-19
Restrict maximum package size to 200kb (for now)

### LINKED ISSUE #287: Package size limit
Should we cap the size of tarballs to some sane limit?

We do have a cap on our storage on DO, so it might make sense to limit package size so we have some leeway.

thomashoneyman:
Fixes #287 by adding a `bytes : Natural` field to `Metadata.dhall` and guarding the API pipeline to fail if a package tarball exceeds a 200kb limit.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -23,6 +23,7 @@ import Node.Buffer as Buffer
 import Node.ChildProcess as NodeProcess
 import Node.Crypto.Hash as Hash
 import Node.FS.Aff as FS
+import Node.FS.Stats as FS.Stats
 import Node.Process as Env
 import Registry.Index as Index
 import Registry.Json as Json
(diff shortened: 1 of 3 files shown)


ci/src/Registry/API.purs
```diff
@@ -362,3 +367,6 @@ runGit args = ExceptT do
       info result.stderr
       pure $ Right unit
     _ -> pure $ Left result.stderr
+
+maxPackageBytes :: Number
+maxPackageBytes = 200_000.0
```
colinwahl, line 372: I never know whether or not to go with `200_000.0` or `204_800.0` for stuff like this :scream:

colinwahl:
I don't see a change to `Manifest.dhall` - did that get forgotten?

Also - do we prefer `bytes` or `length` as the field name? I think `bytes` is more clear, although #287 suggests `length`.

thomashoneyman:
> I don't see a change to `Manifest.dhall` - did that get forgotten?

Ugh, I always forget to commit these because I mainly work out of the CI directory and stage changes there. I must have the change sitting around still and I’ll push it in the morning.

> Also - do we prefer `bytes` or `length` as the field name? I think `bytes` is more clear, although #287 suggests `length`.

Yes, I should have explained this. I think we should use bytes as the field name so that it’s unambiguous what the units of measure are — but I’d like to hear what @f-f thinks too.

f-f:
> Also - do we prefer `bytes` or `length` as the field name?

I have no preference about this, `bytes` sounds good!

thomashoneyman:
@f-f, @colinwahl, this is ready for another look

ci/src/Registry/API.purs
thomashoneyman: As per the registry call, we're doing 200_000! 😅

v1/Metadata.dhall
```diff
@@ -11,7 +11,7 @@ let Repo = ./Repo.dhall
 
 let SemVer = Text
 
-let VersionMetadata = { ref : Text, hash : Text }
+let VersionMetadata = { ref : Text, hash : Text, bytes : Natural }
```
f-f: This type now contains enough data that it's worth documenting the various fields:
```suggestion
-- Information about a single published version
let VersionMetadata =
  {
  -- The git ref this version points to
  , ref : Text
  -- The hash of the source tarball fetched from the repo
  , hash : Text
  -- The size in bytes of the tarball
  , bytes : Natural
  }
```

f-f (approved):
Left only a small documentation comment, but otherwise this looks great! :clap:

Issue #302, opened 2022-01-19
Verify that we retain package names even if all versions are dropped


thomashoneyman:
The registry's initial contents are imported from the existing Bower registry. During that import we do not accept all package versions, for various reasons as described in #300. Some packages have issues with _all_ versions. We need to verify that packages in which all versions have failed are still included in the resulting registry index. This is necessary because it is a security issue to allow an existing package name to be re-registered.

PR #303, opened 2022-01-19
Revert "Remove unreachable repositories from the Bower packages"

thomashoneyman:
As part of #302 I mentioned the security issue inherent in allowing package names to be re-registered when they already existed in the Bower registry. In #191 we removed some unreachable repositories because they could not be used in the new registry, but during the registry call this morning (2021-01-19) we decided that these packages _should_ be listed in the new registry, but without any accompanying versions.

This commit reverts #191 so that these packages can fail and be included in the list of packages unavailable in the registry while preserving the names so they can't be re-registered.


f-f:
I thought in the call we agreed that these happened early enough that it wouldn't be a big deal, but reserving these names is the right thing to do in the end :slightly_smiling_face: 

I'm afraid just reverting this commit is not going to be enough, there are a bunch of commits [here](https://github.com/purescript/registry/commit/a224f0cac15b49b7ede424be566f1f96acc9add2) and [there](https://github.com/purescript/registry/commit/02154a4ff029ffa46a775985afdbdea0f59dc678) that remove packages that turn out to be unreachable :smile: 
A comprehensive approach could be to go to the commit where that file was added, and diff it with the one we have today. Since all the new packages went to another file, no new packages should have been added to the `bower-packages.json` file since the beginning.

f-f:
Ah, [the full history](https://github.com/purescript/registry/commits/master/bower-packages.json) of that file is very small, so going through all these commits should be very quick

thomashoneyman:
Right you are! I've opened a better PR, #308.

PR #305, opened 2022-01-20
Implement simple versions and ranges

### LINKED ISSUE #43: Suggestion: ban fancy semver ranges
I have felt for quite a long time that semver ranges like `^x.y.z` or `~x.y.z` are an antipattern, because they obscure what the actual bounds are, and they encourage you to tighten lower bounds prematurely. For instance, lots of JS tools push you towards using a caret range with whatever version you happened to be using when you were writing the code. Of course, it's a reasonable default at first if you're adding a dependency at version `x.y.z` to declare the bounds as `>=x.y.z <(x+1).0.0` (note: this is what `^x.y.z` means as long as `x >= 1`), but if you later make a change to support newer versions of that dependency, you often won't need to tighten the lower bound at the same time. That is, if I've written `^x.y.z` and I want to allow my package to be used with a new version `(x+1).0.0` of the same dependency, caret ranges encourage me to just write `^(x+1).0.0`, which means that my package can no longer be used with any `x.*` version. Instead, if I was forced to write `>=x.y.z <(x+1).0.0` to begin with, it's easy to see how to relax the upper bound without tightening the lower bound.

To be explicit, my proposal is that the only acceptable version range operators are `>`, `>=`, `<`, `<=`, and `==`.

Note that the caret or tilde ranges aren't part of the semver spec; as far as I'm aware, they were first introduced by the node semver package. Elm does this already (and in fact goes a little further), by requiring that all version ranges are of the form `a.b.c <= v < d.e.f`.

### LINKED ISSUE #301: Implement simple versions & simple version ranges
As per #43, we are going to restrict versions and version ranges in the registry. We can use existing tools like the `node-semver` package's `parseRange` function to produce simple ranges from commonly-used operators like `^` and `~`, but this isn't sufficient to cover the restrictions we'd like to have in place.

Therefore, as part of adopting simple versions and version ranges we should also implement parsing and validation functions that enforce this restriction. This issue tracks implementing those checks and including tests to cover them.

thomashoneyman:
Fixes #43 by implementing new `Version` and `Range` types with parsers according to the spec described there.

In 5b3da446c4120952de0493903f37117371503f08 I added a new `Version.purs` file that implements a `Version` type along with  a parser and printer, enforcing that versions have the form `X.Y.Z` where each place is a non-negative integer. This file also implements a `Range` type with a parser and printer, enforcing that ranges have the form `>=X.Y.Z <X.Y.Z`. I also added tests to ensure this is working properly.

In 4a48ff59528a8a734b65c56d0b58c6f975336433 I then replaced the existing `SemVer` and `Range` types with the new types throughout the application. This allows us to remove almost all of the `SemVer` module and the accompanying tests. We retain `parseRange` since we use that in the legacy import tool to convert ranges.

In f4295c297aa4a1e8c01569bef8ff8c4067d79779 I introduced 'strict' vs. 'lenient' parsers for the `Version` and `Range` types, as sometimes we are working with user input we expect to be valid (ie. when a user uploads a new package) and sometimes we expect the input to be invalid (ie. when converting from an existing Bowerfile). I've also added a `raw` field to both types as we need to preserve the input in the case of the lenient parser; for the strict parser the input is identical to the printing functions. One day we can probably remove the lenient mode, but we need it for the time being.

In d3517d1c99f258a85df3e422605abebdcdddcf88 I went further by doing more work to convert ranges that are technically incorrect but which _could_ be converted into correct ranges. This saves us a lot of versions that would otherwise be kicked out of the registry due to an incorrect format. For example, the range `1.0.0` is invalid but can be rewritten to `>=1.0.0 <1.0.1`.

This change affects several hundred package versions, but the vast majority of those are not due to removing the prerelease identifiers. Instead, it's because we previously accepted many ranges that we shouldn't have, like `*` or `>=1.0.0` or `>1.0.0 <2.0.0 || >5.0.0 <6.0.0`.

However, **237** package versions will be dropped due to using a prerelease identifier in their version, and **1** package version will be dropped due to using build metadata (out of ~13,000 total). The vast majority of these predate the use of package sets and won't affect typical user -- for example, many of them are in the PureScript libraries that were part of the 0.9 release back in 2016.

For this reason I feel comfortable with these versions being dropped and us sticking with the new strict simple ranges.

---

A topic for discussion: What if we just stripped all prerelease identifiers while cleaning version ranges in the legacy import tool? For example, a version range `>=1.0.0-beta <2.0.0` would just be rewritten to `>=1.0.0 <2.0.0`. We'd still remove package versions that include a prerelease identifier, but we could fix packages that depend on those packages.


thomashoneyman:
For reference, here are the stats based on master prior to the `Json` class merging:

```
Packages: 1598 total (1230 totally succeeded, 248 partially succeeded, 120 totally failed)
Versions: 12602 total (11302 totally succeeded, 0 partially succeeded, 1300 totally failed)

Failures by error:
  manifestError: 1231 versions across 308 packages
    missingLicense: 501 versions across 147 packages
    badLicense: 415 versions across 55 packages
    badDependencyVersions: 330 versions across 118 packages
  noDependencyFiles: 68 versions across 36 packages
  resourceError: 0 versions across 31 packages

271 manifest entries with unsatisfied dependencies
```

thomashoneyman:
And here's the results from this pull request:

```
Packages: 1598 total (877 totally succeeded, 409 partially succeeded, 312 totally failed)
Versions: 12598 total (9142 totally succeeded, 0 partially succeeded, 3456 totally failed)

Failures by error:
  manifestError: 3390 versions across 668 packages
    badDependencyVersions: 2323 versions across 467 packages
    missingLicense: 495 versions across 145 packages
    badLicense: 410 versions across 55 packages
    badVersion: 323 versions across 143 packages
  noDependencyFiles: 65 versions across 34 packages

199 manifest entries with unsatisfied dependencies
```

We're looking at about 2,000 versions affected either because their package version is not good, or because they depend on package versions that are no good. This is more significant than I was expecting; I'll need to dig in a little more to see exactly why so many versions are affected.

thomashoneyman:
In the latest commit I've taken some actions to fix ranges that _could_ be valid according to our simple range spec, but which currently aren't converted. Some examples:

* When only an upper bound is specified we can set the lower bound to 0.0.0
  "<1.0.0" -> ">=0.0.0 <1.0.0", "<=0.1.1" -> ">=0.0.0 <0.1.0"

* When ranges unexpectedly end with `<=` we can increment the patch version:
  ">=1.0.0 <=2.0.0" -> ">=1.0.0 <2.0.1"

* When ranges unexpectedly begin with `>` we can increment the patch version:
  ">1.1.0 <2.0.0" -> ">=1.1.1 <2.0.0"

* When ranges are unbounded towards 0, we can set the lower limit to 0.0.0:
  "<1.0.0" -> ">=0.0.0 <1.0.0"

With this change, the new stats are:

```
Packages: 1598 total (1110 totally succeeded, 344 partially succeeded, 144 totally failed)
Versions: 12598 total (10627 totally succeeded, 0 partially succeeded, 1971 totally failed)

Failures by error:
  manifestError: 1969 versions across 455 packages
    badDependencyVersions: 823 versions across 218 packages
    missingLicense: 497 versions across 146 packages
    badLicense: 424 versions across 65 packages
    badVersion: 323 versions across 143 packages
  noDependencyFiles: 1 versions across 1 packages

254 manifest entries with unsatisfied dependencies
```

Many of the new failed versions have nothing to do with prerelease identifiers but rather to do with ranges we should never have accepted, like `*` or `>=1.0.0` with no upper bound. That said, 237 packages have been dropped due to prerelease identifiers and 1 package has been dropped due to using build metadata.

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -186,22 +186,20 @@ withCache { encode, decode } path maybeDuration action = do
       _ -> false
 
     onCacheMiss = do
-      log $ "No cache hit for " <> show path
```
JordanMartinez, line 189: Was this intentional?

ci/src/Registry/Version.purs
```diff
@@ -171,34 +173,128 @@ parseRange mode input = do
...
+-- ">1.1.0 <2.0.0" -> ">=1.1.1 <2.0.0"
+convertRange :: String -> String
+convertRange input = fromRight input do
+  -- We do not accept build metadata; the Node SemVer package will strip it out
+  -- but accept the resulting version, which we don't want to do. Instead we
+  -- fail early if a package has build metadata
+  if String.contains (String.Pattern "+") input then
+    Left "Contains build metadata"
+  else
+    pure unit
```
JordanMartinez: Isn't this just?
```purescript
when (String.contains (String.Pattern "+") input) $ throwError "Contains build metadata"
```

ci/src/Registry/Version.purs
```diff
@@ -171,34 +173,128 @@ parseRange mode input = do
...
+
+  -- Replace exact ranges ('1.0.0') with ranges ('>=1.0.0 <1.0.1')
+  fixExactVersions str = fromRight str do
+    Version version <- parseVersion Lenient str
+    pure $ Array.fold
+      [ ">="
+      , printVersion (Version version)
+      , " <"
+      , printVersion (Version (version { patch = version.patch + 1 }))
+      ]
```
JordanMartinez: Shouldn't this be:
```purescript
version <- parseVersion Lenient str
pure $ Array.fold
  [ ">="
  , printVersion version
  , " <"
  , printVersion $ bumpPatch version
  ]
```

JordanMartinez (approved):
LGTM. Just a few places that can be cleaned up.

ci/src/Registry/Scripts/LegacyImport/Process.purs
thomashoneyman: Yes, I was finding it becoming quite noisy when running the script, but I can add it back in if desired.

ci/src/Registry/Version.purs
thomashoneyman: Good catch! This predated the bumpPatch function and I forgot to wrap back around to it.

ci/src/Registry/Scripts/LegacyImport/Process.purs
JordanMartinez: Nah, just wanted to confirm that this wasn't done by accident.

f-f: I've seen these pop up on repeated runs of the importer and that made me wonder if these should actually be cache misses at all. Removing the log also takes away visibility on the fact that this could be a bug?

thomashoneyman: I can revert it!

thomashoneyman:
In 0bfe903 I added support for trimming prerelease identifiers from dependency ranges when parsing in lenient mode. The new results with this change:

```
Packages: 1598 total (1124 totally succeeded, 331 partially succeeded, 143 totally failed)
Versions: 12601 total (10726 totally succeeded, 0 partially succeeded, 1875 totally failed)

Failures by error:
  manifestError: 1873 versions across 441 packages
    badDependencyVersions: 726 versions across 200 packages
    missingLicense: 497 versions across 146 packages
    badLicense: 424 versions across 65 packages
    badVersion: 323 versions across 143 packages
  noDependencyFiles: 1 versions across 1 packages

255 manifest entries with unsatisfied dependencies
```

There are 83 packages that have moved from 'totally succeeded' to 'partially succeeded' and 23 packages that are now failing entirely.

There are about 400 more package versions that are affected by the `BadDependencyVersions` error, which is generally because they used ranges we don't support (a large number use ranges like `*` or `>=1.0.0`). There are 262 more package versions affected by the `BadVersion` error, which generally means these packages used prerelease identifiers or build metadata or something `SemVer.parse` recognized, but which are not accepted anymore.

ci/src/Registry/Scripts/LegacyImport/Process.purs
thomashoneyman: For what it's worth, I'm not seeing any of these pop up anymore except for when the cache has expired.

ci/src/Registry/Version.purs
```diff
@@ -0,0 +1,307 @@
...
+          , printVersion lhs
+          , ") must be less than right-hand version ("
+          , printVersion rhs
+          , ")"
+          ]
+      else do
+        pure $ Range { lhs, rhs, mode, raw: input }
+
+  let
+    parserInput :: String
```
colinwahl, line 158: Can you move this above `parser` to indicate it should be read first?

ci/test/Registry/Scripts/LegacyImport/Stats.purs
```diff
@@ -83,14 +84,17 @@ exampleFailures = PackageFailures $ Map.fromFoldable
 exampleStats :: Stats.Stats
 exampleStats = Stats.errorStats mockStats
   where
+  unsafeFromRight :: forall e a. Either e a -> a
```
colinwahl, line 87: Can we put this in `Registry.Prelude`? I think we define this in quite a few places.

ci/test/Registry/Version.purs
```diff
@@ -0,0 +1,172 @@
...
+        parsed `Assert.shouldSatisfy` Either.isLeft
+
+  Spec.describe "Valid strict ranges parse correctly and equal the parsed string" do
+    for_ validStrictRanges \range ->
+      Spec.it range do
+        let parsed = Version.parseRange Strict range
+        map Version.printRange parsed `Assert.shouldContain` range
+        map Version.rawRange parsed `Assert.shouldContain` range
+
+  Spec.describe "Invalid lenient ranges pass node-semver conversion but fail registry parsing" do
```
colinwahl: Do we still need this check?

thomashoneyman: Yea, we still need to check that invalid lenient ranges fail registry parsing, but we don't need to check the `node-semver` conversion anymore. There used to be many more invalid lenient ranges and it was useful to see the intermediate representation that `node-semver` produced, but now it doesn't produce anything different from the input range. I'll go ahead and remove that part.

ci/test/Registry/Scripts/LegacyImport/Stats.purs
thomashoneyman: I'm down to do that, but I'd like to do it in a followup PR to keep this one from touching many more files.

colinwahl: Sounds good!

colinwahl (approved):

PR #306, opened 2022-01-22
Parametric Package Sets

f-f:
After reintroducing the `Address` type in #295 we got wondering about the fact that a section of that sum type (the `Local` constructor) was supposed to be in the hands of the clients, that would spec how to interpret its contents in order to fetch packages, but at the same time we were making the type of the content concrete, and not leaving space for clients to eventually introduce other data that they would need there.

In this PR we move from a concrete `Address` type to an `Address a` type, so that clients can specify the type of the extra-Registry packages that they accept:
```dhall
let Address = \(externalPkg : Type) ->
        < Registry : SemVer
        | ExternalPkg : externalPkg
        >
```

A few notes:
- the PR contains some documentation about how to use this feature if you'd like to make a custom packages set. While the package sets in this repo will be parametric, I expect that custom ones will be concrete because they'll be fixed to a specific concrete type of the external packages, and client (though of course there's nothing stopping different clients to use the same concrete type)
- while the provided example has a bunch of boilerplate to add a custom package, note that this is so because we are starting from scratch here. I expect that clients will provide Dhall functions to abstract this away, e.g. in Spago we could have `mkPkgFromGitHub`, etc
- I was afraid that making this parametric would not allow us to use the `assert` to check that we are only shipping `Registry` packages in the set.
  The good news is that the `assert` still works, but the even better news is that we don't need this `assert` at all, since there's no way to create something of a parametric type without having the constructor available!
  I left the check in, but if we decide to merge this then we can just take the check away


v1/Address.dhall
```diff
@@ -12,10 +12,9 @@ let Repo = ./Repo.dhall
...
 let SemVer = Text
 
-let Address =
-    < Registry : SemVer
-    | Repo : { repo : Repo, ref : Text }
-    | Local : Prelude.Location.Type
-    >
+let Address = \(externalPkg : Type) ->
+        < Registry : SemVer
+        | ExternalPkg : externalPkg
```
thomashoneyman: What about using `Registry | External` as the two cases? I think it's already clear from the context that we're talking about packages, whether the package is in the registry or not. If anything, it seems like if we were to use the word 'Package' it might be best on the overall type itself, ie. `let PackageLocation = ...`.

```suggestion
        | External : external
```

README.md
```diff
@@ -265,6 +246,49 @@ As noted in the beginning, Package Sets are a first class citizen of this design
...
+```
+
+..that is, an extra-Registry package in Spago could either point to a local path, or a remote repository.
+
+Here's an example of a package set that is exactly like the upstream, except for the `effect`
+package, that instead points to some repo from GitHub:
+
+```dhall
+-- We parametrize the upstream package set and the Address type by the package type that our client accepts:
+let upstream = https://raw.githubusercontent.com/purescript/registry/master/v1/sets/20200418.dhall SpagoPkg
```
thomashoneyman, line 275: Presumably this can be managed automatically by Spago when upgrading package sets?

README.md
```diff
@@ -265,6 +246,49 @@ As noted in the beginning, Package Sets are a first class citizen of this design
...
+let overrides =
+    { effect = Address.ExternalPkg (SpagoPkg.Repo
+        { ref = "v0.0.1"
+        , repo = Registry.Repo.GitHub
+            { subdir = None Text
+            , githubOwner = "someauthor"
+            , githubRepo = "somerepo"
+            }
+        })
+    }
```
thomashoneyman: What would this look like using the helper functions you're envisioning in Spago?

README.md
f-f: Could you elaborate more what you mean with this?

README.md
f-f: I have an example of this in the upcoming config-related issue for Spago, will crosslink it here

v1/Address.dhall
f-f: We can rename the constructor to `External`, but I'd like to keep the type parameter more explicit (I mean, it's already a generic `Type`, so we should say what it is about at least in the name)

README.md
f-f: The issue is https://github.com/purescript/spago/issues/842

The function:
```dhall
let mkPackageFromGitHub = \(githubOwner : Text) -> \(githubRepo : Text) -> \(ref : Text) ->
      Address.ExternalPkg (SpagoPackage.Repo
        { ref
        , repo = Registry.Repo.GitHub
            { subdir = None Text
            , githubOwner
            , githubRepo
            }
        })
```

And how to use it:
```dhall
let overrides =
      { prelude = Spago.mkPackageFromGitHub "some-author" "some-prelude-fork" "0.0.1"
      }
```

v1/Address.dhall
thomashoneyman: That's fine with me. I'd like if we can name the constructor `External` and leave the variable `externalPkg` or `externalPackage`.

f-f: Done!

thomashoneyman (approved):

PR #307, opened 2022-01-24
Replace jsonic with jsonrepair

thomashoneyman:
In #305 I noticed that our JSON parsing was mysteriously failing, due to changes in #294. At first I thought the implementation of the `Json` class was wrong, but the real culprit turned out to be `jsonic`, introduced in #221.

When we introduced the `Json` class we also started using `parseJsonic` instead of `parseJson` everywhere. This lead to some strange decoding errors. For example, trying to decode `"null"` with `jsonic` will return an error that it expected a `:`. 

When I reviewed `jsonic` again, I noticed that it is _extremely_ liberal in what it accepts. Objects with non-string keys are fine, as are key value pairs of the form `x: y`, and more. However, we only use it because some Bowerfiles contain trailing commas and we want to accept them rather than fail immediately with `Json.parse`.

This PR removes `jsonic` in favor of a simpler tool that pre-processes JSON to conform it to a strict JSON format, which we _only_ use in the legacy import tool (we require JSON input in the main API to be well-formed according to the spec). With this change the legacy import tool is once again working well.


ci/src/Registry/Scripts/LegacyImport/Error.purs
```diff
@@ -110,7 +110,7 @@ instance RegistryJson RequestError where
     BadRequest ->
       Json.encode { tag: "BadRequest" }
     BadStatus error ->
-      Json.encode { tag: "BadStatus ", value: error }
+      Json.encode { tag: "BadStatus", value: error }
```
thomashoneyman, line 113: I also fixed this typo 😬

f-f (approved):
Looks great! :clap:

PR #308, opened 2022-01-25
Revert bower-packages removals

thomashoneyman:
Supersedes #303 and helps with #302. This reverts all commits in the `bower-packages.json` history that removed packages. Instead, we'll just fail for these packages but preserve their names in the registry.


thomashoneyman:
Two packages used to exist in the Bower registry and have since been re-registered in the new registry:

* `bcrypt` was previously [`cxfreeio/purescript-bcrypt`](https://github.com/cxfreeio/purescript-bcrypt.git) and is now [`VFabricio/purescript-bcrypt`](https://github.com/VFabricio/purescript-bcrypt.git)
* `task` was previously [`purescript-contrib/purescript-task`](https://github.com/purescript-contrib/purescript-task) and is now [`ursi/purescript-task'](https://github.com/ursi/purescript-task.git).

I've manually reviewed both of these packages and I think we should simply accept that we made a mistake here and we will accept the two newly-registered packages in favor of the old removed ones. I'm going to remove them from the `bower-packages.json` file.

verify-registry-unique.sh
```diff
@@ -5,9 +5,8 @@ set -euxo pipefail
 
 # This script checks that there are no duplicate entries in the two json files listing packages
 
-total=$(cat bower-packages.json new-packages.json | jq -s "map(length) | add")
-unique=$(cat bower-packages.json new-packages.json | jq -s "map(keys) | add | unique | length"
-)
+bower=$(cat bower-packages.json new-packages.json | jq -s "add | length")
+unique=$(cat bower-packages.json new-packages.json | jq -s "unique | add | length")
```
thomashoneyman: I found this a little easier to verify that these two commands are measuring the same things (the key/value pairs), but if we should only be verifying keys then we could use `map(keys)` for both. What do you think of this change? I can revert it if you want @f-f

JordanMartinez (approved):
LGTM, but I'd say don't merge until getting your question about the `verify-registry-unique.sh` file answered.

verify-registry-unique.sh
f-f: My jq-fu is not strong enough for me to have an opinion on this, so any course of action is good by me, as long as the check still works correctly! (I feel like at some point I managed to find the correct jq incantations and never touched it again. Also keep in mind that we are going to deprecate this thing soon, so probably not worth worrying much about it)

f-f (approved):
:rocket:

PR #309, opened 2022-01-26
Only include target directories and specific files in package tarballs

### LINKED ISSUE #292: Only include files from 'targets' key in package tarballs

* `purs.json` file
* LICENSE file(s)

thomashoneyman:
Fixes #292 by including only:

* A set of fixed files (`purs.json`, `spago.dhall`, `packages.dhall`, `package.json`, `LICENSE`, `README.md`)
* Globs included in the `targets` key

I've also included tests that demonstrate the right files are indeed being copied or removed. Since there is a new fixture directory for the `aff-5.1.2` package, I'd recommend filtering to only `.purs` files in your initial review to keep it manageable. The code changes are quite small -- that eye-popping diff is largely from the new fixtures.


ci/src/Foreign/Node/FS.js
```diff
@@ -1,5 +1,8 @@
-const fs = require("fs");
+const fs = require("fs-extra");
```
thomashoneyman, line 1: This allows us to use Promise-based and more sensible implementations for creating, copying, and removing files and directories.

ci/src/Registry/API.purs
```diff
@@ -202,25 +208,30 @@ addOrUpdate { ref, fromBower, packageName } metadata = do
...
-  liftAff $ FS.rename absoluteFolderPath (tmpDir <> "/" <> newDirname)
-  let tarballPath = tmpDir <> "/" <> newDirname <> ".tar.gz"
-  liftEffect $ Tar.create { cwd: tmpDir, folderName: newDirname, archiveName: tarballPath }
+  let packageDirname = PackageName.print packageName <> "-" <> SemVer.version manifestRecord.version
+  let packagePath = tmpDir <> "/" <> packageDirname
+  let tarballPath = packagePath <> ".tar.gz"
+  liftAff do
+    FS.mkdir packagePath
+    pickFiles { inputDirectory: absoluteFolderPath, outputDirectory: packagePath } manifest
+  liftEffect $ Tar.create { cwd: tmpDir, folderName: packageDirname, archiveName: tarballPath }
```
thomashoneyman, line 217: This is the main difference in this new file. Instead of just renaming the extracted source directory and creating a tarball from it, we create an empty directory and pick the files we want over into it.

ci/src/Registry/API.purs
```diff
@@ -360,3 +371,48 @@ runGit args = ExceptT do
...
+  traverse_ copy (exactFiles <> Set.toUnfoldable globMatches)
+  where
+  -- These files are always included at these exact paths
+  -- See: https://github.com/purescript/registry/issues/292
+  exactFiles :: Array FilePath
+  exactFiles = map (\path -> Path.concat [ inputDirectory, path ])
+    [ "purs.json"
+    , "spago.dhall"
+    , "packages.dhall"
+    , "package.json"
```
thomashoneyman, line 402: I've included these three files because they are small and they feel like they're useful to include if you were to say inspect your dependencies in Spago. It at least gives you what you would need to install and run the project with Spago and a JavaScript package manager. But I can see an argument for removing these.

ci/src/Registry/API.purs
```diff
@@ -360,3 +371,48 @@ runGit args = ExceptT do
...
+  globFiles =
+    [ "README*"
+    , "Readme*"
+    , "readme*"
+    , "LICENSE*"
+    , "License*"
+    , "license*"
+    , "LICENCE*"
+    , "Licence*"
+    , "licence*"
```
thomashoneyman, line 417: `node-glob-basic` doesn't support case-insensitive globbing so I've included common variations of each of these files.

ci/src/Registry/API.purs
```diff
@@ -360,3 +371,48 @@ runGit args = ExceptT do
...
+  :: { inputDirectory :: FilePath, outputDirectory :: FilePath }
+  -> Manifest
+  -> Aff Unit
+pickFiles { inputDirectory, outputDirectory } (Manifest { targets }) = do
+  let
+    copy path = whenM (FS.exists path) do
+      for_ (String.stripPrefix (String.Pattern inputDirectory) path) \path' ->
+        FS.Extra.copy { from: path, to: Path.concat [ outputDirectory, path' ] }
+
+  let sources = Array.nub $ Array.concatMap (un Target >>> _.sources) (Object.values targets)
```
thomashoneyman, line 391: This is _slightly_ dodgy in that our globbing will only consider the standard `*` and `**` operators and nothing else. I don't know what globbing Spago supports, but we should probably spec this and say that we only accept `*` and `**` in glob patterns.

ci/src/Registry/Index.purs
```diff
@@ -134,8 +134,5 @@ insertManifest directory manifest@(Manifest { name, version }) = do
         $ Array.sortBy (comparing (un Manifest >>> _.version))
         $ Array.fromFoldable modifiedManifests
 
-  dirExists <- FS.exists manifestDirectory
-  unless dirExists do
-    liftEffect $ Foreign.Node.mkdirSync manifestDirectory
-
+  FS.Extra.ensureDirectory manifestDirectory
```
thomashoneyman, line 137: This is equivalent behavior to checking if the directory exists and creating it if not.

ci/test-fixtures/aff-5.1.2/.gitignore
```diff
@@ -0,0 +1,9 @@
+/.*
```
thomashoneyman, line 1: All files in the `test-fixtures` directory can be safely ignored for the sake of review. It's just the existing fixtures plus the contents of `aff-5.1.2`. I have moved this out to a top-level directory so that PureScript code contained in fixtures isn't inadvertently included in our build via the `test/**/*.purs` glob pattern.

We can use `aff-5.1.2` any time we need to run tests against a legitimate Registry-compliant package.

thomashoneyman:
With some time to think on it, I'm not sure we can reliably guarantee that we're taking correct files from `targets` for a few reasons:

1. We use the target's `sources` array to determine what files are associated with the target. This is an array of globs. However, these are globs intended for the _compiler_, so they only specify PureScript files. But it would be incorrect to only take in PureScript files, as users may include FFI files, too (and not just .js FFI files, either). It may also be incorrect to only include FFI files (ie. for `MyModule.purs` look for any other files in the same directory named `MyModule.*`), because your FFI files could technically reference other JS intended for bundling.
2. The glob behavior for the compiler is determined by [the implementation of the `Glob` library for Haskell](https://hackage.haskell.org/package/Glob-0.9.2/docs/System-FilePath-Glob.html), which supports a _lot_ of globbing options. In the registry we've so far stuck to supporting only `*` and `**` patterns, but if we want to truly support the same things as the compiler we will have to rely on a Node library that will almost certainly have different behavior from the compiler globs.

In short, the `sources` key used in targets only specifies PureScript code, so using that as the source of truth for all files that need to be in the tarball doesn't quite make sense. We at least have to adjust the globs that we receive, and at that point we get into potentially risky territory where we inadvertently don't include the correct files in the tarball and break someone's package. We also may not be able to support the same globs that the compiler supports.

The only solutions that come to mind for me are:

1. Adjust the `sources` key so that it specifies directories only, and when feeding that to the compiler we turn it into a glob (ie. we transform `src` into `src/**/*.purs`). This has the downside that users can no longer adjust the glob to fit their use case, ie. they couldn't do something like `src/App/**/*.purs` anymore. This seems...not doable, really.
2. Adjust the globs retrieved via `sources` to strip the `.purs` extension, if present, and instead accept files with any extension. We'd have to hope that this works correctly in all cases. Since folks can provide some somewhat tricky globs this may not be possible to prove; we'll also need to move to a fully-featured globbing library via Node.
3. Don't attempt to take files based on the targets key and just accept the resulting larger package sizes. The larger package sizes don't translate into larger code for users, after all, just more weight to store and download. The downside is that packages will (possibly) contain lots of things not necessary for them to run.

With all of this in mind, I'm beginning to think we would be better off _not_ attempting to reduce the number of files in the package and we just package it as-is and deal with the larger package sizes. It's not clear to me that we can reliably trim files in user packages without risk of breaking the packages.

Any thoughts on this, @f-f and @colinwahl?

thomashoneyman:
As noted in [this comment](https://github.com/purescript/registry/issues/292#issuecomment-1022589803), just because we may not want to only include target directories doesn't mean we would avoid trimming packages at all; it still makes sense to trim out things like version control directories (`.git`), common folders that almost never should be accepted (`bower_components`, `node_modules`, `output`), and that sort of thing. But we just wouldn't attempt to only take the target sources anymore, we'd take everything.

f-f:
I don't think we should close this just yet. I understand the issues that come with trying to mess with the globs, but getting rid of the globs entirely doesn't seem like a bad idea to me?

Right now we don't have any legacy globs that we need to support (all the ones that we have are autogenerated by the legacy import tool), so if we decide to spec including only directories (as many other package ecosystems do) then I see nothing wrong with it?
It might just make life easier in the end, because as we found out here globs can be pretty annoying at times.

thomashoneyman:
We could spec the manifest to say that sources are always top-level directories, and that package managers can rely on the top level directory to produce globs for PureScript where `directory` corresponds with a `directory/**/*.purs` glob.

When generating from Bower we can do this, but when generating from `spago.dhall` files I'm not sure we can (since the sources key there is already a glob).

Is there a better place to track this discussion, since this largely affects the manifests and their relationship to the new Spago configuration?

f-f:
> We could spec the manifest to say that sources are always top-level directories, and that package managers can rely on the top level directory to produce globs for PureScript

Yes, this is what I'm proposing

> When generating from Bower we can do this, but when generating from spago.dhall files I'm not sure we can (since the sources key there is already a glob)

I'm going to speculate that all the globs that we read from Spago files in the legacy import can be replaced by directories. We could add a check to ensure that this is the case (while reading in and converting these globs) and error out if the glob is something else

> Is there a better place to track this discussion, since this largely affects the manifests and their relationship to the new Spago configuration?

There's really no other place to have this discussion: we're talking about the Manifest schema and this is the upstream since this is where we spec the schema. Clients will have to follow whatever is specced here.

If you mean "having this discussion in an issue rather than a PR" I have no opinion on where it should be done, as long as it's tracked somewhere

thomashoneyman:
I've added a new issue to track this discussion, because this PR is no longer how we would handle things.

PR #310, opened 2022-01-27
Add version and range spec to registry README

thomashoneyman:
This PR adds the specification of how we handle versions and ranges to the registry README, rather than only containing that information in the implementation itself.


f-f (approved):
Lovely!

PR #312, opened 2022-01-27
Add a documentation comment to SemVer.validRange

thomashoneyman:
We use specific options to control the behavior of the `node-semver` validRange function, and we mostly use it for its ability to convert ranges using operators like `^` and `~` into simpler ranges like `>=X.Y.Z <X.Y.Z`. This PR adds a documentation comment that more fully describes how we're using this function and what the options mean.

### DIFF
--- a/ci/src/Foreign/SemVer.js
+++ b/ci/src/Foreign/SemVer.js
@@ -7,9 +7,13 @@ exports.parseRangeImpl = (rangeString) =>
   // range is not valid, and the converted range if it is.
   //
   // For example:
-  // > validRange("^1.0.0") -> (">=1.0.0 <2.0.0")
+  //
+  // > validRange("^1.0.0")
+  // ">=1.0.0 <2.0.0"
   //
   // In loose mode `validRange` will also fix common errors like including a 'v' prefix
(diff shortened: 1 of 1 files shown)


ci/src/Foreign/SemVer.js
```diff
@@ -1,4 +1,15 @@
...
+  // using only comparative operators (no ^, ~, -, or other range operators), and then
+  // checks whether the result is a valid SemVer range. It then returns `null` if the
+  // range is not valid, and the converted range if it is.
+  //
+  // For example:
+  // > validRange("^1.0.0") -> (">=1.0.0 <2.0.0")
+  //
+  // In loose mode `validRange` will also fix common errors like including a 'v' prefix
+  // or extra spaces:
+  // > validRange(" = v 2.1.5-foo") -> "2.1.5-foo"
```
f-f:
```suggestion
  // > validRange(" = v 2.1.5-foo") -> "2.1.5-foo"
  //
  // We set these options because we should only be using this function to clean up ranges, not to validate them.
```

f-f (approved):

PR #314, opened 2022-01-29
Include publication date in metadata

### LINKED ISSUE #304: Add publishing date to Metadata?
Along the lines of #287 I think it would be great to add some information about when a version was published in the Metadata file.
Still not sure how this would work for legacy imports, but we can probably get information about when a tag was published to GitHub.

thomashoneyman:
Closes #304 by adding the publication date as an `RFC3339String` to the metadata. This is done by fetching the date associated with a particular commit SHA from the GitHub API.

In the main API we do this by first fetching the commit associated with a particular ref, and then fetching the date associated with that commit. There is no date information associated with a ref directly.

I tested this by fetching the dates associated with particular releases in the legacy importer. It works there, returning the correct `RFC3339String`s associated with each commit. However, it's not necessary to include this in the legacy importer because we don't produce metadata there; I haven't verified it is working in the API pipeline as I wasn't sure how to test that without actually executing the pipeline.

### DIFF
--- a/ci/src/Foreign/GitHub.purs
+++ b/ci/src/Foreign/GitHub.purs
@@ -46,7 +46,7 @@ type Address = { owner :: String, repo :: String }
 registryAddress :: Address
 registryAddress = { owner: "purescript", repo: "registry" }
 
-type Tag = { name :: String, sha :: String, date :: RFC3339String }
+type Tag = { name :: String, sha :: String }
 
 newtype PackageURL = PackageURL String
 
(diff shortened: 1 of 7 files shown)


ci/src/Foreign/GitHub.purs
```diff
@@ -66,11 +67,35 @@ parseRepo = unwrap >>> Parser.runParser do
...
+  date <- Promise.toAffE $ runEffectFn3 getCommitDateImpl octokit address sha
+  pure $ RFC3339String date
+
+getTags :: Octokit -> Address -> Aff (Array Tag)
+getTags octokit address = do
+  releases <- getReleases octokit address
+  -- While it would be nice to use `parTraverse` here and speed things up,
+  -- this does make it extremely easy to break GitHub's API rate limiting.
+  releases # traverse \release -> do
+    log $ "  " <> release.sha
```
thomashoneyman, line 95: This is a little noisy, but I like being able to see the commits being fetched. Ideally we'd be able to set various log levels in the application, but so far we don't do that.

ci/src/Registry/Schema.purs
```diff
@@ -163,6 +164,7 @@ type Metadata =
 type VersionMetadata =
   { ref :: String
```
thomashoneyman, line 165: We also have the commit sha that is associated with the ref, if there's any reason to include that in the metadata also. We're forced to get the commit in order to retrieve the published date, so we've got the info regardless.

f-f: I don't think we should include the commit sha, the tag ref should be enough (and we don't use it for anything, just bookkeeping)

f-f (approved):
Looks great!

PR #315, opened 2022-01-29
Always remove some files from packages prior to compression

### LINKED ISSUE #292: Only include files from 'targets' key in package tarballs

thomashoneyman:
Fixes #292 (at least partially) by always removing some files from packages before creating a tarball for the package. This is intended as a replacement for #309.

In #309 we tried to only include some required files (like a `purs.json` file), and otherwise only include files explicitly listed in a manifest's targets. However, there is a big issue with using the targets to decide what files to include.

A target defines sources, where sources are intended for the PureScript compiler to know what files to compile. These sources are included as globs, ie. `src/**/*.purs`. Sources are thus not suitable for knowing all files that need to be included in a package, because they are purely considering PureScript sources and not other possibly required files, like FFI files for various backends. Even if we do a best-effort fixup to the globs to take in all files regardless of their extension, we still run some risk of not including files that are meant to be included, and therefore breaking a package.

There is another possible solution: sources could list directories, not globs, and package managers etc. that rely on the manifest can presume that a directory like `src` means you can do `src/**/*.purs` to pass things to the PureScript compiler.

That's a much bigger conversation, though, so I wanted to put this up as an alternative to #309, as that PR is not usable unless we change the way sources are handled.

### DIFF
--- a/ci/package-lock.json
+++ b/ci/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@octokit/rest": "^18.12.0",
         "aws-sdk": "^2.1050.0",
+        "fs-extra": "^10.0.0",
         "fuse.js": "^6.5.3",
         "jsonrepair": "^2.2.1",
         "semver": "^7.3.5",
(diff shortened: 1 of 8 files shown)


f-f (approved):
Looks great!

PR #317, opened 2022-02-01
Remove targets and sources and only take files from the src directory

### LINKED ISSUE #164: Consider removing targets from the manifest
The current manifest specification includes a `targets` key:
https://github.com/purescript/registry/blob/55bce52392cab4b595ac1f542954cfceeef2d431/v1/Manifest.dhall#L22-L23

where a `Target` is defined as:

https://github.com/purescript/registry/blob/55bce52392cab4b595ac1f542954cfceeef2d431/v1/Target.dhall#L14-L20

...but it's not entirely clear to me why we need to include targets at all in a package manifest, as the registry (implicitly, at least) relies on a single target called `lib`:

https://github.com/purescript/registry/blob/3d091b10d63350e8d374612e0eabbf825394cb40/README.md#the-package-manifest

I understand the value of targets when _developing_ a package -- it's important to be able to specify things to compile (and their extra dependencies) besides `src`, like `tests`, `bench`, `examples`, and so on. For that reason it's vital that a build tool (like Spago) can support additional targets.

But I don't see why the extra targets information needs to be included in the manifest file unless the registry or package managers depending on the registry can do something with that information. If the `lib` target is all the registry cares about, then why not flatten that target into the manifest instead of checking for a `lib` target?

### LINKED ISSUE #292: Only include files from 'targets' key in package tarballs

### LINKED ISSUE #316: Require sources to be directory names in manifest targets
The `Manifest` targets key includes a list of dependencies for the target and a `sources` key that identifies what files should be included for compilation wrt that target:

https://github.com/purescript/registry/blob/335c07cb00f3dada632019996d9f66eed9c6079e/v1/Target.dhall#L18-L19

However, as [described in this comment](https://github.com/purescript/registry/pull/309#issuecomment-1022561203), there are a number of problems with just allowing globs of any kind to be included in the sources key. And so, as [proposed in this comment](https://github.com/purescript/registry/pull/309#issuecomment-1024797588) and [endorsed in this comment](https://github.com/purescript/registry/pull/309#issuecomment-1024923289), we should update the spec to assert that the `sources` key does _not_ contain a list of globs, but rather contains a list of directories where sources can be found.

In short, we'd specify that:

1. The `sources` key contains a list of directories (for example: `[ "src", "test" ]`).
2. Package managers are to understand that these directories imply that PureScript files can be found with the glob pattern `"directory/**/*.purs"`, such as `"src/**/*.purs"`, and this transformation should be made when calling the PureScript compiler.

As a followup:

1. Should we require directories to be top-level only, like `"src"`, or should we allow directory paths like `"src/lib"`?
2. We need to check how many legacy packages would be affected by this restriction, if any, by checking that sources only contain strings of the form `DIR/**/*.purs` and can be stripped to `DIR`.

thomashoneyman:
Fixes #316, fixes #292, and fixes #164. 

This PR removes the `Target` type altogether as described in #164. It also remove the `source` key altogether. It ensures in the API pipeline that when we create a tarball for a package we only include the `src` directory and some always-included files like the LICENSE or README or `purs.json` file.


ci/src/Registry/API.purs
```diff
@@ -260,14 +260,10 @@ runChecks metadata (Manifest manifest) = do
...
 
-  log "Checking that the Manifest includes the `lib` target"
-  Target libTarget <- case Object.lookup "lib" manifest.targets of
-    Nothing -> throwWithComment "Didn't find `lib` target in the Manifest!"
-    Just a -> pure a
-
-  log "Checking that `lib` target only includes `src`"
-  when (libTarget.sources /= [ "src/**/*.purs" ]) do
-    throwWithComment "The `lib` target only allows the following `sources`: `src/**/*.purs`"
+  -- TODO: Why do we only allow `src` as a manifest source?
```
thomashoneyman: @f-f If we only allow `src` as a source for the `lib` target, which is now the only target, then...should we just remove the `sources` key altogether? Or should we relax this restriction?

ci/src/Registry/Schema.purs
```diff
@@ -32,28 +54,47 @@ instance RegistryJson Manifest where
...
+    "repository" := fields.repository
+    "sources" := fields.sources
+    "dependencies" := mapKeys PackageName.print fields.dependencies
 
-  decode json = Manifest <$> Json.decode json
+  decode json = do
+    manifestFields <- Json.decode json
+    let parse = lmap StringParser.printParserError <<< PackageName.parse
+    parsed <- traverseKeys parse manifestFields.dependencies
+    pure $ Manifest $ manifestFields { dependencies = parsed }
```
thomashoneyman, line 45: This weird little conversion is required because we can't just arbitrarily coerce `PackageName`s, so we need to adjust our map keys to something that is coercible to a string.

ci/src/Registry/Scripts/LegacyImport/Bowerfile.purs
```diff
@@ -8,13 +8,28 @@ import Registry.Prelude
...
+      Map.toUnfoldable
+        >>> Array.mapMaybe
+          ( \(Tuple (RawPackageName packageName) version) ->
+              case String.take 11 packageName of
+                "purescript-" -> Just $ Tuple (RawPackageName (String.drop 11 packageName)) version
+                _ -> Nothing
+          )
+        >>> Map.fromFoldable
+
+  fields { dependencies = processNames fields.dependencies }
```
thomashoneyman: I moved the `filterName`s step out of the processing pipeline because it was being applied both to Spago dependencies and Bower dependencies. However, that means stripping out all the Spago dependencies because none of them begin with `purescript-`!

We always preferred the Bower deps over the Spago ones so we weren't hitting this bug, but it could bite us if we ever change that logic. This is really the right place for it.

ci/src/Registry/API.purs
thomashoneyman: Note that if we remove the `sources` key then we can also remove all the code related to the `Source` type, and we can always just take the contents of the `src` directory.

thomashoneyman: I found these historical discussions about why we want to allow sources other than `src` --  https://github.com/purescript/spago/issues/288#issuecomment-505674239, https://github.com/purescript/registry/issues/16

However, I am finding myself agreeing with @hdgarrood -- I don't see what we gain by allowing arbitrary source directories. Quoting:

> I still don't understand what we gain by making source paths configurable by a package author, over having the rule be "the visible source files are always those which are located in the src folder relative to the package manifest file, and only those". That still permits managing a number of packages together in a monorepo, and as far as I'm aware that is the only concern which has been brought up in the context of this issue.

I'm struggling to see the advantages of allowing source directories other than `src`. Our support for monorepo setups means you can publish multiple packages from the same repository already, storing packages in arbitrary subdirectories as you see fit. And if you want to organize your source code into multiple directories you can just do that inside the `src` directory. What additional benefit does supporting multiple source directories bring?

But I do see disadvantages! We diverge from the `src` directory for sources used in all current purescript projects, and tooling can no longer assume that source code is in the `src` directory and must read from the manifest instead, and we have to add extra code to parse this field, ensure it's correct, and bring in these source directories in the registry.

With that all in mind, I'm in favor of dropping this field altogether and requiring that package sources are in the `src` directory.

f-f: Yeah, reading these discussions from scratch a couple of years later gives some perspective on what folks are looking for - it looks like the main concern there was _consuming_ things from an arbitrary place. I believe we now achieve this with the `subdir` field.
That means that we are indeed able to say "a package is something that contains a `.purs.json` file and a directory `src` with the source files", and get rid of the `sources` field in the manifest.

ci/src/Registry/API.purs
```diff
@@ -367,6 +362,31 @@ runGit args = ExceptT do
...
+  , "spago.dhall"
+  , "README*"
+  , "Readme*"
+  , "readme*"
+  , "LICENSE*"
+  , "License*"
+  , "license*"
+  , "LICENCE*"
+  , "Licence*"
+  , "licence*"
```
f-f: I understand we have all of these because the purescript wrapper we use doesn't support case-insensitive globbing.
I feel that it would be easier to have our own tiny wrapper around [glob](https://www.npmjs.com/package/glob) instead of working around the existing wrapper?

ci/src/Registry/Scripts/LegacyImport/Bowerfile.purs
f-f: Great catch!

ci/src/Registry/Scripts/LegacyImport/Manifest.purs
```diff
@@ -230,34 +228,26 @@ toManifest package repository version manifest = do
...
-            _ -> Nothing
-
-        parsePairs = map \(Tuple packageName versionRange) -> case PackageName.parse packageName of
+        parsePairs = map \(Tuple (RawPackageName packageName) versionRange) -> case PackageName.parse packageName of
           Left _ -> Left packageName
-          Right name -> Right (Tuple name versionRange)
+          Right name -> Right (Tuple name (coerce versionRange))
 
         normalizeDeps deps = do
-          let { fail, success } = partitionEithers $ parsePairs $ filterNames deps
```
f-f, line 247: Where are we calling `filterNames` if not here?

ci/src/Registry/API.purs
thomashoneyman: I've replaced `Glob.Basic` with a wrapper around [`fast-glob`](https://github.com/mrmlnc/fast-glob).

ci/src/Registry/API.purs
```diff
@@ -367,44 +361,12 @@ runGit args = ExceptT do
...
-ignoredGlobs =
-  [ "*.*.swp"
-  , "._*"
-  ]
+pickTarballFiles :: { from :: FilePath, to :: FilePath } -> Aff Unit
+pickTarballFiles { from, to } = do
+  let options = { cwd: Just from, include: FilesOnly, caseSensitive: false }
+  acceptedMatches <- FastGlob.match' acceptedGlobs options
+  for_ ([ "src" ] <> acceptedMatches) \match ->
+    FS.Extra.copy { from: Path.concat [ from, match ], to: Path.concat [ to, match ] }
```
thomashoneyman: I've dropped the step where we delete ignored files, because:

1. We only take a few files from the top level and the contents of the `src` directory, so these ignored files would all be dropped _unless_ they were in the `src` directory
2. The ignored  files are the kind of files that are extremely unlikely to be present in a `src` directory; they're almost certainly top-level directories.
3. The less deletion code we have related to user files the better!

thomashoneyman:
This is ready for another review @f-f. The PR now:

1. Removes the `Target` type and the `targets` key in favor of just a `dependencies` key. All sources must be stored in a `src` directory, so there is no longer a `sources` key associated with a target.
2. Removes the logic around deleting specific files, instead _only_ including a few specific files and the contents of the `src` directory.
3. Adds `fast-glob` bindings to replace the `node-glob-basic` library, so that we can do more flexible globbing (like only include directories, or do a case-insensitive match). I chose `fast-glob` instead of `glob` because of its support for promises.

ci/src/Registry/API.purs
```diff
@@ -367,44 +361,12 @@ runGit args = ExceptT do
...
+
+acceptedGlobs :: Array String
+acceptedGlobs = [ "purs.json", "README*", "LICENSE*" ]
```
f-f: I would add here also:
- `bower.json`
- `spago.dhall`
- `packages.dhall`
- `packages.json`

This is so that downstream package managers can do things with these files - e.g. "collect all the JS dependencies of the dependencies of a project" could be one.

Are there any more common files we should add here to allow this kind of flexibility?

f-f (commented):
This looks great, I think we're almost there - I left a small comment about the files that we should include, and I now also wonder if we should add a `files` key to the Manifest file (as `npm` does) so that package authors can specify there which files to include in the packaging?

With the patch as it stands it's impossible to include arbitrary files in the root of the repo, while anything goes in the `src` directory - this is inflexible, so I think we should offer some kind of escape hatch now that we don't have the one from `targets.sources` anymore

ci/src/Registry/API.purs
```suggestion
acceptedGlobs = 
  [ "purs.json"
  , "README*"
  , "LICENSE*"  
  , "bower.json"
  , "spago.dhall"
  , "packages.dhall"
  , "packages.json"
  ]
```

thomashoneyman:
> With the patch as it stands it's impossible to include arbitrary files in the root of the repo, while anything goes in the src directory - this is inflexible, so I think we should offer some kind of escape hatch now that we don't have the one from targets.sources anymore

I think that we should choose one of:

1. Take all files from the package, minus some files we _know_ we don't want, like a `.git` directory
2. Take only a set of files we _know_ we want, like the manifest and `src` directory, and allow the user to specify extra files they want via an optional `files` key.

In other words, if we want to be permissive about files you can include in your package, then we should strip out only files we know for sure shouldn't be included. If we want to be restrictive, then we should only take the `purs.json` file and `src` directory and perhaps some other files as you mentioned, but we should leave an escape hatch open in case we are being _too_ restrictive.

I'm not completely sure which one I prefer. My risk averse self feels like the safe option is to be permissive and only take out files we know for sure we can remove. But I see how removing files keeps package sizes down.

f-f:
I'm for option (2): we can be restrictive if users have an escape hatch. This is also how `npm` does it, and it seems to work nicely.

Though I still think we should always pick up - even if we don't think we _need_ them - at least the various files that we use in the legacy import (`bower.json`, `spago.dhall`, etc) and the ones specifying native dependencies (`package.json` is one, there might be more)

f-f:
> I'm for option (2): we can be restrictive if users have an escape hatch

And I think we can cut this corner for now since it's not strictly necessary to move forward: we could add the `files` key to the schema, and then just throw an exception in the API pipeline if anyone tries to put anything in there. This would allow us to have a look at the usecase at hand, and eventually backtrack on this decision.
(i.e. we can add things to the schema and then take them away later as long as no published Manifest contains them :slightly_smiling_face: )

thomashoneyman:
> I'm for option (2): we can be restrictive if users have an escape hatch. This is also how `npm` does it, and it seems to work nicely.

That's not quite how `npm` does it:
https://docs.npmjs.com/cli/v8/commands/npm-publish#files-included-in-package

Rather, they take all files _except_ some guaranteed exclusions by default, but if you include a `files` key then they take _only_ some guaranteed inclusions plus the files you provide in the `files` key. I would be happy to follow suit.

> And I think we can cut this corner for now since it's not strictly necessary to move forward: we could add the files key to the schema, and then just throw an exception in the API pipeline if anyone tries to put anything in there. This would allow us to have a look at the usecase at hand, and eventually backtrack on this decision.

I'm OK with this, especially if we add the files key to the schema with a link to a discussion like this one. As we've found out recently, if something's there and I can't remember or figure out why, I'm gonna try and remove it!

I'm especially OK with this because it feels safer to take all files for legacy imports, rather than be restrictive, just in case _someone_ was doing something funky in their setup.

f-f:
As we discussed in the call, the way I'd like this to work would be:
- for legacy packages: take all the things except some obvious things we exclude
- for new packages: include specific things, plus a list of things specified in the future `files` field

thomashoneyman:
> plus a list of things specified in the future files field

And, to be clear, this will be a list of glob patterns, where allowable globs are up to us to define? (We could keep it easy on ourselves at first by deferring to `fast-glob`, and later decide what specifically we want to allow.)

thomashoneyman:
@f-f I have implemented the strategy we agreed on in this discussion:

1. Legacy import packages have some files explicitly ignored but are otherwise packaged up with all  files intact
2. Normal packages have only the `src` folder and some explicit files copied over, but users can opt-in to additional files via the optional `files` key

I believe this is up to date with all our suggested changes at this point.

ci/src/Registry/API.purs
```diff
@@ -368,18 +373,67 @@ runGit args = ExceptT do
...
-    , fixupPaths ignoredFiles
-    , Set.toUnfoldable globMatches
-    ]
+  globMatches <- FastGlob.match' ignoredGlobs { cwd: Just path, caseSensitive: false }
+  for_ (ignoredDirectories <> ignoredFiles <> globMatches) \match ->
+    FS.Extra.remove (Path.concat [ path, match ])
+
+-- | Filter out glob patterns that are susceptible to directory traversal attacks
+filterUnsafeGlobs :: FilePath -> Array String -> Aff (Array String)
+filterUnsafeGlobs path globs = case Array.uncons globs of
```
thomashoneyman: If we accept arbitrary globs then we can leak access to the file system, which would allow someone to publish a package that pulls in files from anywhere in the system and puts them in the tarball. We can take two approaches to this:

1. We can reject a package altogether if we detect it uses an unsafe glob (one that accesses anything outside of the package directory itself)
2. We can skip any globs that are unsafe, simply ignoring the user's input and taking files using safe globs only

I've taken the second path here. But I would understand if we want to `throwWithComment` instead if we can determine that the `files` key contains unsafe globs.

f-f: I don't think we should use globs at all, as they feel more of a liability than anything else (hard to parse, and to validate the security of).
We should take this list of paths (which can be files or directories), [canonicalize them](https://nodejs.org/api/fs.html#fsrealpathpath-options-callback), and only check that they only point to subdirectories and not parent directories (this can be done by checking that the current absolute path is a substring of the canonical path we're checking)

f-f: However, we can also just add a TODO in here and skip implementing the `files` field for now - we'd leave it in place and just blow up if anyone is trying to use it.

thomashoneyman: Globs are undeniably useful, though, for example for matching only `.purs` files in another directory. And if someone is going through the effort to explicitly include files via the `files` key, I suspect they're going to want more fine-grained control than just "a directory".

It's hard to parse globs (hence why we defer to `fast-glob`), but is sanitizing them any harder than sanitizing a list of directory paths?

thomashoneyman:
We had _another_ chat in the registry call this morning and agreed to go back to the NPM style as described in my comment above [^1]. That means we take all files by default, but let you selectively choose files to include via a `files` key. The reason for this decision is to preserve things like test directories so that folks downstream can still execute your tests if they want to.

[^1]: https://github.com/purescript/registry/pull/317#issuecomment-1032819187

ci/src/Registry/API.purs
f-f: > It's hard to parse globs (hence why we defer to fast-glob), but is sanitizing them any harder than sanitizing a list of directory paths?

Yes, unless the underlying library offers facilities for that. E.g. I couldn't find anything to canonicalize globs in the `fast-glob` library.

thomashoneyman: I've removed all code related to the `files` key and we can punt on it.

ci/test/Registry/Scripts/LegacyImport/Stats.purs
```diff
@@ -103,10 +102,11 @@ exampleStats = Stats.errorStats mockStats
                       ( Manifest
                           { license: mockLicense
                           , name: mockPackageName
-                          , location: Git { subdir: Nothing, gitUrl: "https://github.com/purescript/foobar.git" }
-                          , targets: Object.empty
+                          , location: Git { subdir: Nothing, gitUrl: "https://github.com/purescript/foobar" }
```
f-f: Why remove the `.git` ending?

ci/src/Registry/API.purs
```diff
@@ -368,18 +370,31 @@ runGit args = ExceptT do
...
--- version control directories. See also:
--- https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files
+-- | We always ignore some files and directories when packaging a tarball, such
+-- | as common version control directories.
+-- |
+-- | See also:
+-- | https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files
+--
+-- Picking only the `src` directory and particular accepted files will in
+-- general avoid this issue, but the user can provide arbitrary globs via
```
f-f: Which issue does this refer to?

ci/src/Registry/API.purs
```diff
@@ -368,18 +370,31 @@ runGit args = ExceptT do
...
-  traverse_ FS.Extra.remove $ Array.fold
-    [ fixupPaths ignoredDirectories
-    , fixupPaths ignoredFiles
-    , Set.toUnfoldable globMatches
-    ]
+  globMatches <- FastGlob.match' ignoredGlobs { cwd: Just path, caseSensitive: false }
+  for_ (ignoredDirectories <> ignoredFiles <> globMatches) \match ->
+    FS.Extra.remove (Path.concat [ path, match ])
+
+acceptedGlobs :: Array String
```
f-f: This seems to be unused now, so we can probably remove it?

ci/src/Registry/API.purs
thomashoneyman: Oops, this is outdated; it was referring to the issue of removing ignored files (taking only the src directory means we automatically are ignoring other files).

ci/test/Registry/Scripts/LegacyImport/Stats.purs
thomashoneyman: I’m not sure — I can’t remember why I removed it. Are we going to check or care if they include a .git ending?

f-f (approved):

PR #318, opened 2022-02-02
Ensure all packages are included in registry index, even if they have no versions

### LINKED ISSUE #302: Verify that we retain package names even if all versions are dropped

thomashoneyman:
Fixes #302 by ensuring that all packages with a valid package name are included in the registry index during the legacy import process. This does not ensure that these packages are uploaded (we're going to have to decide how we want to do that -- write an empty directory?), and the unused binding will leave a warning active until we have added them to the upload process.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -232,7 +232,7 @@ addOrUpdate { ref, legacy, packageName } metadata = do
   log $ "Hash for ref " <> show ref <> " was " <> show hash
   let newMetadata = addVersionToMetadata newVersion { hash, ref, published, bytes } metadata
   let metadataFilePath = metadataFile packageName
-  liftAff $ Json.stringifyJsonFile metadataFilePath newMetadata
+  liftAff $ Json.writeJsonFile metadataFilePath newMetadata
   updatePackagesMetadata manifestRecord.name newMetadata
   commitToTrunk packageName metadataFilePath >>= case _ of
     Left _err ->
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -190,7 +190,7 @@ addOrUpdate { ref, legacy, packageName } metadata = do
 
     runManifest gatherManifest >>= case _ of
       Left err ->
-        throwWithComment $ "Unable to produce a manifest fir legacy package: " <> err
+        throwWithComment $ "Unable to produce a manifest for legacy package: " <> err
       Right manifest ->
         liftAff $ Json.writeJsonFile manifestPath manifest
 
@@ -232,7 +232,7 @@ addOrUpdate { ref, legacy, packageName } metadata = do
   log $ "Hash for ref " <> show ref <> " was " <> show hash
   let newMetadata = addVersionToMetadata newVersion { hash, ref, published, bytes } metadata
   let metadataFilePath = metadataFile packageName
-  liftAff $ Json.writeJsonFile metadataFilePath newMetadata
+  liftAff $ Json.stringifyJsonFile metadataFilePath newMetadata
   updatePackagesMetadata manifestRecord.name newMetadata
   commitToTrunk packageName metadataFilePath >>= case _ of
     Left _err ->
@@ -296,7 +296,7 @@ wget url path = do
     NodeProcess.Normally 0 -> pure unit
     _ -> throwWithComment $ "Error while fetching tarball: " <> result.stderr
 
-mkEnv :: GitHub.Octokit -> Ref (Map PackageName Metadata) -> IssueNumber -> Env
+mkEnv :: GitHub.Octokit -> MetadataRef -> IssueNumber -> Env
 mkEnv octokit packagesMetadata issue =
   { comment: GitHub.createComment octokit issue
   , closeIssue: GitHub.closeIssue octokit issue
@@ -310,11 +310,11 @@ type MetadataRef = Ref MetadataMap
 
 mkMetadataRef :: Aff MetadataRef
 mkMetadataRef = do
+  FS.Extra.ensureDirectory metadataDir
   packageList <- try (FS.readdir metadataDir) >>= case _ of
-    Right list -> pure $ Array.catMaybes $ map (String.stripSuffix $ String.Pattern ".json") list
+    Right list -> pure $ Array.mapMaybe (String.stripSuffix $ String.Pattern ".json") list
     Left err -> do
       error $ show err
-      FS.mkdir metadataDir
       pure []
   packagesArray <- for packageList \rawPackageName -> do
     packageName <- case PackageName.parse rawPackageName of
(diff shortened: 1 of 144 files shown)


f-f:
Note that the Registry Index is not our source of truth, we just accidentally go through it while legacy-importing - I think that in order to consider #302 fixed we should write empty Metadata files for these packages.
This should happen in the `LegacyImport` side, since it's a thing only when we batch-import old names.

This is because "is a Metadata file there?" is our main check for registering a new name, see:
https://github.com/purescript/registry/blob/335c07cb00f3dada632019996d9f66eed9c6079e/ci/src/Registry/API.purs#L105-L123

thomashoneyman:
Ah, I see, I was aiming to write empty manifests — I can write the metadata files instead.

thomashoneyman:
@f-f, @colinwahl I've updated this to write reserved package names to the metadata directory. In 33b23b39cbb8e3ecbfd9d5735f61955be2c0bee3 I started tracking that directory, but I can revert the change if y'all don't feel ready to track this information or want to wait on #322 before tracking this directory.

The change is actually quite small -- just 3 files affected -- so I'd recommend reviewing by filtering only to `.purs` files.

ci/src/Registry/API.purs
```diff
@@ -232,7 +232,7 @@ addOrUpdate { ref, legacy, packageName } metadata = do
   log $ "Hash for ref " <> show ref <> " was " <> show hash
   let newMetadata = addVersionToMetadata newVersion { hash, ref, published, bytes } metadata
   let metadataFilePath = metadataFile packageName
-  liftAff $ Json.writeJsonFile metadataFilePath newMetadata
+  liftAff $ Json.stringifyJsonFile metadataFilePath newMetadata
```
thomashoneyman: I've added `stringifyJsonFile` for when we want to store data and don't need it to be super readable

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -73,32 +76,42 @@ main = Aff.launchAff_ do
...
-    packagesToUpload = Array.filter (not wasPackageUploaded) corePackages
-
   log "Starting upload..."
-  void $ for packagesToUpload \manifest -> do
+  runRegistryM (mkEnv packagesMetadataRef) do
+    log "Adding metadata for reserved package names"
+    forWithIndex_ reservedNames \package repo -> do
+      let metadata = { location: repo, releases: Object.empty, unpublished: Object.empty }
+      liftAff $ Json.stringifyJsonFile (API.metadataFile package) metadata
+      updatePackagesMetadata package metadata
```
thomashoneyman: This is the major change: before we read the packages metadata and begin uploading packages, we write out all the reserved names (updating the metadata ref along the way).

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -73,32 +76,42 @@ main = Aff.launchAff_ do
...
+
+    log "Filtering out packages we already uploaded"
```
thomashoneyman, line 94: This code is unchanged, the diff is just from shifting these lines down.

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -201,16 +214,29 @@ downloadLegacyRegistry = do
...
+    reservedNames :: Map PackageName Repo
+    reservedNames =
+      Map.fromFoldable
+        $ Array.mapMaybe
+            ( \(Tuple (RawPackageName name) address) -> do
+                parsedName <- hush $ PackageName.parse name
+                -- We don't need to preserve any packages that made it through
+                -- processing into the resulting registry index.
+                guard $ isNothing $ Map.lookup parsedName registryIndex
+                { owner, repo } <- hush $ GitHub.parseRepo address
```
thomashoneyman, line 234: I opted to only reserve names for packages that have a valid GitHub repository as well as a valid package name, because that way we can write an empty metadata file for them.

This is acceptable because only one package has a malformed package name (`TypeAhead`) and it makes sense not to reserve that name (it can't be used anyway), and because zero packages have malformed GitHub URLs so no packages will fail this check.

ci/src/Registry/API.purs
f-f: The metadata files are really the stuff that we want to be most readable, much more than the `purs.json` for example :smile: 

This is because we want nice diffs to look at, as we need to be careful about not changing the hashes ever for example (we'll add CI to ensure that, but a pair of human eyes never hurts)

So we should really revert this change

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -73,32 +76,42 @@ main = Aff.launchAff_ do
...
-    wasPackageUploaded { name, version } = API.isPackageVersionInMetadata name version packagesMetadata
-    packagesToUpload = Array.filter (not wasPackageUploaded) corePackages
-
   log "Starting upload..."
-  void $ for packagesToUpload \manifest -> do
+  runRegistryM (mkEnv packagesMetadataRef) do
+    log "Adding metadata for reserved package names"
+    forWithIndex_ reservedNames \package repo -> do
+      let metadata = { location: repo, releases: Object.empty, unpublished: Object.empty }
+      liftAff $ Json.stringifyJsonFile (API.metadataFile package) metadata
```
f-f: Here too
```suggestion
      liftAff $ Json.writeJsonFile (API.metadataFile package) metadata
```

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -73,32 +76,42 @@ main = Aff.launchAff_ do
...
+      updatePackagesMetadata package metadata
+
+    log "Filtering out packages we already uploaded"
+    packagesMetadata <- readPackagesMetadata
```
f-f, line 95: Why do we need to read this again? The ref we hold in memory should be up to date right?

f-f:
> I started tracking that directory, but I can revert the change if y'all don't feel ready to track this information or want to wait on https://github.com/purescript/registry/issues/322 before tracking this directory.

#322 is about the `registry-index`, which is downstream for here. We are definitely good to start tracking the `metadata` directory here!

ci/src/Registry/Scripts/LegacyImport.purs
thomashoneyman: The ref is up to date, but we read the value of the ref here because it's used a couple lines down for filtering out packages (that logic already existed).

f-f (approved):
This is great, thanks!

PR #266, opened 2021-11-08
Find packages that will be dropped from the package set

maxdeviant:
As described in #250, this PR adds a script to find packages that will be dropped from the package set.

The script will output a report of the packages that will be dropped, as well the reason for them being dropped.

PR #323, opened 2022-02-02
Remove NOASSERTION license

### LINKED ISSUE #250: Check what packages will be dropped from package sets
The current package set contains packages that will not be included in the new registry, and therefore won't be included in the package sets produced by the registry (since the package sets only include packages from the registry).

An example: `purescript-parsing` will be dropped because of #249, unless its LICENSE file gets fixed. It would be good to understand what other packages will be dropped from the package sets unless we fix something about them -- _especially_ packages within the core, contrib, web, and node organizations (like `purescript-parsing`).

To handle this, we would need to:

1. Get a list of packages and package versions from the latest current package set
2. Run the legacy import tool to produce a `bower-exclusions.json` file
3. Cross-reference the `bower-exclusions` against the package set to find package versions found in both places

The list of package versions that exist in both the package set and the bower exclusions represents packages currently provided by the set but which will no longer be provided going forward. This is _also_ the list of packages that we really ought to fix, if we're going to manually fix any packages, to provide a smooth upgrade experience for users.

thomashoneyman:
This PR removes the `NOASSERTION` license type from Licensee output. We take license types from various manifest files as well as trying to auto-detect from a LICENSE file (if present); the `NOASSERTION` license type represents a failed parse of a LICENSE file.

After discussion with our legal counsel we determined that we can include packages so long as we can prove we made a 'best effort' to represent the author's intent in licensing. Since we take the license from their other assertions in package manifests, and since Licensee is an automated tool that can fail to parse license files that humans would recognize as conveying meaning, we can drop these parsing failures -- we have an alternate way to determine the license (the manifest files' `license` fields).

Filtering this license type from the Licensee output prevents a NOASSERTION from failing the SPDX license check, so a good number of packages that were previously failing are now passing. Note: this does _not_ allow packages that only specify a license via a LICENSE file through if the parse fails; that results in a package being listed as having no valid license at all. So we're still only accepting licensed packages.

Helps fix #250, and closes #266, because now only packages with invalid dependencies are being dropped -- and that means only `purescript-prettier` is going to be dropped, which is pretty good!

### DIFF
--- a/ci/src/Foreign/Licensee.purs
+++ b/ci/src/Foreign/Licensee.purs
@@ -3,7 +3,7 @@ module Foreign.Licensee where
 import Registry.Prelude
 
 import Control.Parallel as Parallel
-import Foreign.JsonRepair as JsonRepair
+import Data.Array as Array
 import Foreign.Tmp as Tmp
 import Node.ChildProcess as NodeProcess
 import Node.FS.Aff as FS
@@ -33,7 +33,7 @@ detect directory = do
     NodeProcess.Normally n | n == 0 || n == 1 -> do
       let
         parse :: String -> Either String (Array String)
-        parse str = Json.parseJson (JsonRepair.tryRepair str) >>= \json -> do
+        parse str = Json.parseJson str >>= \json -> do
           obj <- Json.decode json
           licenses <- obj .: "licenses"
           spdxIds <- traverse (_ .: "spdx_id") licenses
@@ -46,7 +46,11 @@ detect directory = do
           log "arising from the result: "
           log result.stdout
           pure $ Left error
-        Right out ->
-          pure $ Right out
-    _ ->
+        Right out -> do
+          -- A NOASSERTION result means that a LICENSE file could not be parsed.
+          -- For the purposes of the registry we disregard this result, since
+          -- we retrieve the license via the package manifest(s) as well.
+          pure $ Right $ Array.filter (_ /= "NOASSERTION") out
+    _ -> do
+      log $ "Licensee process exited unexpectedly: " <> result.stderr
       pure $ Left result.stderr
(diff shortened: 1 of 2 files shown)


thomashoneyman:
This change has resulted in:

+34 packages succeeding, -4 partially succeeding, and -30 failing import
+378 versions succeeding import
-185 versions removed because of unsatisfied dependencies

Full change:

```
Packages: 1629 total (1125 totally succeeded, 332 partially succeeded, 172 totally failed)
Versions: 12628 total (10748 totally succeeded, 0 partially succeeded, 1880 totally failed)

Failures by error:
  manifestError: 1878 versions across 443 packages
    badDependencyVersions: 727 versions across 201 packages
    missingLicense: 498 versions across 147 packages
    badLicense: 424 versions across 65 packages
    badVersion: 326 versions across 144 packages
  resourceError: 0 versions across 60 packages
  noDependencyFiles: 1 versions across 1 packages

255 manifest entries with unsatisfied dependencies
```

```
Packages: 1629 total (1159 totally succeeded, 328 partially succeeded, 142 totally failed)
Versions: 12628 total (11126 totally succeeded, 0 partially succeeded, 1502 totally failed)

Failures by error:
  manifestError: 1500 versions across 409 packages
    badDependencyVersions: 727 versions across 201 packages
    missingLicense: 518 versions across 159 packages
    badVersion: 326 versions across 144 packages
  resourceError: 0 versions across 60 packages
  noDependencyFiles: 1 versions across 1 packages

70 manifest entries with unsatisfied dependencies
```

ci/src/Foreign/Licensee.purs
```diff
@@ -33,7 +33,7 @@ detect directory = do
     NodeProcess.Normally n | n == 0 || n == 1 -> do
       let
         parse :: String -> Either String (Array String)
-        parse str = Json.parseJson (JsonRepair.tryRepair str) >>= \json -> do
+        parse str = Json.parseJson str >>= \json -> do
```
thomashoneyman, line 36: This wasn't necessary to include, because Licensee produces normal JSON output without need of repair.

f-f (approved):
Looks good!

f-f:
Is #250 definitely fixed by this?

thomashoneyman:
It’s fixed in the sense that we know what is dropped from the recent package sets, but not in the sense that we could keep running this to get updated results going forward. But I don’t think we need to do that.

Issue #331, opened 2022-02-06
Proposal: Reconsider how we authenticate package location changes


thomashoneyman:
The PureScript manifest includes a field called `repository`, which indicates either a GitHub owner + repo or a URL to an arbitrary Git repository online, along with an optional `subdir` key that notes where in the repository the manifest can be found.

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/v1/Manifest.dhall#L22-L23

However, the only time we ever use the repository key (and thus the `Repo` type) is via the Registry API:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/v1/Operation.dhall#L7-L13

We _have_ to take the URL and `subdir` key before we can fetch the package in an `Addition`; without this information, we can't retrieve the manifest. We don't have to take this information in for an `Update` because we can look up the package's location via the registry index...unless the user changed the package's repository, in which case they have to submit an `Addition` instead of an `Update`, despite the package already being registered.

I think it would be simpler to store the package location in the package metadata, not the manifest. We need the package location in order to retrieve its manifest, so storing it in the manifest provides no benefit. Worse, the manifest location can easily be false; a user could send one location via the API, while storing a different location in the manifest. We could validate that the manifest matches with the API call, but this is one more brittle check in an already-complicated pipeline. (We've forgotten to include this check all along, for example!).

Storing the location in the version metadata, by comparison, means the metadata tells you exactly where the registry built _that_ version from. There is only a single source of truth: the API call, which is then put into the metadata, verified by whether we could actually retrieve the package and its manifest or not. And it's one less piece of information that has to be maintained in a manifest file by the user, or their package manager, or by us.

Nothing is lost with this change, as far as I can tell. Package managers don't use the repository information because they download packages from the registry. If they _did_ want the information, it's still available via the registry metadata, which package managers already access to get other information about packages. Package manager configurations are unaffected; they still need to tell the registry the package location, same as before.

Therefore, I'm proposing:

1. We remove the `repository` key from the manifest, and we add it to the version metadata
2. We continue to take the repository via the registry API, exactly as we do now.

thomashoneyman:
I forgot we already _do_ store this information in the version metadata! All the more reason to remove it from the manifest, where it is just duplicated and possibly incorrect information. My proposal then is actually:

1. We remove the `repository` key from the manifest
2. We otherwise make no changes

f-f:
[...]

PR #333, opened 2022-02-06
Add a TODO to authenticate repo location changes

thomashoneyman:
It's important that we check package ownership when the we're asked to change the location of a package, but we don't have any TODO comments tracking this. I'm just adding one to make sure we don't forget!


f-f:
I explained in https://github.com/purescript/registry/issues/331#issuecomment-1030825172 why I don't think we should go down this route

PR #337, opened 2022-02-07
Remove `isPackageSetInRegistry` check

### LINKED ISSUE #336: Is it safe to remove the `isPackageSetInRegistry` check?
Now that our published package sets are parametric over the `externalPackage` type - it shouldn't be possible for us to construct a value with the `Registry.External` type. 

In that case - is it safe to remove the `isPackageSetInRegistry` check?

colinwahl:
Intends to close #336 

Now that our published package sets are parametric over the externalPackage type - it isn't possible for us to construct a value with the Registry.External type. 

This removes the `isPackageSetInRegistry` as it is now guaranteed, which cleans up the package set dhall files a bit.


f-f (approved):
👏

Issue #332, opened 2022-02-06
Proposal: Unify the 'Addition' and 'Update' operations


thomashoneyman:
The registry API consists of three operations: `Addition`, `Update`, and `Unpublish`:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/v1/Operation.dhall#L10-L13

The first two operations apply when a user wants to publish their package. An `Addition` is for publishing a brand-new package the registry doesn't know about, or publishing a version of an existing package from a new location. An `Update` is for publishing a new version of a package already in the registry. They differ in a few ways: an `Update` doesn't require a location, for example, because we just look up the package's existing location we've stored.

However, I think separating an `Addition` from an `Update` is unnecessarily complicated, both for users and for us. Their implementations are essentially the same, modulo some verification checks to ensure (for example) that you aren't trying to `Update` a package that doesn't exist (and therefore you should have used an `Addition`). I think we can do away with the distinction between the two and just have a `Publish` operation and an `Unpublish` operation.

For context, here's the implementation of an `Addition`:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/ci/src/Registry/API.purs#L108-L114

And here's the implementation of an `Update`:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/ci/src/Registry/API.purs#L116-L123

In both cases we check for the existence of a metadata file. If it's an `Addition` and we can't find one, we create it; if we do find one, then we redirect to be an `Update` instead. If it's an `Update` and we can't find a metadata file, we just throw an exception. Otherwise, though, we just move directly into `addOrUpdate`, where we use the metadata's `location` field to fetch the repository to package, at which point we switch to using the package's manifest for further processing.

This all has me asking: why is the distinction between an `Addition` and an `Update` necessary? There isn't much difference between the two operations, and these distinctions lead to unnecessary ambiguity. In each case we can have a clearer API by unifying the options:

[...]

Issue #328, opened 2022-02-04
Clarify stance on legacy support


thomashoneyman:
The PureScript registry is built to support packages using the [new manifest format](https://github.com/purescript/registry/blob/master/v1/Manifest.dhall) in a `.purs.json` file. However, we also support importing legacy packages from bower (#215), and the API supports adding new versions of legacy packages via the `legacy` flag:

https://github.com/purescript/registry/blob/d69c9baf804fb7040881a5c388c2926643393bb4/v1/Operation.dhall#L10-L11

However, the spec doesn't include any mention of what support we do and do not have for "legacy" (ie. Bower or Spago pre-registry) packages. We need to clarify this.

The registry will provide some support for legacy packages for at least the first year of operation. Registry support only means that we guarantee legacy packages can be added to the registry; we do not guarantee that registry packages remain compatible with legacy package managers like Bower. We do, however, guarantee support for older versions of Spago. In practice this means that we will preserve the existing package sets repository and continue publishing package sets usable with older versions of Spago there.

A legacy package is defined as a package that is published without a `.purs.json` manifest file, and it can be one of two things: a project containing a `bower.json` file, or a project containing a `spago.dhall` file and accompanying `packages.dhall` (ie. a Spago file using a configuration format pre-registry -- not [the new proposed Spago format](https://github.com/purescript/spago/issues/842)).

A legacy package can be uploaded to the registry via an `Addition` or `Update` operation by setting `legacy: true`. Setting the legacy flag means that the registry will:

1. Attempt to parse the package's legacy package files, such as `bower.json`, `spago.dhall`, and `packages.dhall` files, in order to create a `.purs.json` manifest for the package
2. Include the generated `.purs.json` file in the resulting package tarball
3. Attempt to use the legacy package files to publish the package version to Pursuit (this will fail for Spago)
4. Attempt to use the generated `.purs.json` file to publish the package version to the package sets

Legacy packages cannot use monorepo setups because their version is parsed from the `ref` provided in the addition or update operation. Attempting to upload legacy packages via a monorepo setup will fail.

---

@f-f I do need your help understanding exactly what has to happen to guarantee the `packages.dhall` file in the package-sets repo remains compatible with old Spago clients for the next year, even as we're publishing package sets using the new package sets format in the registry. This may need its own issue to track backwards-compatible mirroring from the registry to the package sets repo.

thomashoneyman:
[...]

PR #338, opened 2022-02-07
Remove `legacy` and `addToPackageSet` flags from operations

thomashoneyman:
As described in #332:

> > An `Addition` takes an `addToPackageSet` flag, and an `Update` does not
> 
> This is a historical accident that came from the fact that we added this flag long before we specced properly how to do package sets here. I think we could just take this flag away and always try to add packages to the set?
> 
> > removing the `legacy` flag. We know a package is a legacy package if it doesn't have a `.purs.json` file
> 
> I think this is a good idea, yeah. Again the `legacy` flag was added when we did not know better - now we know that we can just use the Manifest existence to decide which path to take in the API

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+.psc*
 examples
 registry-index
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -105,21 +105,21 @@ readOperation eventPath = do
 runOperation :: Operation -> RegistryM Unit
 runOperation operation = case operation of
   -- TODO handle addToPackageSet
-  Addition { packageName, legacy, newRef, newPackageLocation } -> do
+  Addition { packageName, newRef, newPackageLocation } -> do
     -- check that we don't have a metadata file for that package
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
+.psc*
```
thomashoneyman, line 1: Every time I accidentally open the root of the registry repository I get a `.psc-ide-port` file that I have to remember not to commit 😆

ci/src/Registry/API.purs
```diff
@@ -168,7 +168,7 @@ addOrUpdate { ref, legacy, packageName } metadata = do
   log $ "Package extracted in " <> absoluteFolderPath
 
   -- If this is a legacy import, then we need to construct a `Manifest` for it
-  when legacy do
+  unlessM (liftAff $ FS.exists manifestPath) do
```
thomashoneyman, line 171: As [described in this issue](https://github.com/purescript/registry/issues/328#issuecomment-1030534576), a "legacy" package is any package without a `purs.json` manifest file.

f-f (approved):

PR #340, opened 2022-02-08
Standardize use of Node.Path

thomashoneyman:
While working on a PR recently I noticed that we are a bit inconsistent with how we handle paths. In some places we concatenate paths with `<> "/" <>` and in others we use Node's path concatenation. Sometimes we import `Node.Path` as `Path`, or as `FilePath`, or as `Node.Path`.

This PR standardizes us on using Node path concatenation when working with file paths (to keep compatibility with Windows users) and importing qualified as `Path`.

### DIFF
--- a/ci/src/Registry/PackageUpload.purs
+++ b/ci/src/Registry/PackageUpload.purs
@@ -6,6 +6,7 @@ import Data.Array as Array
 import Effect.Aff as Aff
 import Foreign.S3 as S3
 import Node.FS.Aff as FS
+import Node.Path as Path
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Version (Version)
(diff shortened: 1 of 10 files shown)


ci/src/Registry/PackageUpload.purs
```diff
@@ -28,12 +29,7 @@ upload { name, version } path = do
   -- check that the file for that version is there
   let
     packageName = PackageName.print name
-    filename = Array.fold
-      [ packageName
-      , "/"
-      , Version.printVersion version
-      , ".tar.gz"
-      ]
+    filename = Path.concat [ packageName, Version.printVersion version <> ".tar.gz" ]
```
f-f: I don't think we should change this, it's the key that it's used when reading from DO APIs

thomashoneyman: Great catch! You're right, this is not a filepath.

f-f (approved):

PR #346, opened 2022-02-10
Make published field optional

### LINKED ISSUE #342: Make published date optional in version metadata, unless we can guarantee we can always get it
We store a package's publication date in metadata:

https://github.com/purescript/registry/blob/fb20bec4af37a96353bc98c3f30391c77e2472d9/ci/src/Registry/Schema.purs#L164

...which we retrieve in the GitHub provider by looking at the commit:

https://github.com/purescript/registry/blob/fb20bec4af37a96353bc98c3f30391c77e2472d9/ci/src/Registry/API.purs#L151

However, since we are provider-agnostic, there is no guarantee that we will be able to retrieve the date of a given `ref`. For example, in the registry call this morning @f-f described fetching from a filesystem online, where we may not necessarily be able to retrieve a published date.

This isn't critical information; there's no harm in making it optional if we do end up always being able to figure this out, but there is harm in changing this field from required to optional as far as package managers are concerned (it'd be a breaking change for anyone consuming metadata).

thomashoneyman:
Fixes #342 by making the `published` field optional in package metadata.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -231,7 +231,7 @@ addOrUpdate { ref, packageName } metadata = do
   uploadPackage uploadPackageInfo tarballPath
   log $ "Adding the new version " <> Version.printVersion newVersion <> " to the package metadata file (hashes, etc)"
   log $ "Hash for ref " <> show ref <> " was " <> show hash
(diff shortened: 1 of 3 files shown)


thomashoneyman:
Closing as per the discussion in #342.

PR #347, opened 2022-02-10
Centralize tests in one location

thomashoneyman:
While working on a few other registry-related things, I noticed that we have CI checks spread in a few locations. This PR centralizes all testing related to registry development into one place (our test suite), which means fewer places to keep in mind when adjusting our test infrastructure.

I've verified that these modified checks fail if you leave code unformatted or you write a Dhall file that doesn't typecheck.


thomashoneyman:
Well, this is weird! Running `nix-shell ../shell.nix --run "spago test"` or just `spago test` locally works just fine. But in CI the `dhall` CLI tool is just hanging until the tests time out. I wouldn't expect it to be a permissions issue, because we do exactly the same thing with the `dhall-to-json` tool and the `examples` directory.

I'm not exactly sure what's going on here. Any ideas?

thomashoneyman:
Huh. For some reason it's just the `Address.dhall` file that's hanging. All the other files typecheck just fine.

thomashoneyman:
Well, I have no clue why the `Address.dhall` file hangs when checked by Dhall in CI. I guess if we don't know why this is, then I can close this PR and just leave that test in a shell script. But it feels like there has to be some obvious issue that I'm not seeing.

f-f:
The only difference with the other files is that they typecheck to `Type`, while `Address` is a `Type -> Type`.

Though the command you're using works just fine locally, so I'm puzzled about this. I can take a deeper look in the next days if you'd like

thomashoneyman:
If you find a moment where you’re curious! But it isn’t high priority.

PR #349, opened 2022-02-10
Add stm

the-dr-lazy:


thomashoneyman:
Sorry, it looks like `purescript-stm` is already registered:
https://github.com/purescript/registry/blob/7d5cf1ef99d80ee5db96b714caed497b0a5468bc/bower-packages.json#L1248

PR #350, opened 2022-02-12
Implement authenticated operations

thomashoneyman:
I don't think this technically fixes any issues, but it relates to the conversation in #155 and implements an authorized unpublish operation as per https://github.com/purescript/registry/issues/155#issuecomment-1034142912.

### DIFF
--- a/ci/package.json
+++ b/ci/package.json
@@ -15,6 +15,7 @@
     "@octokit/rest": "^18.12.0",
     "aws-sdk": "^2.1050.0",
     "fast-glob": "^3.2.11",
+    "decimal.js": "^10.3.1",
     "fs-extra": "^10.0.0",
     "fuse.js": "^6.5.3",
     "jsonrepair": "^2.2.1",
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -18,6 +18,7 @@ import Foreign.GitHub (IssueNumber)
 import Foreign.GitHub as GitHub
 import Foreign.Node.FS as FS.Extra
 import Foreign.Object as Object
+import Foreign.SSH as SSH
 import Foreign.Tar as Tar
 import Foreign.Tmp as Tmp
 import Node.ChildProcess as NodeProcess
@@ -32,7 +33,7 @@ import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.PackageUpload as Upload
 import Registry.RegistryM (Env, RegistryM, closeIssue, comment, commitToTrunk, readPackagesMetadata, runRegistryM, throwWithComment, updatePackagesMetadata, uploadPackage)
-import Registry.Schema (Manifest(..), Metadata, Operation(..), Location(..), addVersionToMetadata, isVersionInMetadata, mkNewMetadata)
+import Registry.Schema (AuthenticatedData(..), AuthenticatedOperation(..), Manifest(..), Metadata, Operation(..), Location(..), addVersionToMetadata, isVersionInMetadata, mkNewMetadata, unpublishVersionInMetadata)
 import Registry.Scripts.LegacyImport.Error (ImportError(..))
 import Registry.Scripts.LegacyImport.Manifest as Manifest
 import Registry.Types (RawPackageName(..), RawVersion(..))
@@ -121,7 +122,40 @@ runOperation operation = case operation of
         addOrUpdate { packageName, ref: updateRef } metadata
       (throwWithComment "Metadata file should exist. Did you mean to create an Addition?")
 
-  Authenticated _ -> throwWithComment "Authenticated operations not implemented! Ask us for help!" -- TODO
+  Authenticated auth@(AuthenticatedData { payload }) -> case payload of
+    Unpublish { packageName, unpublishReason, unpublishVersion } -> do
+      let metadataFilePath = metadataFile packageName
+      liftAff (FS.exists metadataFilePath) >>= case _ of
+        false -> throwWithComment "No metadata file found. Did you mean to create an Addition?"
+        _ -> pure unit
+
+      metadata <- readPackagesMetadata >>= \packages -> case Map.lookup packageName packages of
+        Nothing -> throwWithComment "Couldn't read metadata file for your package."
+        Just m -> pure m
+
+      case metadata.owners of
+        Nothing ->
+          throwWithComment $ String.joinWith " "
+            [ "Cannot verify package ownership because no owners are listed in the package metadata."
+            , "Please publish a package version with your SSH public key in the owners field."
+            , "You can then retry unpublishing this version by authenticating with your private key."
+            ]
+        Just owners ->
+          liftAff (SSH.verifyPayload owners auth) >>= case _ of
+            Left err ->
+              throwWithComment $ String.joinWith "\n"
+                [ "Failed to verify package ownership:"
+                , "  " <> err
+                ]
+            Right _ -> do
+              let updatedMetadata = unpublishVersionInMetadata unpublishVersion unpublishReason metadata
+              liftAff $ Json.writeJsonFile metadataFilePath updatedMetadata
+              updatePackagesMetadata packageName updatedMetadata
+              commitToTrunk packageName metadataFilePath >>= case _ of
+                Left _err ->
+                  comment "Unpublish succeeded, but metadata could not be updated.\n\ncc @purescript/packaging"
+                Right _ -> do
+                  comment "Package version unpublished!"
 
 metadataDir :: FilePath
 metadataDir = "../metadata"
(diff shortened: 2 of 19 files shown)


ci/src/Registry/Schema.purs
```diff
@@ -177,6 +240,13 @@ addVersionToMetadata version versionMeta metadata = do
   let releases = Object.insert (Version.printVersion version) versionMeta metadata.releases
   metadata { releases = releases }
 
+unpublishVersionInMetadata :: Version -> String -> Metadata -> Metadata
+unpublishVersionInMetadata version unpublishReason metadata = do
+  let versionKey = Version.printVersion version
+  let releases = Object.delete versionKey metadata.releases
+  let unpublished = Object.insert versionKey unpublishReason metadata.unpublished
+  metadata { releases = releases, unpublished = unpublished }
```
thomashoneyman: I don't fully recall the plan for unpublishing. For now I just delete the version from the metadata, so that querying the metadata releases doesn't list the version anymore, but presumably we'll also want to delete it in the storage backend?

v1/Metadata.dhall
```diff
@@ -9,25 +9,31 @@ let Map = (./Prelude.dhall).Map.Type
...
-  {
   -- The git ref this version points to
-  , ref : Text
+  { ref : Text
   -- The hash of the source tarball fetched from the repo
   , hash : Text
   -- The size in bytes of the tarball
   , bytes : Natural
+  -- The published date of the version as an RFC3339String
+  , published : RFC3339String
```
f-f: Dhall [now supports datetimes](https://github.com/dhall-lang/dhall-lang/pull/1191) - so we should probably use that here?
You can look at the tests in the linked PR to see how it looks like, but the TLDR is that things like `2000-01-01T12:00:00+08:00` typecheck to `DateTime` where `DateTime = { date : Date, time : Time, timeZone : TimeZone }`

Though we should check how they are dealt with in `json-to-dhall`, since that's our usage.

ci/src/Registry/Schema.purs
f-f: Yes, and it's a bit more involved than this:
- we want to pull the version out of the `published` versions and into the `unpublished` versions.
  Then, we have to pick one of these:
  1. we implement how to deal with multiple unpublished versions having the same version number
  2. or when publishing a new version we check that the new version is not in the list of published _and_ unpublished versions (right now we check only published IIRC).
  I'm for option (2)
- we absolutely want to pull the unpublished version from the storage backend(s)
- we should allow unpublishing only for a limited time after the publishing happened (e.g. 48h) or anytime for DMCA reasons, but only by the packaging team. We can probably just match on GitHub usernames for the latter instance.
- as an aside: we probably want to store the time a version has been unpublished, together with keeping the time it was published at. E.g.
  ```dhall
  let UnpublishedVersion = { reason : Text,  unpublishedDate : DateTime, publishedDate : DateTime }
  ```

ci/src/Registry/API.purs
```diff
@@ -122,7 +123,40 @@ runOperation operation = case operation of
...
-  Unpublish _ -> throwWithComment "Unpublish not implemented! Ask us for help!" -- TODO
+  Authenticated auth@(AuthenticatedData { payload }) -> case payload of
+    Unpublish { packageName, unpublishReason, unpublishVersion } -> do
+      let metadataFilePath = metadataFile packageName
+      liftAff (FS.exists metadataFilePath) >>= case _ of
+        false -> throwWithComment "No metadata file found. Did you mean to create an Addition?"
+        _ -> pure unit
+
+      metadata <- readPackagesMetadata >>= \packages -> case Map.lookup packageName packages of
+        Nothing -> throwWithComment "Couldn't read metadata file for your package."
```
f-f: This is an internal error, and we should probably get a ping to `@purescript/packaging`

f-f (commented):
This looks lovely, great work @thomashoneyman!

I left a couple of minor comments, but this is almost ready to go

v1/Metadata.dhall
thomashoneyman: I think we should keep the synonym, because Dhall does not support encoding `DateTime` to JSON[^1].

[^1]: https://hackage.haskell.org/package/dhall-json-1.7.9/docs/Dhall-JSON.html

thomashoneyman:
@f-f I believe I've addressed everything other than letting the packaging team unpublish things. I'm not sure the best way to handle permissions there (just store a list of GitHub usernames in code, and trust code review in the future to prevent it being updated?), and it's not vital that it's done right now.

If you have an idea for the best way to do it then I can add it to this PR.

ci/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@octokit/rest": "^18.12.0",
     "aws-sdk": "^2.1050.0",
+    "decimal.js": "^10.3.1",
```
thomashoneyman, line 18: The `precise-datetime` library relies on `decimal` under the hood, and I forgot to install it earlier. However, I _would_ like to just switch to our own `ISO 8601` type later on.

ci/src/Registry/Schema.purs
```diff
@@ -154,32 +214,45 @@ type UnpublishData =
...
+  , unpublished :: Map Version UnpublishedMetadata
   }
 
-type VersionMetadata =
+type PublishedMetadata =
   { ref :: String
   , hash :: Sha256
-  , published :: RFC3339String
   , bytes :: Number
+  , timestamp :: RFC3339String
```
f-f: I find `timestamp` less informative than `published`, could we switch this back?

This would also flatten out the `published` and `unpublished` fields in the `UnpublishedMetadata` below

ci/src/Registry/API.purs
```diff
@@ -325,9 +422,18 @@ mkMetadataRef = do
         Aff.throwError $ Aff.error $ StringParser.printParserError err
     let metadataPath = metadataFile packageName
     metadataStr <- FS.readTextFile UTF8 metadataPath
-    metadata <- case Json.parseJson metadataStr of
+    metadataRaw <- case Json.parseJson metadataStr of
       Left err -> Aff.throwError $ Aff.error $ "Error while parsing json from " <> metadataPath <> " : " <> err
       Right r -> pure r
+    let
+      fixupMetadata = do
```
f-f: It took me a while to understand what we are "fixing up" here, maybe we could switch to a more informative name, e.g. something like `parseMetadataVersionKeys`?

ci/src/Registry/Schema.purs
thomashoneyman: I changed it because:

* I think it's more consistent to refer to published and unpublished releases via the key names `published` and `unpublished` rather than the key names `releases` and `published` (we talk about 'published releases' and 'unpublished releases' frequently)
* Since the names `published` and `unpublished` are already being used to describe particular releases, it feels a little confusing to also use those names for timestamps, ie. `["prelude"].published["1.0.0"].published` to get the time the release was published, or `["prelude"].unpublished["1.0.0].unpublished` to get the time when it was unpublished. Then again, I suppose plenty of APIs do things like this.

I'd really like to stick with using `published` to refer to published releases, but if you'd like to switch the `timestamp` fields back to being `published` or `unpublished` then I'm OK with that.

```dhall
{ published : Map SemVer { published : RFC3339String, ... }
, unpublished : Map SemVer { published : RFC3339String, unpublished : RFC3339String, ... }
}
```

f-f: Yeah, that makes sense. How would you feel about `publishedTime` and `unpublishedTime`?

thomashoneyman: That's fine with me!

f-f: Cool, let's go with this then!

ci/src/Registry/API.purs
```diff
@@ -117,12 +124,92 @@ runOperation operation = case operation of
...
+                { unpublished = mapKeys Version.printVersion updatedMetadata.unpublished
+                , published = mapKeys Version.printVersion updatedMetadata.published
+                }
+
+              updatePackagesMetadata packageName updatedMetadata
+              commitToTrunk packageName metadataFilePath >>= case _ of
+                Left _err ->
+                  comment "Unpublish succeeded, but metadata could not be updated.\n\ncc @purescript/packaging"
+                Right _ -> do
+                  comment "Package version unpublished!"
```
f-f, line 212: I just realized: we want to delete the tarball from the storage backend here, after the unpublish went through

thomashoneyman: Oops, that’s right. I’ll do that when I’m back on Tuesday!

thomashoneyman:
@f-f I've added support for deleting packages from S3, but I don't have the SPACES keys set up in my `.env` so I'm unable to test it out. That said, I believe this is ready to go.

f-f (approved):
Looks great now! Thank you for doing this :heart: 

And no worries about testing, we can test this out with a bogus package or a package that we control once we run the first mass-import

Issue #330, opened 2022-02-06
Proposal: Don't privilege GitHub as a git provider in the manifest format


thomashoneyman:
PureScript packages contain a manifest that indicates where the package contents are stored. This reference is described by the `Repo` type, which stores the location of a Git repository and an optional subdirectory indicating where the package is stored:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/v1/Repo.dhall#L12-L24

This `Repo` type can be a URL to a Git repository, or it can be a special GitHub format which we privilege because we can use the GitHub API to retrieve information about the package:

https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/v1/Repo.dhall#L3-L8

However, I feel that privileging GitHub runs counter to our goals to be platform-agnostic, and in fact we don't need to privilege it at all in order to use the GitHub API. We know a package is on GitHub if the package repository URL contains 'github.com', after all! Since we can tell who is hosting the package based on the Git URL, there's no need to provide a special format for GitHub packages, and in fact our compatibility guarantee for the manifest means that we have to be careful about [adding new providers like Sourcehut](https://github.com/purescript/registry/pull/204) with their own special format later on.

Of course, we could decide to never add special formats for other providers. Or we could adopt a convention like always prefixing object keys with the provider name, like how we do `{ githubOwner, githubRepository }` for GitHub, so it's always unambiguous. But this all feels unnecessary when we could instead only accept simple Git URLs and decide who the provider is based on the URL.

So that's my proposal: don't provide a special format for GitHub or any other provider, and instead only accept the simple URL to a git repository that all non-GitHub projects would need to use anyway.

f-f:
I feel that framing this as "privileging" a provider versus another one is misleading. The main reason why we store information about a package coming from GitHub is so that we can fetch nicely packaged tarballs instead of having to clone repositories (that is sometimes a hairy business, e.g. dealing with symlinks, etc):
https://github.com/purescript/registry/blob/62b93f1cf16d97520979925b5a0fcc525f6c77fc/ci/src/Registry/API.purs#L154-L156

We are not enabling anything just because a package comes from GitHub, but the fact that we have this information allows us to exploit that and use special APIs. I don't see a reason why we should lose this information and make everything go through the same code path: folks are free to go through the generic `Git` path, but I feel like we should _add_ more providers if they have special APIs we can use to make our life easier, rather than take away this possibility.

[...]

PR #351, opened 2022-02-15
Rename repository field to location

### LINKED ISSUE #345: Discussion: Use of 'repository' terminology
In https://github.com/purescript/registry/issues/331#issuecomment-1032266571, @f-f made it clear that packages do not need to be stored in Git repositories; in fact, all we really care about is that we can retrieve a package's source files given a location and a unique reference to the source files corresponding with a particular version. @f-f mentioned the terminology "tarball provider" as a more accurate name than "git repository" for picking up package sources from ie. the GitHub API.

However, the spec and schemas make many references to packages being Git repositories, and the use of "repository" to describe where we get a package's sources from is used pervasively.  

Examples include the `Repo` type:

https://github.com/purescript/registry/blob/a9b6251d4d9a6629e72baa8e53ec9f6ba1e6a7ca/v1/Repo.dhall#L3

and the `Manifest` type:

https://github.com/purescript/registry/blob/f62ee2560f1de4a5fc90becfed13511239559f48/v1/Manifest.dhall#L22-L23

and the `Metadata` type:

https://github.com/purescript/registry/blob/cc22a35f0fecf5d88efe88e90b2162a633a3924e/v1/Metadata.dhall#L17-L18

I think we need to remove these references to "git repositories" and "git refs". In addition, I think we should seriously consider renaming the `repository` field in the manifest and the `Repo` type to terminology that makes it more obvious that packages don't _have_ to be Git repositories, even if that's all we support at first.

The `Metadata` type already uses a potential alternative:

https://github.com/purescript/registry/blob/cc22a35f0fecf5d88efe88e90b2162a633a3924e/v1/Metadata.dhall#L27-L28

I feel like the name `location` and the description "a pointer to the code" are much more accurate to the registry's agnostic stance than the name `repository`, and I think it's worth considering standardizing on this instead.

thomashoneyman:
Fixes #345 by renaming `repository` to `location` throughout the schema and application. Feels much nicer!


thomashoneyman:
I also went ahead with the `gitUrl` rename as discussed in #330

f-f (approved):
This is great, thank you!

PR #358, opened 2022-02-28
Implement authenticated transfer operation

thomashoneyman:
As per the discussion in https://github.com/purescript/registry/issues/331#issuecomment-1040992522, this implements a `Transfer` operation that allows users to transfer their package from one location to another.


f-f (approved):
Looks great, thank you!

PR #361, opened 2022-03-10
Update Map JSON instance so it is usable with non-coercible keys

thomashoneyman:
The current `Map` instance requires that keys can be coerced to strings, because it encodes to JSON as an object (which requires string keys). However, this is a problem for frequently-used smart constructors in the codebase like `PackageName` and `Version`.

This PR replaces the `Coercible` constraint in the class with a new `StringEncodable` constraint, which is a class specifically for values that can be encoded as JSON strings (ie. which can be used as map keys when written to JSON). This removes the need to constantly call `mapKeys` and `traverseKeys` when encoding and decoding maps using smart constructors as their keys.

### DIFF
--- a/ci/src/Registry/Schema.purs
+++ b/ci/src/Registry/Schema.purs
@@ -12,7 +12,6 @@ import Registry.Json as Json
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Version (Range, Version)
-import Text.Parsing.StringParser as StringParser
 
 -- | PureScript encoding of ../v1/Manifest.dhall
 newtype Manifest = Manifest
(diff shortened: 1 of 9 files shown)


ci/src/Registry/Schema.purs
```diff
@@ -37,13 +36,11 @@ instance RegistryJson Manifest where
...
     "owners" := fields.owners
     "description" := fields.description
-    "dependencies" := mapKeys PackageName.print fields.dependencies
+    "dependencies" := fields.dependencies
 
   decode json = do
     manifestFields <- Json.decode json
-    let parse = lmap StringParser.printParserError <<< PackageName.parse
-    parsed <- traverseKeys parse manifestFields.dependencies
-    pure $ Manifest $ manifestFields { dependencies = parsed }
```
thomashoneyman, line 46: The point of this PR is to avoid having to write this code out every time we use a `PackageName` or `Version` or other smart constructor as map keys.

f-f (approved):
Lovely!

Issue #360, opened 2022-03-10
Check that compile build as part of uploading


colinwahl:
As part of our package[1] upload CI checks, we'd like to ensure that packages compile, and reject the upload if they don't.

To do this, we need to figure out a few things:
1. What version of the PureScript compiler do we use?
2. How do we compile packages if they may be for different backends?
3. How do we get all of the dependencies for the package (required for `purs` to compile)

For (1), @thomashoneyman and I were thinking that we could use the last two major versions of the compiler, first building with the latest major version, then checking with the last major version, then failing if the package compiled with neither. Should this check fail, we can notify the registry maintainers to see if something needs to change. The motivation behind this is that only `0.13+` is available through `easy-purescript-nix` at the moment. Presumably `0.13` support will stay once `0.15` releases, so maybe we can expand to checking more versions in the future.

For (2), we should be able to compile with `purs compile ... -g corefn`, which should be able to type check the package regardless of the backend used.

For (3), there are two options: right our own solver, or push that to the user (i.e. the package manager). I believe that the latter option is preferred. To handle this, we can add some sort of `Map PackageName Version` parameter which is a list of specific versions for each transitive dependency required for the package to build. We can then fetch the tarballs from the registry to get access to the source code for each dependency & pass them in to the `purs compile` check.

[1] really, this is a package at a specific version, but I'll just call it a package

f-f:
What's the usecase for compiling packages when uploading? Only to generate docs? If yes, in https://github.com/purescript/registry/issues/326 we agreed that, as you mention in (3), users will be required to provide a build plan (including the compiler version) when publishing a package.

thomashoneyman:
We have to do this to generate docs, and we could also say that your package has to compile to be registered if we wanted to. The implementation is the same either way, it's just whether we:

1. Run checks and register the package, then compile the docs, and fail to push to Pursuit if compilation fails
2. Add compilation to the checks, considering failed compilation a failed check, so the package is only registered if it compiles. At that point, push the docs to Pursuit (which now will definitely succeed, modulo network errors).

thomashoneyman:
We've started work on this, by the way!

f-f:
Right, I see. I feel like compilation being part of the checks is the option that makes most sense!

thomashoneyman:
This was completed in #402.

PR #365, opened 2022-03-16
Clone instead of downloading tarballs

### LINKED ISSUE #356: Adjust publishing to always clone repositories, not fetch tarballs
We have added support for publishing packages with `purs.json` files to `purs publish`[^1]. However, publishing that way has a few restrictions:

* Packages must be hosted on GitHub
* When publishing, you must have a valid SemVer tag checked out with a clean Git working tree to publish from (or an untracked `purs.json` file)
* There must be a `bower.json` file or a `purs.json` file in the root of the repository

Unfortunately, we can't use `purs publish` with our current publishing pipeline because we fetch tarballs directly, which means there is no Git information to draw from [^2]. Instead, we need to always clone the Git repository so that we can check out the proper tag and run `purs publish`.

Of course, always cloning a Git repository is against the design of the registry overall, which is not tied to Git, to tagged versions, or to GitHub. Still, we already have guards in place that prevent packages from using all of the registry's features [^3] already, because supporting them would constitute breakage for older package managers. Always cloning the Git repository is just another restriction in place until we can implement proper support for the registry in `purs publish`.

Full support for `purs publish` requires breaking changes both to the compiler and to Pursuit and needs a full specification of the changes we need to make, but we can at least ensure everything works in compatibility mode here in the registry by cloning instead of fetching tarballs.

[^1]: https://github.com/purescript/purescript/pull/4233
[^2]: See [Registry.API L151-163](https://github.com/purescript/registry/blob/8dcf5c74b2f9b61b525c0da15a5139281e8e6d8b/ci/src/Registry/API.purs#L151-L163)
[^3]: Specifically, we ban packages not from GitHub ([Registry.API L142-143](https://github.com/purescript/registry/blob/8dcf5c74b2f9b61b525c0da15a5139281e8e6d8b/ci/src/Registry/API.purs#L142-L143) and #15) and the use of `subdir` ([Registry.API L145-146](https://github.com/purescript/registry/blob/8dcf5c74b2f9b61b525c0da15a5139281e8e6d8b/ci/src/Registry/API.purs#L145-L146) and #16)

thomashoneyman:
Fixes #356 by cloning repositories instead of fetching tarballs from GitHub. This is only necessary because of the legacy `purs publish` workflow we're currently supporting, but as soon as `purs publish` does not require a Git repository to be checked out we can go back to tarballs.

We could possibly keep the existing tarball implementation, and then clone the repo when we go to produce the Pursuit docs, but this felt like needless duplication — the tarball implementation is easy to put back when we're ready for it.

This is also necessary to complete #360.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -13,6 +13,7 @@ import Data.Interpolate (i)
 import Data.JSDate as JSDate
 import Data.Map as Map
 import Data.PreciseDateTime as PDT
+import Data.RFC3339String (RFC3339String)
 import Data.RFC3339String as RFC3339String
 import Data.Set as Set
 import Data.String as String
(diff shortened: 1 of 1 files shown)


f-f:
This looks good, but I'd like to keep the current tarball implementation behind a flag - this is so that:
1. we are forced to keep it compiling, so it's easy to go back
2. whenever we need to switch we can just remove the flag and the temporary code, without fiddling with anything else

I mean something like:
```purescript
let temporarilyCloneRepos = true
if temporarilyCloneRepos
then cloneRepo
else fetchTarball
```

thomashoneyman:
Well, the checks aren't running due to GitHub's outages, but I've run the tests locally and they pass.

f-f (approved):
Looks great, thanks!

PR #367, opened 2022-03-21
Add build plan to Addition and Update operations

thomashoneyman:
As part of #360, this adds support for a `buildPlan` argument to the addition and update operations. This allows us to compile package documentation in the API and push the result to Pursuit. Without the build plan we don't have a way to produce the resolutions format the compiler requires for package publishing, other than calling out to Spago or Bower in CI (which I'd prefer not to do).

### DIFF
--- a/ci/src/Registry/Schema.purs
+++ b/ci/src/Registry/Schema.purs
@@ -59,8 +59,6 @@ derive newtype instance Eq Owner
 derive newtype instance Show Owner
 derive newtype instance RegistryJson Owner
 
--- TODO: ToEncodable class
-
 -- | A compiler version and exact dependency versions that should be used to
 -- | compile a newly-uploaded package as an API verification check.
 -- |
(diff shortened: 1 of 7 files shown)


ci/src/Registry/Schema.purs
```diff
@@ -58,6 +59,32 @@ derive newtype instance Eq Owner
 derive newtype instance Show Owner
 derive newtype instance RegistryJson Owner
 
+-- TODO: ToEncodable class
+
```
thomashoneyman:
```suggestion
```

thomashoneyman:
Yeesh! I always forget to import things in Dhall.

f-f (approved):
Looks great!

The only thing I'm wondering about is if we should make the build plan optional. Though it would complicate the logic, and I'm not sure what other usecases we'd have other than for the legacy packages uploading, so this is probably just an empty concern.

thomashoneyman:
We can always relax required -> optional without breaking anything. But we’d want to implement our own solver then so we can produce the build plan (or call out to Bower in CI, or Spago in CI, for legacy packages).

PR #369, opened 2022-03-21
Fix Eq instances for Version and Range

thomashoneyman:
The `Version` and `Range` types use derived `Eq` instances, which will compare on all fields of the record. For example, this means that two versions `1.0.0` will not be considered equal if their parse modes were different, or their raw input strings were different. 

This PR fixes these instances to only consider the parsed versions in the equality checks.

### DIFF
--- a/ci/src/Registry/Version.purs
+++ b/ci/src/Registry/Version.purs
@@ -128,7 +128,7 @@ newtype Range = Range
   }
 
 instance Eq Range where
-  eq (Range a) (Range b) = a.lhs == b.lhs && a.rhs == b.rhs
+  eq = eq `on` (\(Range { lhs, rhs }) -> [ lhs, rhs ])
 
 instance RegistryJson Range where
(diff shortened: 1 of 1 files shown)


ci/src/Registry/Version.purs
```diff
@@ -126,7 +127,8 @@ newtype Range = Range
   , raw :: String
   }
 
-derive instance Eq Range
+instance Eq Range where
+  eq (Range a) (Range b) = a.lhs == b.lhs && a.rhs == b.rhs
```
colinwahl: Any reason to not make this

```
eq = eq `on` (\(Range r) -> [ r.lhs, r.rhs ])
```
?

colinwahl: Any reason to not make this

```
eq = eq `on` (\(Range r) -> [ r.lhs, r.rhs ])
```
?

colinwahl (approved):

PR #372, opened 2022-03-24
Add Spec helper for running RegistryM

thomashoneyman:
@colinwahl and I have encountered a few cases in which it would be helpful to test a function that runs in `RegistryM`. However, test specs by default run in `Aff`. This helper allows us to write specs for functions in `RegistryM` by providing a dummy environment.

Ideally we continue to limit how many functions _must_ be run in `RegistryM`, but some functions rely on things like logging and are tedious to pull apart.


f-f (approved):
Great idea!

PR #377, opened 2022-04-04
Add Cyril's stale packages script

JordanMartinez:
I'm not sure if this URL check is already done in CI or elsewhere, but this is the script @kl0tl used before running the script that released all packages across core, contrib, node, and web in the 0.14 ecosystem updates.

I ran it just now and there aren't any stale packages.

### DIFF
--- a/stale-packages.sh
+++ b/stale-packages.sh
@@ -1,0 +1,45 @@
+#! /usr/bin/env bash
+
+# This script finds packages whose repositories have been
+# transferred to a new owner but whose corresponding
+# URL in the registry hasn't been updated to reflect this change.
+
+set -euo pipefail
+
+function repository_status {
+  local repository_url; repository_url="$1"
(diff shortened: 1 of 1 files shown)


thomashoneyman:
This is a good thing to have. However, as discussed in #471, a more complete solution would be to identify any package that has transferred repos since being listed in the new-packages or bower-packages files, and then process a [`Transfer`](https://github.com/purescript/registry/blob/8408a993aa93ff2e9403fe42f0306793d520b7a6/ci/src/Registry/Schema.purs#L197) operation for it. The [transfer operation has been implemented](https://github.com/purescript/registry/blob/8408a993aa93ff2e9403fe42f0306793d520b7a6/ci/src/Registry/API.purs#L264), so this just consists of constructing the proper payload and calling `runOperation` on it.

I'm going to close this because it's not the implementation we need for the registry, but thank you for sending this over!

JordanMartinez:
Sounds good!

Issue #373, opened 2022-03-24
Test utils for generating manifests


colinwahl:
There are existing tests that need to have a valid manifest hierarchy to check behavior (i.e. the registry index tests use the manifests defined here: https://github.com/purescript/registry/blob/master/ci/test/Support/Manifest.purs). Creating these by hand is quite tedious & error prone, especially getting the dependencies right. In the near future, for testing that we are correctly validating build plans for generating pursuit docs & creating the correct resolutions file, we will be in need once again of a valid manifest hierarchy.

Ideally we would have some way of generating these manifests that depend on each other automatically, either by requesting new manifests one-by-one or by generating a whole tree of manifests at a time. I'm not exactly sure what it'd look like or how it'd work at the moment, but here are a few important points I can think of:

* We may need to be able multiple versions of a single package (this is used in the registry index tests)
* Dependency lists should only refer to items that we've already generated a valid manifest for

@thomashoneyman Suggested that we could have a predefined set of "dependency" manifests, and then our generation logic could just use a subset of those as dependencies, only generating the other information. This may work, but could also lead to a pretty flat hierarchy. If that's a problem, we'll need to do a bit more thinking.

thomashoneyman:
Is this fixed in #378?

colinwahl:
Aha - yes, I forgot to link the issue.

PR #378, opened 2022-04-05
Add Manifest fixtures to test suite

colinwahl:
This PR adds fixtures for the `Manifest` type, making it easier to create a `Manifest` for testing.

In particular, this is meant to replace the painful experience of hand-writing a `Manifest` yourself for testing. See the deleted `Test.Support.Manifest` to see how bad it used to be :)

Packaged with the manifest are some utilities for setting the name/dependencies/version/description. These can be used to get fine grained control over just the fields from the manifest that you'd like to change - see the new implementation in `Test.Registry.Index` for how it could be used in action.

Alongside those utilities is a utility that take a depth argument and produces a perfect binary tree of manifests, where the `Manifest` at index 0 is the root of the tree, and all transitive dependencies are contained within the tree. As part of uploading to pursuit we'd like to verify that the build plan provided by the user fulfills all transitive dependencies of a package, and this utility will be useful for generating fake manifests to test said procedure.

I also had to fix one bug in `Registry.Schema` - we had forgotten to `encode` the `files` field, meaning that `Manifest` encoding/decoding did not round trip! Sadly, our existing round trip tests did not include a `Manifest` that had the `files` field filled in, so I replaced them with a test that actually includes all fields so we can ensure they round trip.


ci/test/Fixture/Manifest.purs
```diff
@@ -0,0 +1,133 @@
...
+  fixture = unsafeFromRight $ Version.parseVersion Version.Strict "1.0.0"
+
+instance Fixture License where
+  fixture = unsafeFromRight $ SPDX.parse "MIT"
+
+instance Fixture Owner where
+  fixture = Owner
+    { email: fixture
+    , keytype: fixture
+    , public: fixture
```
thomashoneyman: Can we put real (dummy) values in here, such as:

```suggestion
    { email: "user@example.com"
    , keytype: "ssh-ed25519"
    , public: "AAAAC3NzaC1lZDI1NTE5AAAAIF6B6R9yp8yFM3wYcOhGO0EyZAefKWkvsTET+KeaegY/"
    }
```

ci/test/Fixture/Manifest.purs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+instance Fixture Location where
+  fixture = GitHub { owner: fixture, repo: fixture, subdir: Nothing }
```
thomashoneyman, line 43: Same here for some kind of dummy values, like `"owner"` and `"repo"`

ci/test/Fixture/Manifest.purs
thomashoneyman: The idea being that the fixtures are by default _valid_, even if they're dummy values.

thomashoneyman (commented):
Other than putting in real default values for the fixtures, I'm quite happy with this! Thanks!

ci/test/Fixture/Manifest.purs
colinwahl: Yes - that makes sense!

f-f (approved):
Looks great!

colinwahl:
This intends to fix #373 - forgot to link the issue!

PR #380, opened 2022-04-19
Add package sets conventions to SPEC

thomashoneyman:


SPEC.md
```diff
@@ -110,11 +110,82 @@ This section is meant to be reasonably self-contained, so someone writing a new
...
+
+```json
+{
+  "version": "v2.3.1",
+  "date": "2022-04-19",
+  "compiler": "v0.14.9",
+  "packages": {
+    "aff": {
+      "version": "v5.0.0",
+      "hash": "sha256-fP5dDqsc3/rjZ+WTKFrfIGzuwFnmSQLQg3o+NC17jpY="
```
thomashoneyman: @cdepillabout Does the package set file need to include the package hash directly, or is it possible for tools like `purescript2nix` to look up the hash in the package metadata file (which is stored in this repository under the `metadata` directory)? It would be a little nicer to not include this information in the package set file itself.

SPEC.md
```diff
@@ -110,11 +110,82 @@ This section is meant to be reasonably self-contained, so someone writing a new
...
+3. `date`, which is a date string in YYYY-MM-DD format that describes which day this package set was produced.
+4. `packages`, which is an object in which keys are package names and values are are an object `{ version, hash }`. The version in this object is a SemVer string representing the package version and the hash is an SRI-encoded string that hashes the package contents at that version.
+
+For example, in JSON:
+
+```json
+{
+  "version": "v2.3.1",
+  "date": "2022-04-19",
+  "compiler": "v0.14.9",
```
thomashoneyman: I opted to bring this information into the package set file itself. We already have the info (it's in the package set name), but if any tooling were to want to use that information it would be better for it to be inside the file and easily parsed via JSON rather than forcing them to parse the file name string.

SPEC.md
```diff
@@ -110,11 +110,82 @@ This section is meant to be reasonably self-contained, so someone writing a new
...
+    }
+  }
+}
+```
+
+#### Package Set Naming Conventions
+
+Package sets use the following naming convention:
+
+`MAJOR.MINOR.PATCH+YYYY-MM-DD-purs-MAJOR.MINOR.PATCH`
```
thomashoneyman, line 150: I added the hyphens to YYYY-MM-DD for readability. Are y'all OK with that @f-f @colinwahl?

SPEC.md
thomashoneyman: For context, this is the information that would be stored for each package:
https://github.com/purescript/registry/blob/0c9d437a570f8e3a0a8febf20eae1325637ae99d/v1/Metadata.dhall#L19-L28

SPEC.md
```diff
@@ -110,11 +110,82 @@ This section is meant to be reasonably self-contained, so someone writing a new
...
+The registry provides a curated package set that lists packages known to build together with a particular compiler version. Package managers such as Spago use this package set to decide what versions of your dependencies to install.
+
+#### Package Set Files
+
+Package sets are stored in the `registry` repository under the `package-sets` directory. Each package set is stored as both a Dhall file and a JSON file. A package set is an object or record comprised of the following fields:
+
+1. `version`, which is a string containing the SemVer package set version for this set.
+2. `compiler`, which is a string containing the SemVer compiler version used to verify the package set.
+3. `date`, which is a date string in YYYY-MM-DD format that describes which day this package set was produced.
+4. `packages`, which is an object in which keys are package names and values are are an object `{ version, hash }`. The version in this object is a SemVer string representing the package version and the hash is an SRI-encoded string that hashes the package contents at that version.
```
thomashoneyman: If we need to include the hash in the package set file to help support Nix workflows, then ideally we'd make it optional so that it's still easy to make your own package sets.

SPEC.md
colinwahl: It is fine with me! I also find this more readable. I think I suggested this on last week's call but forgot to type it down in the issue.

JordanMartinez: I think the hyphens help.

JordanMartinez:
I'm not familiar with the details, but from the little I do know, this looks good.

SPEC.md
colinwahl: Is it worth mentioning that the first `MAJOR.MINOR.PATCH` is the package set version, whereas the `purs-MAJOR.MINOR.PATCH` is the version of the compiler the package set was verified with? 

In particular those two versions are not intended to match.

thomashoneyman: Yeah, I think it’s worth mentioning. I do that later but maybe I should bring it up closer to these names

SPEC.md
cdepillabout: > Does the package set file need to include the package hash directly, or is it possible for tools like purescript2nix to look up the hash in the package metadata file (which is stored in this repository under the metadata directory)?

From the perspective of Nix, every file that is downloaded and used during the build process needs to be hashed somewhere.

So, for instance, `purescript2nix` would need one of the following cases to be true:

- The package set file contains hashes of each of the individual source tarballs.  The package set file has a dhall hash in the user's `packages.dhall` file, which was probably added by `spago`.  (It sounds like this is explicitly what you want to avoid...)

- The package set file does not contain hashes of each package, but it _does_ contain a hash of the metadata file.  The metadata file contains hashes for each of the individual source tarballs.  The package set file has a dhall hash in the user's `packages.dhall` file, which was probably added by `spago`.

    In this case, `purescript2nix` would first download the package set file using the hash in the user's `packages.dhall`, then download the metadata file using the hash from the package set file.  It could then download all the individual source tarballs using the hashes in the metadata file.

- The package set file does not contain hashes of each package.  It also does not contain a hash of the metadata file.  However, `spago` adds a URL to the metadata file and a hash that Nix can use to the user's `packages.dhall` file.

    In this case, `purescript2nix` would be able to download both the package set file and the metadata file based on the URLs and hashes in the user's `packages.dhall` file.

The main thing that wouldn't work for building with Nix is if there is no "hashed path" to get to the metadata data file from the user's `packages.dhall` file in their own repo.  Nix doesn't really have any good way of using non-hashed files as build inputs.

(The above is assuming that `spago` still uses the `spago.dhall` and `packages.dhall` files.  My understanding is that there may be some changes with `spago` to support the new registry stuff, so `packages.dhall` may not be the correct file name anymore.  I haven't really followed what is going on with this.)

cdepillabout: > which is stored in this repository under the metadata directory

Oh, I guess another possibility is that the user's `packages.dhall` contains a URL and dhall hash of the entire PureScript Repository.  The PureScript Repository contains both the package set file and the metadata file.  This would also work for Nix/`purescript2nix`.

SPEC.md
f-f: Does it still make sense to have all of this information in the package set name given that we have it inside the package set file too? Once we have that info there then the only reason to have it in the name would be aesthetic, which I wouldn't consider worth when weighted against e.g. the ease to type in a much shorter name

SPEC.md
f-f: I think it's important to note that we wouldn't want the hashes to be in the package set files because they are meant to be edited by hand, so bumping hashes en-masse manually would not be practical.

Since this proposal includes generating a JSON version of the package set we could take that further and have a "master package set file" that is practical to edit by hand, and then generate other files (that could contain the hashes we need) from there. Would this make sense?

SPEC.md
thomashoneyman: At least in the case of Spago, as an end user you usually just have the URL to use to determine what package set you’re on. So the advantage is that at a glance you can tell what version, date, and compiler that set is on, without having to go look at the set’s contents.

The long names are more to type, that’s for sure. I think it makes sense to switch to short versions if we anticipate it changing often. In my spago use, though, I almost always use automatic upgrades so it wouldn’t save me any effort to use short names. But maybe others do this manually more often than I do.

All-in-all I vote to keep the long names in place but I’m open to the short names.

SPEC.md
thomashoneyman: Yea, one difficult part is how the hashes should work in the case of package set overrides. The metadata files always contain the hashes, and even with overrides so long as the set points to registry dependencies that means the hashes are accessible to tooling without any manual work on the user’s part:

But if to use the metadata files they must be contained in a hash themselves it’s less clear how to make sure this works with Nix even when you have a package set with overrides.

thomashoneyman:
@f-f, @colinwahl: in f2fcddf I added the process for adding a batch of new package versions to the package sets automatically, as discussed on the registry call last week. Please comment there to let me know if you see any issues with how I've written it. Perhaps we can workshop a little in this week's call.

Along the way, I had a few questions that I'd like to clarify with each of you. We can break these out to independent issues, quote-reply in this discussion, or chat through them on the registry call — let me know what's most convenient.

---

**Do We Respect Version Bounds?**
What are our specific criteria wher defining that packages "work" together? In the package sets we only require that they compile together. However, package manifests specify version bounds for dependencies. Do we respect these version bounds when producing a new "valid" package set? Or do we only look at whether they compile together?

**Keeping Track of Package Versions More Recent Than the Set**
Packages will sometimes be denied entry into a particular package set because they would have caused another package to break (for example, introducing Halogen 7 might break Halogen Hooks). Then, later, other packages might update to depend on the new version of that package (Halogen Hooks gets updated for Halogen 7), and we are able to upgrade multiple packages together to produce a new set manually.

How should we keep track of packages that have released newer versions than we have allowed into the package set, so that we have a way to gauge when we should do a manual intervention? Do we keep a running open issue, such as we do in the `package-sets` repo? Do we keep a script around we can run at any time to check?

When we agree on a process for doing manual package set interventions when versions are falling behind, I can create a new issue to track developmen.

**Excluding Some Package Versions Immediately**
We can exclude some package versions from the next package set without having to check whether the package compiles with the set. For example, if a package specifies any dependencies that aren't already in the set, we can immediately exclude the package version. Also, if a package version is uploaded that is lower than its version in the existing set, then it can be excluded (we won't attempt to replace `aff@7.0.0` in the package sets with say an `aff@6.5.5` security release).

If we are in agreement about these two (and any other) checks, then I can draft a new issue to track notifying users immediately that their package won't be added to the package set in these specific circumstances.

**Storing Upload Logs in Metadata**
A package is uploaded to the registry right away, but it won't be considered for inclusion in a package set until the batch is entered at once. The new batch is determined based on the metadata files, which means that we don't have a pointer to the issue the user opened to register their package version, and therefore can't notify them of their package's inclusion or exclusion from the set.

We need a way to look up the issue associated with a particular package upload.@colinwahl and I suggest that we include a new field in the `metadata` schema that stores the logs for the package upload. For now it would store the link to the GitHub issue (which is long-lived, unlike the GitHub action log). If we eventually move to our own server, the metadata fields can point at wherever we store logs there.

SPEC.md
thomashoneyman: > I think it's important to note that we wouldn't want the hashes to be in the package set files because they are meant to be edited by hand, so bumping hashes en-masse manually would not be practical.

Would this be less of an issue if the hash field is optional? The hash field would be included in the master package set (which is generated by the registry, largely automatically), but local package set overrides do not have to include it unless you want to. This would make local overrides non-Nix-compatible, and you can add the hash if you want to be Nix-compatible.

It might only be feasible if there were package manager commands or a dedicated CLI for managing your package set, like:

```purs
spago override aff 7.0.0
```

might edit the package set or manifest file to update the Aff version, automatically pulling the correct hash from the registry, since that can be looked up easily.

thomashoneyman:
Some decisions made in the registry call this morning:

> Do We Respect Version Bounds?

No, package sets do not respect version bounds. A package set is considered valid if all included packages compile. We're making this decision because, historically speaking, when there has been a disagreement between version bounds and a compiling package set it invariably is because the version bounds are wrong.

> Keeping Track of Package Versions More Recent Than the Set

We are going to implement a script, `registry-version-check`, which can be used to read in the metadata directory and latest package set and determine which packages have been published and are at versions more recent than those in the package set. It can reuse the logic used to determine what packages should be included in the day's batch in automatic package sets.

> Excluding Some Package Versions Immediately

We will exclude packages from the batch being considered for the day's package set, without attempting to compile them, if they fail any of the following criteria:

* There must not be a more recent SemVer version of the package already in the package set
* All dependencies must be present in the package set (if multiple packages are published together, they will be ordered so that dependents wait on their dependencies to be added)

> Storing Upload Logs in Metadata

We are going to add a new field, `publishLogs`, to the package metadata. This field will link to the logs for the package upload. For example, so long as we are using GitHub Actions this field will contain a link to the issue used to upload the package. If we switch over to another provider or our own server, then we will link to wherever logs are stored there instead.

Not all packages will have logs. Some packages are uploaded without a corresponding log (for example, packages uploaded via the registry importer). Those packages will just have an empty string for their logs link. cc: @f-f — are you OK with this?

thomashoneyman:
@cdepillabout @ddellacosta @f-f 

On the registry call this morning I suggested a way to support Nix tooling as easily as possible in the PureScript registry. Since the registry is not a package manager, its scope only includes packages, metadata about packages, and the package sets. I'd love to hear each of your thoughts on the solution I've outlined below — whether to poke holes in it, suggest alternatives, or sign off.

**tl;dr**
 I propose that we take two steps.

First, we add a new field to the package sets JSON file called `metadataHash`. This field hashes the contents of the registry metadata directory at the time the package set was published. This allows Nix tools to look up hashes for all remote build inputs, including individual packages, from a single hashed file (the remote package set).

Second, we add a new (possibly optional) field to the Spago config file, also called `metadataHash`. This is for users who have additions or overrides to the remote package set that are more recent than the set. These users will need to provide Nix tools with a metadata hash more recent than the one listed in the package set. This field is automatically filled in by Spago. Spago config files are already committed to source control, so this represents little additional friction for users.

My full reasoning is below and is meant for folks who may not know Nix or the registry structure very well. Read it if you'd like _all_ the details!

---

We would like it to be easy for Nix tooling to integrate with the PureScript Registry, especially without those tools having to generate additional files for the user to commit (as is done, for example, by `spago2nix`).

The registry has three parts relevant to this problem where we can make changes to accommodate Nix tooling:

1. The registry backend stores tarballs of the source code of packages that have been uploaded to the registry
2. The registry metadata (stored in the `purescript/registry` repo in the `metadata` directory) records a JSON metadata file for each package in the registry. This file follows the [Metadata.dhall](https://github.com/purescript/registry/blob/1208527f0f2ea5921c2509008da02c17b566dd6a/v1/Metadata.dhall)) schema — for example, here's [the metadata file for the `amqp` package](https://github.com/purescript/registry/blob/1208527f0f2ea5921c2509008da02c17b566dd6a/metadata/amqp.json). Note that the metadata file contains a Nix-compatible hash for every version of the package.
3. The package sets (stored in the `purescript/registry` repo in the `sets` directory) record versioned package set JSON files, where each JSON file follows the [format specified in the SPEC](https://github.com/purescript/registry/blob/faaad9e8ba813bfac0005056ae5a22ce526483ad/SPEC.md#package-set-files).

As far as Nix is concerned, each of these resources represents a build input that is not local to the user's project. Every resource that Nix tools are going to download as part of a build is going to have to have a hash so that Nix can verify its integrity.

A contemporary PureScript project will at least rely on the remote package set and the remote packages themselves. That means we at least need a hash of the package set and a hash for each package used by the project.

**Reducing Hash Totals**

Unfortunately, the more hashes we have to provide to Nix tools, the bigger the headache for end users. This is the problem with `spago2nix`, for example: you have to generate hashes for every package and then commit the giant resulting Nix file. We'd like to avoid that.

Fortunately, the metadata directory stores the hashes for every package version ever uploaded. If we have the metadata directory, we have all the package hashes. Instead of having to maintain a hash for 50 different packages, we can just maintain a hash for the snapshot of the metadata we want to look at.

This leaves us with two hashes that Nix tools need to be able to consume: 

1. The remote package set file
2. The metadata directory

**The Metadata Hash**

I would like to go one step further and reduce to a single hash. It _is_ possible to do this, but there are tradeoffs.

Package sets are made up of a few metadata fields and then a list of packages and their associated versions. The package set only contains packages we have metadata for, and they are immutable once published. For this reason, we _have_ a snapshot of the metadata directory at the time the package set was published.

Therefore, it's possible to add a `metadataHash` to the package sets file. This field would specify the hash of the metadata directory at the time of the package set release. This would leave us with a single hash for users / Nix tools to worry about: the hash of the upstream package set file. Spago today already produces this hash, so this is adds zero friction to today's workflows.

However, this is problematic in the presence of package set additions or overrides. For example, it's common to depend on package set `1.0.0`, which specifies `aff@7.0.0`, but override it to use `aff@8.0.0` instead. The registry must have metadata for `aff@8.0.0`, but your package set file specifies a hash from before `aff@8.0.0` was published, so as far as Nix is concerned, `aff@8.0.0` has no metadata.

So it seems that we can't necessarily add the `metadataHash` field to the package sets file and be done with it. Instead, we have to retain the ability for a user to use a metadata hash that differs from the metadata that existed when the package set was published.

As soon as users are making local additions and overrides we move beyond the ability of the registry to be helpful. Instead, we'll need to rely on external tooling, such as package managers. When installing an override or addition, the package manager could hash the metadata directory and store the resulting hash in a field in the package manager config file.

As an example: the Spago config file could include a new field, `metadataHash`, which overrides the `metadataHash` listed in the package set file. We could make this field required, where its initial value is taken directly from the package set file, or we could make it optional — I'm not sure the best option.

thomashoneyman:
Alternately, @cdepillabout had suggested that a package manager could hash the entire contents of the `registry` repository at a particular commit, which does get us back to a single hash even in the presence of overrides and additions (if you make an addition, you also update the repository hash).

In this case, perhaps _this_ hash could be optional? Non-Nix users don't necessarily need to have a hash recording what commit they're using from the registry repository, but they could opt-in.

For example, something like this in a Spago config:

```dhall
{ name = "my-package"
  -- optional field?
, registryHash = "sha256-abcdefg"
, packageSet = "1.0.0+2022-01-01-purs-0.15.0"
, packageSetOverrides = { "aff" = "8.0.0", "halogen-hooks" = "https://github.com/thomashoneyman/purescript-halogen-hooks#1.0.0" }
}
```

f-f:
Yeah, I'm leaning more this last idea of just keeping a hash of the registry repo, and only for Nix users. Though we'd need to store the commit hash too right?

thomashoneyman:
The registry hash _could_ be used to perform integrity checks in Spago, both for the package set and for individual package tarballs. Maybe it’s something users could opt into if they want those checks, Nix or no Nix?

f-f:
This is only valid for package sets that come from the registry, but people could have custom ones (e.g. for alternate backends)

thomashoneyman:
Yea, that's a good point. We have to include a package set hash in the Spago config file as well in the case that the user is pointing to a non-registry package set.

thomashoneyman:
After a conversation with @f-f, we're thinking of enabling integrity checks (ie. verify hashes, both for Nix and non-Nix users) with the below rules:

1. Three kinds of hashes are required. First, the registry hash (which includes the metadata files with package version hashes). Second, the package set hash (necessary because you may be relying on a package set outside the registry). Third, individual package hashes for overrides / additions that are outside the registry.
2. All three hashes can be computed by Spago. Spago can hash the registry repository, the package set JSON file, and the source of the external package (so long as it is in a repository, I suppose).
3. Spago will enable integrity checks by default, e.g. on install / build it will verify that the package set and individual packages match their hashes.
4. Spago can automatically upgrade hashes via a `spago freeze` command. `spago freeze registry` for the registry hash, `spago freeze package-set` for the package set, `spago freeze packages` for overrides / additions, and `spago freeze all` for everything.
5. Users can opt-out of all of this by setting `enableIntegrity = false` in their configuration. Spago will not compute or check hashes for anything.

I believe this gives tools like `purescript2nix` access to all the hashes they need. It also allows users to verify that the package set and packages they are using are exactly what they expect, even if they aren't Nix users. Finally, because this is optional, users can opt-out and skip the ceremony. cc: @cdepillabout and @ddellacosta — does this give a tool like `purescript2nix` everything it needs to Nixify a PureScript project without generating additional files?

There are still some open questions around this. For example:

1. Should we let users opt-in to a particular level of granularity for their integrity checks? For example, maybe I want to freeze the package set, but I don't care about freezing individual packages (I trust the registry to not change them out from under me). Additions and overrides, even to external packages, could then omit the hashes as well.

2. How do we freeze the registry, given that packages can be unpublished at any time (it's not truly immutable)? My suggestion: because this is quite rare, and because unpublishing will cause fetching the package to return a 404, Spago can rely on the 404 error to say that the package has probably been unpublished and you should update your registry hash.

thomashoneyman:
So long as the spec looks good other than the hashes, I'd like to merge this. The hashes are outside the registry (they're in the package manager's domain), and it doesn't look like the decision there is relevant to the package sets format, so no need to hold this up for that discussion. However, we can continue that discussion here or in an new issue.

SPEC.md
```diff
@@ -110,11 +110,109 @@ This section is meant to be reasonably self-contained, so someone writing a new
...
+
+#### Automatic Package Sets Process
+
+The registry attempts to produce a new package set automatically every day, so long as packages have been uploaded that could be added or updated. No packages are ever dropped from a package set automatically; the only time packages are dropped from the package sets are during manual releases.
+
+Every day, the registry CI executes the following steps:
+
+First, we read the contents of the latest package set release and gather all package versions that have been uploaded to the registry since that release. These package versions are the "batch" of packages that we are considering for automatic inclusion to the next package set.
+
+Second, we filter out any packages where, based on their metadata and manifest files alone, we know they can't be added to the package set. This happens for one of three reasons: they have a dependency that isn't in the package sets, or they had multiple releases since the last package set, in which case we only take the highest version, or c) they already have a higher version number released  in the previous package set. In any of these cases, we remove the package from consideration.
```
f-f:
```suggestion
Second, we filter out any packages where, based on their metadata and manifest files alone, we know they can't be added to the package set. This happens for one of three reasons:
1. they have a dependency that isn't in the package sets, or
2. they had multiple releases since the last package set, in which case we only take the highest version, or
3. they already have a higher version number released  in the previous package set. In any of these cases, we remove the package from consideration.
```

f-f (approved):
Only a minor formatting suggestion, otherwise looks great!

ddellacosta:
@thomashoneyman 

> I believe this gives tools like purescript2nix access to all the hashes they need. It also allows users to verify that the package set and packages they are using are exactly what they expect, even if they aren't Nix users. Finally, because this is optional, users can opt-out and skip the ceremony. cc: @cdepillabout and @ddellacosta — does this give a tool like purescript2nix everything it needs to Nixify a PureScript project without generating additional files?

I apologize for being consistently slow to respond. The honest truth is that I probably should never have committed to doing anything substantial with purescript2nix given my personal and work obligations recently. It's just bad luck that @cdepillabout is also pretty slammed lately.

In any case, thank you, @f-f and all the others involved for all your hard work on this. I see no reason that the approach you've settled on shouldn't work well for purescript2nix, but I'm hoping to have the time to spike something out in the next month or so. That said I'm happy you haven't waited to merge this in, I'm glad to know we weren't blocking you. I'll follow up when I have something more substantial to discuss or show.

PR #381, opened 2022-04-20
CI: Add support for exercising multiple compilers

thomashoneyman:
This PR adds all the compilers available in `easy-purescript-nix` to the Nix shell so that the registry can easily call out to them. For now this is done manually, but if I were a better Nixer I'm sure we could write a function that iterates through the pursPkgs and pulls out all the available compilers. I'd like to do that in a followup PR unless it's obvious how to do it now.

This unblocks work on compiling packages for upload to Pursuit.

You can use the new `callCompiler` function to execute the compiler with some given arguments at a particular compiler version.

### DIFF
--- a/ci/spago.dhall
+++ b/ci/spago.dhall
@@ -43,7 +43,6 @@
   , "precise-datetime"
   , "prelude"
   , "profunctor-lenses"
-  , "psci-support"
   , "record"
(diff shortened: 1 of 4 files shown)


f-f (approved):
Looks great!

PR #382, opened 2022-04-21
CI: Verify build plan in runChecks

thomashoneyman:
This PR adds a verification check to ensure that a user's submitted build plan corresponds with their submitted manifest. If all dependencies present in the manifest are not also present in the build plan and at a version within the supplied range, then we'll reject the package. This ensures that the build plan and manifest are not out of sync, and that when we build the package we're using dependencies that are actually listed in the manifest.

This is part of the work towards building packages in CI.

### DIFF
--- a/ci/spago.dhall
+++ b/ci/spago.dhall
@@ -43,7 +43,6 @@
   , "precise-datetime"
   , "prelude"
   , "profunctor-lenses"
-  , "psci-support"
   , "record"
   , "refs"
   , "safe-coerce"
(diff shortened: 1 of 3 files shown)


thomashoneyman:
Once #381 has merged, we can take the next step and actually download the dependencies in the build plan, compile them, and then use the result to push documentation to Pursuit.

f-f (approved):
I got a little puzzled by the usage of `Either` here (because of its semantics, the way its `Functor` instance works, etc) but this looks good overall!

thomashoneyman:
> I got a little puzzled by the usage of `Either` here (because of its semantics, the way its `Functor` instance works, etc) but this looks good overall!

I was going to use a dedicated sum type, but since this code spans only a few lines I didn’t feel it made things much clearer. We can do that, though, if you’d prefer!

f-f:
> We can do that, though, if you’d prefer!

No I think it's alright, let's move on with this :)

PR #385, opened 2022-04-22
CI: Fix git clone and get publish time in API pipeline

thomashoneyman:
While working on the legacy importer, I noticed that cloning and fetching the timestamps for Git repositories wasn't working. First, because we weren't checking the ref time in the correct directory, and second because the output of `runGit` included a newline after the results of stdout, which we needed to trim off in order to parse the date correctly.

This is now fixed, and executing the legacy import tool successfully reaches the step where we upload to our backend.

### DIFF
--- a/ci/spago.dhall
+++ b/ci/spago.dhall
@@ -43,7 +43,6 @@
   , "precise-datetime"
   , "prelude"
   , "profunctor-lenses"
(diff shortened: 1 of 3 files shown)


f-f (approved):
Wonderful, thank you!

PR #389, opened 2022-04-30
add purescript-signal

CarstenKoenig:
Hi I just updated signal to PureScript 0.15 and I hope you can reintegrate it again.

Thanks


JordanMartinez:
@CarstenKoenig [`signal`](https://github.com/purescript/registry/blob/master/bower-packages.json#L1159) is already in the registry, so I think this PR should be closed.

CarstenKoenig:
oh sorry - thought I'd need to reapply.

Is there anything else I have to do?

JordanMartinez:
No. If you want the package in the package set, you'll need to submit a PR to that repo--but only after all of its deps are in the package set.

CarstenKoenig:
well I surely have misread that  - sorry again.

I had a quick look and seems it was only commented out so it should be no big deal (let's see if the docker-nix part works for me).

Thanks for the explanation.

PR #394, opened 2022-05-05
Use flakes for development

thomashoneyman:
This PR migrates the repository to use flakes for local development. For the typical user, this just comes down to switching from:

```console
$ nix-shell
```

to

```console
$ nix develop
```

However, it has some other niceties. Here are a few:

1. There is no more manual updating of hashes for things like `easy-purescript-nix` or `easy-purescript-dhall`. They can now be updated with `nix flake update`.
2. We no longer pin to a non-reducible nixpkgs checkout. Instead, it's locked just like everything else with hashes. It can also be updated with `nix flake update`.
3. We no longer manually add PureScript compilers to the shell; instead, they are automatically pulled from easy-purescript-nix when it is updated.
4. Non-Flakes users can still use this repo by calling `nix-shell` via the compatibility shell.nix file.
5. Helpful scripts are included in the shell (we could in the future move them to [devshell](https://github.com/numtide/devshell) if we wanted them to be more discoverable).
6. The Nix invocation in CI now ~~is cached and~~ reuses our same pinned nixpkgs from our flake, instead of pointing to yet another Nix channel we have to remember to update.

**Edit**: Well, can't restore the Nix store from cache, so it's not cached. I experimented with some ways to cache this without an external service provider like nixbuild.net or cachix, but the result is just too complicated.


thomashoneyman:
@colinwahl I'd like your review just to make sure this works for you. You're still able to easily get into a dev environment?

thomashoneyman:
Well, while the Nix store cache may not work without further digging, at least I've fixed the PureScript cache. Drops from ~4 minutes to ~3 minutes. But we should still be able to shave another couple minutes off by fixing the Nix caching.

colinwahl:
@thomashoneyman I can confirm I can still use my normal `nix-shell` development workflow on this branch - or did you want me to try out flakes to verify?

thomashoneyman:
Just want to make sure you’re still able to use everything either with flakes or non-flakes

colinwahl:
Okay! In that case I can confirm the non-flakes workflow still works for me, although I suppose I should figure out what flakes are! :laughing:

f-f (approved):
I haven't had the chance to test this locally yet, but it looks lovely! 👏

PR #401, opened 2022-05-16
Fixup and run the legacy importer

thomashoneyman:
This PR makes a few minor tweaks to the legacy import tool so that it runs successfully and also executed it so that we now have all core, web, contrib, node, and deprecated packages (plus a few extra packages that were depended on by those orgs in the past).


ci/src/Registry/API.purs
```diff
@@ -392,7 +393,7 @@ runChecks (BuildPlan buildPlan) metadata (Manifest manifest) = do
     unresolvedDependencies =
       Array.mapMaybe (uncurry dependencyUnresolved) (Map.toUnfoldable manifest.dependencies)
 
-  unless (Array.null unresolvedDependencies) do
+  unless (isLegacyImport || Array.null unresolvedDependencies) do
```
thomashoneyman, line 400: We don't construct build plans for legacy packages, so we shouldn't verify them either.

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -55,12 +55,13 @@ main = Aff.launchAff_ do
     isCorePackage :: Manifest -> Maybe _
     isCorePackage (Manifest manifest) = case manifest.location of
       -- core
-      GitHub { owner: "purescript" } -> Just manifest
+      GitHub { owner: "purescript", repo } | repo /= "purescript-metadata" -> Just manifest
```
thomashoneyman: `metadata` really ought not be in the registry at all, but it's included because it does apparently pass the checks for constructing a valid manifest. However, it then fails to be uploaded because it has no src directory containing purs files.

metadata/accounting.json
```diff
@@ -1,8 +0,0 @@
-{
```
f-f, line 1: Why are these files deleted? I thought we'd keep them around because they are "reserved names" (packages that used to exist, but got lost, and that we don't want people to register)

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -253,7 +253,7 @@ parBounded t f = do
   -- The number of threads here can be tweaked for performance, but push it too
   -- high and you'll run into issues with STDOUT when using CLI tools, which are
   -- used pervasively in processing.
-  for_ (Array.range 1 3) \_ -> do
+  for_ (Array.range 1 1) \_ -> do
```
f-f: Did you intend to commit this? This means that there's no parallelism, so we might as well just remove this chunk of code?

ci/src/Registry/Scripts/LegacyImport.purs
f-f: Rather than adding a special case here we could deal with it in the same way as we deal with the deprecated packages (or well, I thought we were doing that, but then in this PR we remove their metadata files so I'm not sure what's the status for those)

f-f:
I thought we would need to fix rate-limiting issues with GitHub as well to make this run properly (e.g. handle HTTP 409 answers), did those disappear?

ci/src/Registry/Scripts/LegacyImport/Process.purs
thomashoneyman: I should really just remove the parallelism altogether when fetching releases (where we had rate limiting issues), but preserve it elsewhere. I can do that in a little bit.

thomashoneyman:
@f-f In the latest commit I have done a few things to ensure we're getting all the packages correctly:

1. I have removed parallelism, which reduces how frequently rate-limiting kicks in. It's also simpler. Since we're no longer doing active dev on this script it's OK that it's considerably slower now.
2. I explicitly handle the two kinds of request failures I saw (too many requests or a request hanging, never timing out), and retry the requests until they succeed or up to a limit. In my testing I never saw the limit reached, so every package file has been successfully fetched.
3. I have adjusted how we handle the `metadata` package. This is technically a valid package, at least in some versions, but no one is supposed to download it. I've explicitly barred it from the registry, so there's no metadata file and attempting to add the package later will fail a check. This package can't be published. This step is perhaps contentious; I don't know how you'd like to handle this, but most versions of `metadata` can't be uploaded anyway because they don't have a `src` directory.

metadata/accounting.json
thomashoneyman: I should have clarified in the PR. These files are removed because:

1. A previous (flaky) run of the legacy importer mistakenly thought there are no valid versions of this library
2. The revised run of the legacy importer, without errors, can see that there _are_ valid versions of this library.
3. We aren't uploading packages outside core/contrib/web/node, or reserved names, so this package never gets a new metadata file written.

This package should never have had an empty metadata file in the first place.

ci/src/Registry/Scripts/LegacyImport.purs
```diff
@@ -137,7 +149,7 @@ downloadLegacyRegistry = do
   newPackages <- readRegistryFile "new-packages.json"
 
   let
-    allPackages = Map.union bowerPackages newPackages
+    allPackages = Map.delete (RawPackageName "metadata") $ Map.union bowerPackages newPackages
```
thomashoneyman, line 152: We don't process the `metadata` package, and later in the API we don't allow it to be uploaded either. It's a special package with a special meaning, but it contains no usable code and there's no sense in it, for example, being listed on Pursuit. In addition, most versions can't be uploaded anyway because they contain no `src` directory.

ci/src/Registry/Scripts/LegacyImport/Manifest.purs
```diff
@@ -160,18 +164,36 @@ constructManifestFields package version address = do
...
+    url = i "https://raw.githubusercontent.com/" address.owner "/" address.repo "/" tag "/" filePath
+    fileCacheName = String.replace (String.Pattern ".") (String.Replacement "-") filePath
+    cacheKey = i fileCacheName "__" name "__" tag
+    mkError = ResourceError <<< { resource: FileResource resource, error: _ }
+
+    request = Http.defaultRequest
+      { url = url
+      , responseFormat = ResponseFormat.string
+      , timeout = Just (Aff.Milliseconds 5000.0)
+      }
```
thomashoneyman, line 181: This reformulation ensures that requests eventually time out instead of hanging for forever. If they do, we retry them.

ci/src/Registry/Scripts/LegacyImport/Manifest.purs
```diff
@@ -160,18 +164,36 @@ constructManifestFields package version address = do
...
+    go n = Process.withCache serialize cacheKey Nothing do
+      liftAff (Http.request request) >>= case _ of
+        Left AX.TimeoutError -> do
+          if n == limit then do
+            log $ i "[FAILED]: Unable to retrieve " filePath " because the request timed out."
+            throwError $ mkError BadRequest
+          else do
+            let
+              backoff = Math.pow 2.0 (Int.toNumber n + 1.0)
+              delay = backoff * 500.0
```
thomashoneyman, line 192: Requests that error out now have this simple exponential backoff. In my testing, this solved all problems: no request makes it all the way through to a fully failed state. One of the retries always works.

ci/src/Registry/Scripts/LegacyImport/Process.purs
```diff
@@ -236,39 +233,3 @@ withCache { encode, decode } path maybeDuration action = do
 
     false -> do
       onCacheMiss
-
--- | Run a computation over an indexed structure in parallel, using a bounded
--- | queue to avoid issues with conflicts over resources like STDOUT.
---
--- Inspired by the work originally done in:
--- https://github.com/natefaubion/purescript-dodo-printer/blob/540dba0442abe686c0b211868d6f423e5df81b69/test/Snapshot.purs#L54-L61
-parBounded
```
thomashoneyman, line 245: No parallelism means we don't need this anymore. Sayonara!

ci/src/Registry/Scripts/LegacyImport.purs
f-f: If we never have a metadata file (for the `metadata` package, heh) then people could register a different `metadata` package in the future. I think we want to prevent that, so we'd want an empty metadata file for it.

metadata/accounting.json
f-f: I think we are not keeping a list of the "packages with no versions for which we should preserve names", but maybe we should?

ci/src/Registry/Scripts/LegacyImport.purs
thomashoneyman: We can do that, too, but since it’s part of runChecks no one can actually register it. The attempt to register it would fail. Still, I’ll add the empty metadata file and code that makes sure we regenerate one if it isn’t there (if we blow the directory away and regenerate it)

f-f (approved):

PR #402, opened 2022-05-16
Add support for publishing to Pursuit

### LINKED ISSUE #154: Upload docs to Pursuit
Users can manually upload documentation to Pursuit today. But it's easy to forget to upload docs after
publishing a library and the existing process is error-prone. For these reasons, Pursuit documentation
is frequently out of date or missing versions altogether -- even for core libraries!

As part of the new registry pipeline, we should support a step to automatically upload package
documentation to Pursuit. This will help ensure that Pursuit has up-to-date documentation for all versions
of packages in the registry.

However, in order to do that we have to make sure that both:
1. We compile with a compatible compiler version
2. We compile with compatible dependency versions (in other words "we have a working build plan")

We are not there yet because:
1. users cannot specify which compiler versions their package is compatible with
2. we don't have a working solver on the Registry yet.

So our idea for how to approach this is:
- we can start including "publish to Pursuit on the behalf of the user" as an optional step, warning the user if we didn't succeed (while still not failing the pipeline)
- resolve the blockers so that docs publishing always succeeds, by
  - accepting compiler bounds in the `Manifest`
  - implementing a solver that works on the Registry

thomashoneyman:
This PR implements publishing packages to Pursuit after they've been uploaded to the registry and was implemented with @colinwahl, who is essentially the PR co-author. Fixes #154.

Once a package has been added to the registry and storage backend, this optional step kicks in and _only_ for non-legacy packages. We use the provided build plan to download all the package dependencies, and then we compile the package's documentation with the compiler version they asked for. We take the output of calling `purs publish` and POST it over to Pursuit, which completes the publishing process. If there's a failure along the way, we just report it back to the user for now.


ci/test/scripts/PublishPursuit.purs
```diff
@@ -0,0 +1,68 @@
+-- | A test script that exercises publishing to Pursuit only.
```
thomashoneyman, line 1: This script can be used to test out the publishing to Pursuit pipeline, but it expects that the `metadata` directory contains the correct metadata for the package you're trying to publish. In this case you'd need the metadata for `effect@3.0.0` and `prelude@5.0.0`. This can be taken from #401 if you don't already have local copies. For example:

```console
$ spago run -m Registry.PublishPursuit
```

ci/src/Registry/API.purs
```diff
@@ -434,7 +440,136 @@ runChecks (BuildPlan buildPlan) metadata (Manifest manifest) = do
       , incorrectVersionsError
       ]
 
-wget :: String -> String -> RegistryM Unit
+getUnresolvedDependencies :: Manifest -> BuildPlan -> Array (Either (PackageName /\ Range) (PackageName /\ Range /\ Version))
```
thomashoneyman, line 448: This function wasn't changed. We pulled it out so we could write tests for it.

ci/src/Registry/API.purs
```diff
@@ -434,7 +440,136 @@ runChecks (BuildPlan buildPlan) metadata (Manifest manifest) = do
...
+            throwWithComment $ String.joinWith "\n" [ "Failed to parse output of publishing. cc @purescript/packaging", "```" <> err <> "```" ]
+          Right json ->
+            pure json
+
+  authToken <- liftEffect (Node.Process.lookupEnv "GITHUB_TOKEN") >>= case _ of
+    Nothing -> do
+      logShow =<< liftEffect Node.Process.getEnv
+      throwWithComment "Publishing failed because there is no available auth token. cc: @purescript/packaging"
+    Just token ->
+      pure token
```
thomashoneyman: This part is a bit funky. We need a GitHub authentication token to authenticate with Pursuit, and here I'm just pulling the token from the CI build. But we may want to explicitly set a token for Pursuit publishing somewhere, perhaps for a specific GitHub user account (`purescript-registry`?) and use that instead. What are your thoughts, @f-f?

thomashoneyman: I've added a token secret to this repo called `PACCHETTIBOTTI_PURSUIT_TOKEN` which is used by the API now, and there's a corresponding entry in the local .env. The user account used to push to Pursuit is @pacchettibotti.

ci/test/scripts/PublishPursuit.purs
f-f: If I understand correctly this test would still hit Pursuit right? Is that desirable? If yes, then I guess this is not a test that we'd run often?

ci/src/Registry/API.purs
```diff
@@ -2,16 +2,24 @@ module Registry.API where
 
 import Registry.Prelude
 
+import Affjax as AX
 import Affjax as Http
+import Affjax.RequestBody as AXRB
+import Affjax.RequestHeader as AXRH
+import Affjax.ResponseFormat as AXRF
```
f-f: If we're only importing a couple of types from these modules then we could just import them unqualified rather than use these shorthands, and if we're using them more extensively then a more readable name (e.g. `RequestBody` rather than `AXRB`) would be clearer.

Judging by their usage I'd do:
```suggestion
import Affjax.RequestBody as RequestBody
import Affjax.RequestHeader (RequestHeader(..))
import Affjax.ResponseFormat as ResponseFormat
```

ci/src/Registry/API.purs
```diff
@@ -2,16 +2,24 @@ module Registry.API where
 
 import Registry.Prelude
 
+import Affjax as AX
```
f-f: We're already importing this as `Http`
```suggestion
```

ci/test/scripts/PublishPursuit.purs
thomashoneyman: That's right, it hits Pursuit, and it's purely for use in one-off testing. Once this is being exercised by the pipeline regularly we could remove the test.

f-f (approved):

PR #405, opened 2022-05-18
Ban purescript- prefixes by default

thomashoneyman:
This PR adds the `purescript-` prefix as a disallowed package name, though individual packages (such as https://github.com/rightfold/purescript-purescript-compiler-backend-utilities) can be added to an allowlist.

### DIFF
--- a/ci/src/Registry/PackageName.purs
+++ b/ci/src/Registry/PackageName.purs
@@ -51,7 +51,9 @@ parse inputStr = flip Parser.runParser inputStr do
   -- represents the legacy naming scheme and is almost certainly an error.
   -- However, packages can be explicitly blessed so they can use the prefix.
   let
-    allowedPrefixNames = []
+    allowedPrefixNames =
+      [ "purescript-compiler-backend-utilities"
+      ]
     isBlessedPackage = inputStr `Array.elem` allowedPrefixNames
     hasPureScriptPrefix = isJust $ String.stripPrefix (String.Pattern "purescript-") inputStr
 
--- a/ci/src/Registry/PackageName.purs
+++ b/ci/src/Registry/PackageName.purs
@@ -82,7 +82,6 @@ parse inputStr = flip Parser.runParser inputStr do
   -- Make sure that we consume all the string in input
   Parse.eof <?> charErr
 
-  -- Put together the string, stripping the "purescript-" prefix if there
   let
     chars = List.concat $ map NEL.toList $ List.Cons firstChunk nextChunks
     name = fromCharArray $ List.toUnfoldable $ chars
--- a/ci/src/Registry/PackageName.purs
+++ b/ci/src/Registry/PackageName.purs
@@ -6,6 +6,7 @@ module Registry.PackageName
 
 import Registry.Prelude
 
+import Data.Array as Array
 import Data.List as List
 import Data.List.NonEmpty as NEL
 import Data.String as String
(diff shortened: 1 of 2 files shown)


thomashoneyman:
@f-f What do you want to do about pinging the packaging team in the case of this specific error message? Do you think we will be able to just look for those failures, or should we get an explicit and specific ping?

One thing we could do is change the parsing error for package names to be a sum type, where each of the error types is enumerated and the current error messages are the `printError` output. That way we can case on the error if it occurs in the publishing pipeline.

ci/src/Registry/PackageName.purs
```diff
@@ -69,7 +85,7 @@ parse = Parser.runParser do
   -- Put together the string, stripping the "purescript-" prefix if there
```
f-f: This comment will need an update

ci/src/Registry/PackageName.purs
```diff
@@ -37,21 +38,36 @@ print :: PackageName -> String
 print (PackageName package) = package
 
 parse :: String -> Either Parser.ParseError PackageName
-parse = Parser.runParser do
+parse inputStr = flip Parser.runParser inputStr do
```
f-f, line 41: Why this change? I have trouble reading the new version, while the previous one is very clear to me

ci/test/Main.purs
```diff
@@ -160,6 +160,8 @@ goodPackageName = do
 
   parseName "a" "a"
   parseName "some-dash" "some-dash"
+  -- A blessed prefixed package
+  parseName "purescript-compiler-backend-utilities" "purescript-compiler-backend-utilities"
```
f-f, line 164: I am not sure if this is intentional, but we're changing the logic here: the current one will just strip the purescript prefix, but the new one will fail if there's any purescript prefix.
Unless we're doing the prefix-stripping somewhere else, this will fail whenever we try to parse anything into a `PackageName`.

thomashoneyman: We do the prefix-stripping elsewhere. This is a good thing because failing in `PackageName.parse` will tell us right away we forgot to strip a package name, versus silently succeeding and manipulating the package name along the way.

For example, we might shuffle code around and forget to verify there is no PureScript prefix when parsing package manifests; our code will process fine and the resulting package name will have no prefix in the registry, but its manifest file will still have the prefix. Instead, it would be better to fail right away so we can remember to update the code and verify the prefix is not present properly.

ci/src/Registry/PackageName.purs
thomashoneyman: It's because we need access to the input string in the body of `parse` now, which we didn't before. I could also do:

```suggestion
parse inputStr = inputStr # Parser.runParser do
```

Is that any clearer to you?

ci/src/Registry/PackageName.purs
thomashoneyman:
```suggestion
```

f-f:
> What do you want to do about pinging the packaging team in the case of this specific error message? Do you think we will be able to just look for those failures, or should we get an explicit and specific ping?

As long as we get the parsing error to emerge (i.e. if we don't swallow it anywhere else) then I don't think we need to special case it. At first we would be looking at all the failures anyways. Though we might still want an explicit tag, since we could start to ignore notifications from the registry repo when not tagged?

ci/test/Main.purs
f-f: Sounds good!

f-f (approved):

thomashoneyman:
I think it's OK as-is, and we can add the explicit tag if we decide we're going to unwatch the repository at some point.

PR #411, opened 2022-05-25
Add support for 'files' key

thomashoneyman:
This PR implements support for a `files` key in the manifest which has the following rules:

* Some files are always included regardless of the `files` key: the `src` directory, `purs.json` file, `LICENSE` file, and so on
* If the `files` key is not present, then all other files are excluded from the tarball
* If the `files` key is present, then the globs provided at that key are included in the tarball

I know we previously discussed globs being a potentially unsafe option. However, I think globs provide a significant benefit in terms of being able to be selective about what files to include in your project vs. directories alone. 

I spent a good chunk of time researching potential security issues with globs, and I believe the implementation here avoids them:

1. It avoids directory traversal attacks [as described in the Node documentation](https://nodejs.org/en/knowledge/file-system/security/introduction) by sanitizing the result of globbing and catching any paths outside the globbed directory.
2. It follows symlinks by using `realpath` to canonicalize paths, which ensures that a path to a symlink can't then lead to a file outside the base directory getting passed through.
3. It isn't possible to match globs in our codebase without this protection in place since it's built in to `match`, so we'll never forget to have this security check in place.
4. Files that we always include or always ignore are handled separately from this key, so there is no way for users to mess with that. Users can only ever add files that we don't explicitly ignore.
5. Since this is inclusive only, using "not" patterns (ie. `"!src/**/*.purs"`) has no effect and poses no risk.

At the moment there is no error thrown if you provide a glob that resolves to a path outside of the base directory but that path doesn't actually exist. Instead, that path is simply omitted. If we would like to throw an error when we know a glob _would_ be outside the base directory if it existed, then we will need to pre-sanitize the globs instead of only sanitizing the output as is done in this PR. In an earlier version I pre-processed globs, but it's considerably simpler to verify the outputs so I removed that code.

### DIFF
--- a/ci/src/Foreign/FastGlob.purs
+++ b/ci/src/Foreign/FastGlob.purs
@@ -1,76 +1,24 @@
 module Foreign.FastGlob
   ( GlobOptions(..)
   , Include(..)
-  , SafePathError
-  , SafePathErrorReason(..)
   , SanitizedPaths
   , match
   , match'
-  , printSafePathError
   ) where
 
 import Registry.Prelude
 
-import Control.Monad.Except (runExceptT)
 import Control.Promise (Promise)
 import Control.Promise as Promise
 import ConvertableOptions (class Defaults)
 import ConvertableOptions as ConvertableOptions
-import Data.Array as Array
 import Data.Compactable (separate)
 import Data.String as String
 import Effect.Aff as Aff
 import Node.FS.Aff as FS
 import Node.Path as Path
 
-type SafePathError = { path :: FilePath, reason :: SafePathErrorReason }
-
-data SafePathErrorReason
-  = DirectoryTraversal
-  | BaseDirectoryMissing
-  | PathMissing
-
-derive instance Eq SafePathErrorReason
-
-instance Show SafePathErrorReason where
-  show = case _ of
-    DirectoryTraversal -> "DirectoryTraversal"
-    BaseDirectoryMissing -> "BaseDirectoryMissing"
-    PathMissing -> "PathMissing"
-
-printSafePathError :: SafePathError -> String
-printSafePathError { reason, path } =
-  Array.fold
-    [ "Sanitizing '"
-    , path
-    , "' failed: "
-    , case reason of
-        DirectoryTraversal -> "file paths cannot be outside the base directory."
-        BaseDirectoryMissing -> "the base directory does not exist."
-        PathMissing -> "the provided path does not exist."
-    ]
-
-type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array SafePathError }
-
-sanitizePaths :: FilePath -> Array FilePath -> Aff SanitizedPaths
-sanitizePaths baseDirectory paths = do
-  sanitized <- traverse (sanitizePath <<< { baseDirectory, path: _ }) paths
-  let { right, left } = separate sanitized
-  pure { succeeded: right, failed: left }
-
-sanitizePath :: { baseDirectory :: FilePath, path :: FilePath } -> Aff (Either SafePathError FilePath)
-sanitizePath { baseDirectory, path } = runExceptT do
-  absoluteRoot <- liftAff (Aff.attempt $ FS.realpath baseDirectory) >>= case _ of
-    Left _ -> throwError { reason: BaseDirectoryMissing, path: baseDirectory }
-    Right canonical -> pure canonical
-
-  absolutePath <- liftAff (Aff.attempt $ FS.realpath $ Path.concat [ absoluteRoot, path ]) >>= case _ of
-    Left _ -> throwError { reason: PathMissing, path }
-    Right canonical -> pure canonical
-
-  case String.indexOf (String.Pattern absoluteRoot) absolutePath of
-    Just 0 -> pure path
-    _ -> throwError { reason: DirectoryTraversal, path }
+type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array FilePath }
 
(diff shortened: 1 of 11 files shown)


ci/src/Foreign/FastGlob.purs
```diff
@@ -1,27 +1,84 @@
...
+    Just 0 -> pure path
+    _ -> throwError { reason: DirectoryTraversal, path }
 
 data Include = FilesAndDirectories | FilesOnly | DirectoriesOnly
 
 derive instance Eq Include
 
 -- https://github.com/mrmlnc/fast-glob#options-3
 type GlobOptions =
-  ( cwd :: Maybe FilePath
```
thomashoneyman, line 22: I discovered that `fast-glob` will crash if you do not provide the `cwd` argument, even though it says that it defaults to `process.cwd()`. But grepping through the codebase we always need to provide this argument, so I've just pulled it out to be an explicit argument and not part of the options record.

ci/test/Main.purs
```diff
@@ -115,6 +123,33 @@ manifestEncoding = do
...
+      { failed } <- FastGlob.match cwd [ "./shell.nix" ]
+      case Array.head failed of
+        Just { reason } | reason == DirectoryTraversal -> pure unit
+        _ -> Assert.fail $ "Expected failure with " <> show DirectoryTraversal
+
+    -- A glob that is technically a directory traversal but which doesn't
+    -- actually match any files won't throw an error since there are no results.
+    -- If we want to throw an error, then we have to pre-sanitize without
+    -- using `realpath` and use `Path.resolve` instead.
+    Spec.it "Traversal to a non-existing file" do
```
colinwahl, line 142: I'm not sure how big of an issue this is, but this decision leaks information about whether or not files exist on the system. 
Does anyone have any insight into whether that is a concern?

ci/src/Foreign/FastGlob.purs
```diff
@@ -1,27 +1,84 @@
...
+    , "' failed: "
+    , case reason of
+        DirectoryTraversal -> "file paths cannot be outside the base directory."
+        BaseDirectoryMissing -> "the base directory does not exist."
+        PathMissing -> "the provided path does not exist."
+    ]
+
+type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array SafePathError }
+
+sanitizePaths :: FilePath -> Array FilePath -> Aff SanitizedPaths
```
colinwahl: `sanitizePaths` is an internal function that always runs after `match`, which calls `fast-glob` under the hood.

Is it possible for the `BaseDirectoryMissing` & `PathMissing` error cases to happen after `fast-glob`? I don't think it is.

In that case, should we remove those cases altogether, just having a `Either String FilePath`, and then simplify error display cases? An initial thought would just be to crash if the `absoluteRoot`/`absolutePath` don't exist.

ci/test/Main.purs
thomashoneyman: That's a good point. It's possible, with this setup, to know whether your directory traversal hit a file (ie. produced an error) or did not (ie. simply didn't match). That's because we are sanitizing the output of `fast-glob` instead of the input.

We can partially address this by checking the input globs to see whether they would be a directory traversal before passing them along to `fast-glob`, and fail them before matching.

However, we still have to sanitize the outputs because we can't be 100% sure that a glob is safe even after checking with something like `resolve` or `realpath`. That's because those functions are meant for paths, not dynamic patterns like `**`, so they won't check all the literal paths that can be turned up by the glob. Worse, in `realpath`'s case it actually looks up the file on the file system, so it will barf on valid glob patterns like `test/**/*.purs` that would match something like `test/Main.purs`. So we'd want to do something with `resolve` as a pre-pass, which isn't too bad.

We need to check the outputs because those are actual file paths, which means we can be _actually_ sure we're safe by using `realpath` on them -- `realpath` will check the filesystem itself and follow symlinks, so any path turned up by fast-glob can be checked just as if a user had provided explicit paths to each file they want.

ci/src/Foreign/FastGlob.purs
thomashoneyman: You're right, it's not possible for `sanitizePath` to fail with either of these errors _if_ it is run after `fast-glob` matches paths. After all, if the base directory or given path don't exist, then fast-glob wouldn't have matched them. I think we could move both these functions to the `where` clause of `match` to guarantee they can't be used elsewhere, and then use `unsafeCrash` if the base directory or matched path don't exist, because those mean either:

a) we provided an incorrect base directory in our code and need to fix it;
b) fast-glob produced a match for a file which doesn't exist, which means a serious internal error in fast-glob

Alternately, we could provide for the case where fast-glob barfed by keeping the `PathMissing` error, but unsafe crash if the base directory is missing. But I like your suggestion best -- just move to an `Either String _` where the error is for directory traversals.

thomashoneyman:
I'd like to hear what @f-f has to say before moving forward on any of these conversations so we're all on the same page.

ci/src/Foreign/FastGlob.purs
f-f: This all sounds good to me!

ci/test/Main.purs
f-f: > Does anyone have any insight into whether that is a concern?

I don't think this is a concern?
This program is going to be run in CI, and all the sensitive information is stored in environment variables and files. I cannot think of a reason why just knowing that a file exists is going to provide an exploit vector.

f-f (approved):
Beautiful :sparkles: 
Gread work!

PR #413, opened 2022-05-26
Import remaining packages

thomashoneyman:
This PR contains the metadata files for all packages in the registry, minus those that fail our checks. They were generated using this branch, which includes the support for the `files` key from #411.

Almost all packages are successful. Only a handful fail, and they have been noted with specific exemptions in the legacy importer. In the PR comments I describe what might be done about them.

I've manually reviewed fifty or so packages and everything seems to check out with the use of the `files` key. Source directories are copied, LICENSE files in all their variations are copied over (such as LICENSE_THIRDPARTY), and so on.

This doesn't actually need #411 to merge, because even if we handle that key differently none of these packages are using it.


f-f:
@thomashoneyman we are squash-merging, so I'd prefer if we could have the commit with the metadata changes (i.e. `646e7093acd17653e85acd06f81f77d8efded807`) go in a separate PR, so that the code changes are not going to be buried in a giant commit.

thomashoneyman:
@f-f What do you want to do about the packages that are over the size limit of 200_000? Should we bump the limit to be higher, given that these are libraries with legitimate source directories of that size?

We could do something like:

- Under 200_000 bytes: you're OK
- Under 2_000_000 bytes: a warning is emitted, for the packaging team to review, but you're OK
- Over 2_000_000 bytes: your package fails

I'm not really sure what the best approach is. After all, it's OK for your library to be big, if it's legitimate source code.

f-f:
@thomashoneyman yeah, having a cutoff (with a warning) at 2MB sounds great!

PR #423, opened 2022-06-13
Write registry index on disk as part of legacy importer script

thomashoneyman:
This minor change to the legacy importer ensures that we write out the registry index to disk. We can then push to the registry-index repository manually.

Typically this would occur as part of the addition / update operation, but we don't currently push to other repositories when doing additions and updates, and the addition / update operation is skipped for packages that have already been uploaded to our registry backend. Since we're not automatically staying in sync, I've made this an extra step at the end of the legacy import process that always writes all packages to the registry index on disk so we can guarantee it's up to date based on the metadata.


thomashoneyman:
The result of running this version and then pushing the resulting directory can be seen here:
https://github.com/purescript/registry-index/pull/1

f-f (approved):

PR #181, opened 2021-06-08
Fix polyform libs names

paluh:

PR #430, opened 2022-06-20
Drop invalid polyform packages

paluh:
As discussed here: #181


thomashoneyman (dismissed):

thomashoneyman:
@paluh So far we have not allowed names to be "unregistered", because this would allow someone to later reuse the name, leading anyone depending on your library to then accidentally install someone else's once they take over the name. I'll defer to @f-f about making an exception in this case so long as you feel quite sure there aren't users of these libraries.

f-f:
[My latest comment](https://github.com/purescript/registry/pull/181#issuecomment-986055023=) on #181 said something like "I think it's alright to delete these names, since no one can use the packages anyways and we haven't imported the packages yet"

However, since then we have imported all the packages and metadatas, populated the index, etc. In other words, we're much closer to launch than we were then, and I'm not sure anymore we should make an exception here

paluh:
Ah, ok :-) Sorry for such a long delay. I just wanted cleanup my own mistakes... but maybe `batteries-*` are not so valuable namespaces and we can leave it where it is :-)

f-f:
No worries, and thank you for coming back to this!

paluh:
Thank __you__ for being so responsive guys to my stupid propositions and thanks a lot for your work on the registry!

PR #432, opened 2022-06-22
Add check to disallow publishing a previously unpublished version

thomashoneyman:
Package authors should not be able to publish a version that is the same as an already-published and then unpublished version.


colinwahl (approved):
Looks good - one question though:

When we unpublish a package, do we remove it from the `published` field, and move it to the `unpublished` field?

Edit: I found where we do it https://github.com/purescript/registry/blob/b0513025e96a795e6396b34d8bbc6b7f729464a1/ci/src/Registry/Schema.purs#L265-L269 - looks good

PR #440, opened 2022-06-28
Fix tarball nondeterminism due to timestamps

### LINKED ISSUE #439: Tarball creation is non-deterministic
Tarball creation in the registry is non-deterministic: packaging the same package version twice can lead to different hashes and file sizes. For example, if you:

1. Delete the metadata entry for a package version, such as `argonaut-codecs@9.1.0`
2. Run the registry importer (which should only pick up this one version, if you've run it recently)
3. Diff the `metadata` directory

You will frequently see that the tarball size has changed by 1-2 bytes and the hash has changed. This is happening because our use of `Tar.create` is including timestamps. At the very least the temporary directory we use to tar the package is going to change every time we run the importer.

It would really be better if we had a stable hash for the contents of the package, so it doesn't matter at what time we hashed the package and it only matters what the bits in the package are. We can do that by omitting timestamps:

https://stackoverflow.com/questions/32997526/how-to-create-a-tar-file-that-omits-timestamps-for-its-contents

thomashoneyman:
Fixes #439 by making tarball creation depend only on the bits in the package itself, not metadata like timestamps. This ensures that if we have ever regenerate packages in the registry (which we currently do often) the hashes won't change, because the file contents haven't changed either. This is also why Nix sets timestamps on everything to the epoch, for example.

### DIFF
--- a/ci/test/Foreign/Tar.purs
+++ b/ci/test/Foreign/Tar.purs
@@ -23,7 +23,7 @@ tar = do
 
     Spec.it "Tarballs are identical if contents are identical" do
       tarball1 <- createTarball
-      Aff.delay (Aff.Milliseconds 3000.0)
+      Aff.delay (Aff.Milliseconds 100.0)
       tarball2 <- createTarball
       tarball1 `Assert.shouldEqual` tarball2
   where
--- a/ci/test/Foreign/Tar.purs
+++ b/ci/test/Foreign/Tar.purs
@@ -23,7 +23,7 @@ tar = do
 
     Spec.it "Tarballs are identical if contents are identical" do
       tarball1 <- createTarball
-      Aff.delay (Aff.Milliseconds 10.0)
+      Aff.delay (Aff.Milliseconds 1000.0)
       tarball2 <- createTarball
       tarball1 `Assert.shouldEqual` tarball2
   where
--- a/ci/test/Foreign/Tar.purs
+++ b/ci/test/Foreign/Tar.purs
@@ -23,7 +23,7 @@ tar = do
 
     Spec.it "Tarballs are identical if contents are identical" do
       tarball1 <- createTarball
-      Aff.delay (Aff.Milliseconds 1000.0)
+      Aff.delay (Aff.Milliseconds 3000.0)
       tarball2 <- createTarball
       tarball1 `Assert.shouldEqual` tarball2
   where
--- a/ci/test/Foreign/Tar.purs
+++ b/ci/test/Foreign/Tar.purs
@@ -1,0 +1,46 @@
+module Test.Foreign.Tar where
+
+import Registry.Prelude
+
+import Effect.Aff as Aff
+import Foreign.Node.FS as FSE
+import Foreign.Tar as Tar
+import Foreign.Tmp as Tmp
+import Node.FS.Aff as FS
+import Node.FS.Aff as FSA
+import Node.FS.Stats as FS.Stats
+import Node.Path as Path
+import Registry.Hash as Hash
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+tar :: Spec.Spec Unit
+tar = do
+  Spec.describe "Tar" do
+    Spec.it "Successfully produces tarball from directory" do
+      { bytes } <- createTarball
+      bytes `Assert.shouldSatisfy` (_ > 0.0)
+
+    Spec.it "Tarballs are identical if contents are identical" do
+      tarball1 <- createTarball
+      Aff.delay (Aff.Milliseconds 10.0)
+      tarball2 <- createTarball
+      tarball1 `Assert.shouldEqual` tarball2
+  where
+  createTarball = do
+    packageTmp <- liftEffect Tmp.mkTmpDir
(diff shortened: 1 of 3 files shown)


ci/test/Foreign/Tar.purs
```diff
@@ -0,0 +1,46 @@
...
+tar :: Spec.Spec Unit
+tar = do
+  Spec.describe "Tar" do
+    Spec.it "Successfully produces tarball from directory" do
+      { bytes } <- createTarball
+      bytes `Assert.shouldSatisfy` (_ > 0.0)
+
+    Spec.it "Tarballs are identical if contents are identical" do
+      tarball1 <- createTarball
+      Aff.delay (Aff.Milliseconds 10.0)
```
colinwahl: I think `tar` might only keep track of timestamps down to second precision - can we bump this delay up to cross that boundary?

https://unix.stackexchange.com/questions/397130/tar-how-to-preserve-timestamps-down-to-more-than-a-second-of-precision/397132#397132

thomashoneyman:
```suggestion
      Aff.delay (Aff.Milliseconds 1000.0)
```

thomashoneyman: I noticed this failure even with a 10.0 millisecond delay (see the failing test in 24584e6) but I'll bump it to a second just to be sure.

ci/test/Foreign/Tar.purs
```diff
@@ -0,0 +1,46 @@
...
+tar :: Spec.Spec Unit
+tar = do
+  Spec.describe "Tar" do
+    Spec.it "Successfully produces tarball from directory" do
+      { bytes } <- createTarball
+      bytes `Assert.shouldSatisfy` (_ > 0.0)
+
+    Spec.it "Tarballs are identical if contents are identical" do
+      tarball1 <- createTarball
+      Aff.delay (Aff.Milliseconds 1000.0)
```
thomashoneyman: I can try with 3 seconds as well:

```suggestion
      Aff.delay (Aff.Milliseconds 3000.0)
```

but if this works I'd like to drop it back down to 10ms, as that also produced a failing test.

colinwahl (approved):
Could the original failing test have been nondeterministic based on whether or not the small (10 or 100 ms) delay crossed a second boundary?

My concern is that if we don't specifically always cross a second boundary in this test, we don't know if we are actually producing a test which would fail without the changes to the `tar` function.

Maybe this isn't very important - other than that looks good.

thomashoneyman:
I'm not totally sure I'm understanding your concern.

From my perspective, we're able to tell that a delay in creating an identical (contents-wise) tarball that is as short as 10ms (and perhaps even shorter) will produce a difference in timestamps and therefore produce a different hash. A delay even longer, such as 3000ms, also produces the effect. The failing test is demonstrated at least in [24584e6](https://github.com/purescript/registry/pull/440/commits/24584e6e024fed62fb81bfafc699119e119a2720). So we know that without the changes to `tar` we get a failing test with a 10ms delay, and with those changes we get a passing test at 10ms, 100ms, 1s, and 3s.

The original 10ms failing test isn't nondeterministic, as far as I can tell; several runs always produces a failing result on my machine. I can't tell immediately by reviewing the `node-tar` repository, but the failure with the short time makes me think it potentially always uses the POSIX behavior for timestamps (or else I wouldn't expect failures under 1s).

Would you mind clarifying what is giving you pause? I'd like to make sure I'm understanding you properly.

colinwahl:
> I'm not totally sure I'm understanding your concern.
> 
> From my perspective, we're able to tell that a delay in creating an identical (contents-wise) tarball that is as short as 10ms (and perhaps even shorter) will produce a difference in timestamps and therefore produce a different hash. A delay even longer, such as 3000ms, also produces the effect. The failing test is demonstrated at least in [24584e6](https://github.com/purescript/registry/pull/440/commits/24584e6e024fed62fb81bfafc699119e119a2720). So we know that without the changes to `tar` we get a failing test with a 10ms delay, and with those changes we get a passing test at 10ms, 100ms, 1s, and 3s.
> 
> The original 10ms failing test isn't nondeterministic, as far as I can tell; several runs always produces a failing result on my machine. I can't tell immediately by reviewing the `node-tar` repository, but the failure with the short time makes me think it potentially always uses the POSIX behavior for timestamps (or else I wouldn't expect failures under 1s).
> 
> Would you mind clarifying what is giving you pause? I'd like to make sure I'm understanding you properly.

My only concern is that repeated runs of the `10ms` test without our changes to the `tar` bindings would be nondeterministic - if tar only kept timestamps down to the seconds place, I'd expect it to sometimes fail but other times work fine. However, I don't know if tar actually behaves like that, and you ran it multiple times, so I think the existing test is fine.

thomashoneyman:
I think it's at least fine because `tar` should have _definitely_ failed at the 3 seconds mark, and it didn't. I get what you mean about subsequent changes possibly giving a false positive because of `tar`'s precision, but on repeated runs I'm not ever seeing a false positive. I'd like to not add a couple-second delay to the tests if we can help it, so I'm going to merge this as-is, but we can always bump that limit if we need to.

colinwahl:
Sounds good!

PR #441, opened 2022-06-28
Trim input strings when reading registry index files

thomashoneyman:
I noticed when working on daily runs of the registry importer that appending a new manifest to the end of an existing registry index entry can cause the whole file to be lost, and only the latest manifest be written. As an example, see:
https://github.com/purescript/registry-index/commit/3c1ae75b329dfb8962ab45ec162f614dbf3d573a

This happens if a file happens to have been written with a trailing newline character. This will cause the JSON parsing of the file to fail on the empty string after it is split by newlines. Technically speaking, if everything is auto-generated then there shouldn't be a trailing newline. In practice, though, we sometimes have to edit these files and normal file conventions on Linux dictate a trailing newline.

This PR fixes the issue by trimming the input string before attempting to parse it as JSON. In addition, I've changed the registry index implementation to always emit a trailing newline, for consistency's sake.

### DIFF
--- a/ci/src/Registry/Index.purs
+++ b/ci/src/Registry/Index.purs
@@ -88,7 +88,7 @@ readPackage directory packageName = do
 
   contentsResult <- try do
     contents <- FS.readTextFile ASCII path
-    pure $ hush $ traverse Json.parseJson $ String.split (Pattern "\n") $ String.trim contents
+    pure $ hush $ traverse Json.parseJson $ String.split (Pattern "\n") $ contents
 
   pure case contentsResult of
     Left _ -> Nothing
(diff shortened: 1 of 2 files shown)


thomashoneyman:
As support for always having a trailing newline (beyond the convention), the Crates registry does the same thing:
https://github.com/rust-lang/crates.io-index/commit/d511f68fa91e266ba7a20b5f37e7a4801423c289#diff-9d57bbe2bb2ed7c8992b823045fa18a6e3a86cc1d83dad96593896937f755d9a

colinwahl (approved):

PR #442, opened 2022-06-29
Improve use of Octokit to reach the GitHub API

thomashoneyman:
While working on scheduled registry importing I noticed a few issues with our use of Octokit to hit the GitHub API. The most important issue is that if a request failed we have no information about it — we don't know the status code, the error message returned by GitHub, or anything. This made it difficult to do retries or cache 404 failures permanently.

This PR makes a few updates to our use of Octokit when making requests:

1. It introduces a new error type, `GitHubAPIError`, which captures a few of the relevant fields that GitHub returns when a request fails. Most importantly, we now receive the status code and error message (but also the rate limit information, in case we want to use it).
2. It introduces two GitHub-suggested plugins. The first provides automatic throttling in the case we bump up against our rate limit. The second provides automatic retries up to 3 times in the case a non-permanent failure occurred. This saves us from having to implement this logic ourselves and (as far as I can tell) doesn't use up as much of our request quota.

Accordingly, use of Octokit for requests now returns in `ExceptT GitHubAPIError Aff a` instead of `Aff a`. While it's technically true that any `Aff` function can throw, I feel that the new type signature more accurately reflects the possible failure mode.

### Next steps

**Conditional Requests**
In the future, we really ought to make use of [conditional requests](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests) when fetching requests, which will allow us to skip rate-limiting headaches altogether. Essentially, we make a request to check if there are any _new_ releases since our last check. If there are, we fetch them and burn a request. If there aren't, we read from our own cache and don't use up a request.

I didn't implement that in this PR because Octokit doesn't give any access to request headers (as far as I can tell) so we'd need to drop down to a lower level and coordinate it with our on-disk cache. That's a bit much for one PR, so I've stopped here.

**Rate Limiting in CI**
GitHub reports that actions can make up to 1,000 requests per hour using the GITHUB_TOKEN. This is obviously not going to work for us to do batch imports, because fetching releases alone requires 1,700 requests (one per package). Without using conditional requests we will blow through our limit without completing a single run.

However, we can make up to 5,000 requests per hour using a personal token such as the PACCHETTIBOTTI_TOKEN. This allows us to use the tool as-is without reaching our limit, and so long as pacchettibotti doesn't take another 3,300 actions within the hour we should be safe.

#### Other notes
I ran the registry importer again and saw a few 404 errors come through, but no other failures. I then re-ran the importer with an expired cache and saw no errors at all: the 404s were skipped, and everything else fetched successfully.

### DIFF
--- a/ci/package-lock.json
+++ b/ci/package-lock.json
@@ -9,6 +9,8 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
+        "@octokit/plugin-retry": "^3.0.9",
+        "@octokit/plugin-throttling": "^3.7.0",
         "@octokit/rest": "^18.12.0",
         "aws-sdk": "^2.1050.0",
         "decimal.js": "^10.3.1",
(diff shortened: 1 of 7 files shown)


f-f (approved):
Looks great! :clap:

Issue #387, opened 2022-04-25
Implement automatic package sets


thomashoneyman:
The new package sets follow an automatic release process detailed in #380. (We can update the link to be to the spec once that merges). We should implement the automatic package sets releases following that spec.

thomashoneyman:
Specifically, the process for making an automatic package set release is here:
https://github.com/purescript/registry/blob/b09f15620be7f565c2cd09bf13d05693aba779e8/SPEC.md#automatic-package-sets-process

PR #443, opened 2022-06-29
Automated Package Sets: Find package set inclusion candidates from recently uploaded packages

colinwahl:
First steps towards #387

This PR introduces a script that will be used to implement automatic package sets. Right now, it does the following:

1. Fetch the registry index (used to get at manifests)
2. Fetch the registry metadata
3. Find all package uploads within the last 24 hours (via referencing the `publishedTime` field in metadata)
4. Fetch the latest package set
5. Preprocess the package uploads found in step (3), via:
  * Checking that the package isn't in the package set
  * If the package is in the package set, making sure that the uploaded version is later than the version found in the package set
  * Checking that each dependency for the upload is in the package set (See note)
6. Reports the preprocessed package upload candidates.

After this PR, we can find mutually dependent package uploads, and then continue with validation as laid out in #380.
 
Note: I added the check that each dependency for an upload is in the package set after some brief conversations with @thomashoneyman. However, as I thought more about it, I think the check is too aggressive. It is possible for a user to upload a package whose dependencies are not in the latest package set, but also to upload those dependencies all in the same day. I think the ideal workflow in that case is to group those uploads together as a batch and attempt to add them all to the package set at once - if it works, then they are all included in the package set, and if it doesn't, the smallest viable subset of them is included. The existing dependencies check doesn't allow us to gracefully handle this situation.

If people agree that the dependencies check is premature at this point, I'll remove it and we can handle it in the later stages.

### DIFF
--- a/ci/src/Registry/Schema.purs
+++ b/ci/src/Registry/Schema.purs
@@ -59,6 +59,43 @@ derive newtype instance Eq Owner
 derive newtype instance Show Owner
 derive newtype instance RegistryJson Owner
 
+newtype PackageSet = PackageSet
+  { compiler :: Version
+  , packages :: Map PackageName Version
+  }
+
+derive instance Newtype PackageSet _
+derive newtype instance Eq PackageSet
+derive newtype instance Show PackageSet
+
+instance RegistryJson PackageSet where
+  encode (PackageSet plan) = Json.encode plan
+  decode = map PackageSet <<< Json.decode
+
+newtype LegacyPackageSet = LegacyPackageSet (Map PackageName LegacyPackageSetEntry)
+
+derive instance Newtype LegacyPackageSet _
+derive newtype instance Eq LegacyPackageSet
+derive newtype instance Show LegacyPackageSet
+
+instance RegistryJson LegacyPackageSet where
+  encode (LegacyPackageSet plan) = Json.encode plan
+  decode = map LegacyPackageSet <<< Json.decode
+
+newtype LegacyPackageSetEntry = LegacyPackageSetEntry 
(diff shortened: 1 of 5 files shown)


thomashoneyman:
I agree wrt the dependency check and I think we can remove it

ci/src/Registry/Schema.purs
```diff
@@ -59,6 +59,43 @@ derive newtype instance Eq Owner
 derive newtype instance Show Owner
 derive newtype instance RegistryJson Owner
 
+newtype PackageSet = PackageSet
+  { compiler :: Version
```
thomashoneyman, line 64: Can we add the other two fields from the spec (the publish date as an RFC3339String and the package set version)?

ci/src/Registry/Schema.purs
```diff
@@ -61,7 +62,9 @@ derive newtype instance RegistryJson Owner
 
 newtype PackageSet = PackageSet
   { compiler :: Version
+  , date :: RFC3339String
   , packages :: Map PackageName Version
+  , version :: Version
```
thomashoneyman, line 67: Admittedly we can’t use the Version type here because we accept _any_ SemVer version but the registry Version type does not allow build metadata (which we do use for registry versions). If we still have the SemVer type we can use it and if not we can just do a string or newtype here.

colinwahl: Actually I think this is just a `MAJOR.MINOR.PATCH` version, and the more complicated version you're referencing is the title of the package set.

thomashoneyman (approved):

PR #444, opened 2022-07-01
Rework GitHub resource requests

thomashoneyman:
This PR changes the way we fetch and cache resources from the GitHub API. Specifically, it ensures that all requests to GitHub for resources goes through Octokit (building on its support for automatic retries and throttling from #442). In addition, it adds support for [conditional requests](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests) to allow us to make requests to the GitHub API and only burn a request if the resource has changed.

I've noticed a significant improvement in reliability: across the board, we're simply not seeing random request failures for various resources anymore. This is a big deal because we're about to start doing daily imports in CI, and that script is going to be relying on the same caching strategy used in this PR. This new setup makes it easier to implement that script.

In service of this goal I've made four other changes:

**GitHub FFI**
I moved the implementations for the various GitHub endpoints out of the FFI. This makes it easier to implement more of these in the future without dropping down to the FFI by always relying on either the `request` or `paginate` helpers. It also allows us to hook into the new caching mechanism directly via the `requestCached` function.

**Move away from filesystem-only caching**
I removed our bespoke cache that always relied on the file system for two reasons. First, because always writing and reading directly to the filesystem when accessing one of the tens of thousands of files in the cache is slow and jittery. Now we have a cache that reads from memory when possible, and only falls back to the file system if there is a cache miss. And second, because we don't need to rely on manual cache expiry because we can use conditional requests to GitHub to determine whether to read from the cache in-memory or to know it's time to expire the cache.

We can also skip caching some things altogether; we use the registry index as a manifest cache, and only cache manifests we had to produce because they have errors or because they are newer than the registry index.

**Update JSON representation for `Either`**
I changed our JSON encoding and decoding for the `Either` type to use a tagged sum representation. Before, it tried to omit the constructors by trying to parse the success case and then fall back to the failure case. However, if a Left could also possibly be parsed as a Right, then it was parsing them that way; for example, caching and then decoding `Json` always succeeded, causing errors to be falsely parsed as a "successful" cache read. The new encoding is unambiguous and matches the Argonaut implementation.

**Revert changes to `Tar.create` in #440**
In #440 we made tarballs deterministic by stripping timestamps. At least I _thought_ that's what we were doing; in fact, now that I've uploaded some files, we made them deterministic by removing all files that have timestamps (oops). The tests didn't catch it because they simply tested that the resulting tarball was more than 0 bytes, but an empty tarball is 29 bytes. Adjusting the test to check that the result is more than 30 bytes shows that the change in #440 is broken. I've reverted that commit.

Luckily, we don't _have_ to have deterministic tarballs. So long as the tarball we hashed is the tarball that the user downloads, it doesn't matter that it's nondeterministic in production. But it would be nice to address this in the future anyway.

I ran the importer like forty times while testing, and I'm confident that these changes are faster (once files are cached) and significantly more reliable. I also made a few little cleanup changes in the legacy importer around type signatures so that it's easier to read.

-----

There is _one_ problem with all this. GitHub doesn't honor their promise to not burn a request if it returns a 304 Not Modified status. This is in direct contrast to their documentation, which is explicit that it does not count against the rate limit to make a conditional request. I expect this will be fixed in the future, but for now we manually expire the requests cache after a couple hours.

The below request should not consume a request, but it does:
```
$ curl -I https://api.github.com/repos/purescript-contrib/purescript-aff/tags -H "If-Modified-Since: Mon, 06 Jun 2022 08:35:18 GMT"
HTTP/2 304 
server: GitHub.com
date: Fri, 01 Jul 2022 00:45:23 GMT
content-type: application/json; charset=utf-8
cache-control: public, max-age=60, s-maxage=60
vary: Accept, Accept-Encoding, Accept, X-Requested-With
etag: W/"b4d93e006374524077b021bc6341230f0b8d1b9973eb10542c293fd0d44844c9"
last-modified: Mon, 06 Jun 2022 08:35:18 GMT
x-github-media-type: github.v3; format=json
link: <https://api.github.com/repositories/31439226/tags?page=2>; rel="next", <https://api.github.com/repositories/31439226/tags?page=3>; rel="last"
access-control-expose-headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset
access-control-allow-origin: *
strict-transport-security: max-age=31536000; includeSubdomains; preload
x-frame-options: deny
x-content-type-options: nosniff
x-xss-protection: 0
referrer-policy: origin-when-cross-origin, strict-origin-when-cross-origin
content-security-policy: default-src 'none'
x-ratelimit-limit: 60
x-ratelimit-remaining: 59
x-ratelimit-reset: 1656639712
x-ratelimit-resource: core
x-ratelimit-used: 1
x-github-request-id: 9014:0D1B:1C01611:35E3F28:62BE43A6
```

So we've got all the infrastructure in place to handle this properly (we get the 304 response and read from cache properly). But because it burns requests still, I've temporarily disabled it.


ci/src/Foreign/GitHub.purs
```diff
@@ -1,32 +1,346 @@
...
+      -- has not changed, and GitHub promises not to consume a request if so.
+      --
+      -- Unfortunately, GitHub is not currently (2022-07-01) honoring this
+      -- promise, so we (temporarily) only retry after N hours have passed. Once
+      -- they start honoring the promise again we can remove the modified time
+      -- guard below.
+      --
+      -- TODO: Remove DateTime.diff when GitHub honors requests again.
+      Right payload
+        | checkGitHub, DateTime.diff now cached.modified >= Duration.Hours 4.0 -> do
```
thomashoneyman, line 326: This line skips checking with GitHub if the entry is less than 4 hours old. We should delete this once GitHub stops burning requests that return a 304.

ci/src/Foreign/GitHub.purs
```diff
@@ -1,32 +1,346 @@
...
+            log $ "CACHE EXPIRED: " <> route
+            let headers' = Object.insert "If-Modified-Since" (toGitHubTime cached.modified) headers
+            result <- Except.runExceptT $ runRequest octokit (Route route) headers' args
+            case result of
+              -- A 304 response means the resource has not changed and we should
+              -- return from cache.
+              Left err | err.statusCode == 304 -> do
+                -- TODO: Remove this line when GitHub starts honoring rate limit
+                -- exclusions for conditional requests again.
+                liftEffect $ Cache.writeJsonEntry route payload cache
```
thomashoneyman, line 336: Same here: we don't need to update the cache modification time if we're using GitHub to determine whether there has been a change via the 304 response. We only have to do this for our time diff check (above), and otherwise shouldn't be touching the cache at all.

ci/src/Foreign/Tar.js
```diff
@@ -1,54 +1,42 @@
...
-          // 'no mtime' and 'portable' aren't enough on their own:
-          // https://github.com/npm/node-tar/issues/176#issuecomment-391904257
-          stat.mtime = null;
-          stat.birthtime = null;
-        },
-        file: archivename,
-      },
-      [foldername]
-    );
+exports.createImpl = function(cwd, foldername, archivename){
```
thomashoneyman: I've reverted this because -- spoiler alert -- this fixed nondeterminism by _filtering out all the files!_ Our tests checked that the resulting tarball was more than 0.0 bytes, but a tarball is 29.0 bytes with nothing in it, so they didn't catch it. Obviously this isn't going to work, because it totally breaks packaging.

The good news: I've spoken with some Nix folks and deterministic tarballs are only a productivity benefit to us; as far as Nix and `fetchurl` are concerned, they don't care. So long as the tarball that was downloaded is the one that was hashed we're good to go.

thomashoneyman:
I've contacted GitHub support wrt the rate limiting issue and will see if they respond:
https://github.community/t/conditional-requests-still-consuming-rate-limit/259478

ci/src/Foreign/GitHub.purs
f-f: Should we open an issue somewhere so that the GitHub folks notice? Maybe there's one already open and someone came up with a workaround.

ci/src/Foreign/Tar.js
f-f: Great!

ci/src/Foreign/GitHub.purs
f-f: Oh, noticed now https://github.com/purescript/registry/pull/444#issuecomment-1173130540

ci/src/Foreign/GitHub.purs
```diff
@@ -1,32 +1,346 @@
...
+  ]
+
+toGitHubTime :: DateTime.DateTime -> String
+toGitHubTime = Formatter.DateTime.format formatRFC1123
+
+-- | A helper function for implementing GET requests to the GitHub API that
+-- | relies on the GitHub API to report whether there is any new data, and falls
+-- | back to the cache if there is not. Should only be used on GET requests; for
+-- | POST requests, use `request`.
+requestCached
```
f-f: Functions with this many arguments can get confusing (even if the arguments have different types). I wonder if it would make sense to pass in a record here?

ci/src/Foreign/GitHub.purs
```diff
@@ -1,32 +1,346 @@
...
+
+derive newtype instance Eq Route
+derive newtype instance Ord Route
+derive newtype instance Json.StringEncodable Route
+
+foreign import requestImpl :: forall args r. EffectFn6 Octokit Route (Object String) (Record args) (Object Json -> r) (Json -> r) (Promise r)
+
+-- | Make a request to the GitHub API. WARNING: This function does not make use
+-- | of the cache. Prefer `requestCache` when possible.
+request :: forall args. Octokit -> Route -> Object String -> Record args -> ExceptT GitHubAPIError Aff Json
```
f-f: Would it make sense to rename these to `uncachedRequest` and `cachedRequest`?

ci/src/Foreign/Tar.js
```diff
@@ -1,54 +1,42 @@
-const tar = require("tar");
+const tar = require("tar")
 
 // get toplevel directory in tar
-exports.getToplevelDirImpl = function (filename) {
-  return function () {
```
f-f: I suspect you simply reverted the previous commit, but together with it go a bunch of formatting changes - I'd like to keep the formatting fixes though

ci/src/Registry/API.purs
```diff
@@ -122,22 +130,22 @@ readOperation eventPath = do
 
 -- TODO: test all the points where the pipeline could throw, to show that we are implementing
 -- all the necessary checks
-runOperation :: Operation -> RegistryM Unit
-runOperation operation = case operation of
+runOperation :: Octokit -> Operation -> RegistryM Unit
```
f-f, line 133: We're passing `octokit` already to the `RegistryM` env, so there should be no need to pass it in here?

ci/src/Registry/API.purs
```diff
@@ -602,6 +611,36 @@ mkEnv octokit packagesMetadata issue =
   , uploadPackage: Upload.upload
   , deletePackage: Upload.delete
   , packagesMetadata
+  , cache
+  , githubToken
+  }
+
+wget :: String -> FilePath -> RegistryM Unit
```
f-f, line 648: These have moved to the `Utils` module, but I see it's deleted in this PR. Accidental or refactoring?

ci/src/Registry/API.purs
```diff
@@ -635,13 +674,19 @@ isPackageVersionInMetadata packageName version metadataMap =
...
 
 checkIndexExists :: FilePath -> Aff Unit
-checkIndexExists dir = do
-  log "Checking if the registry-index is present..."
-  whenM (not <$> FS.exists dir) do
-    error "Didn't find the 'registry-index' repo, cloning..."
-    Except.runExceptT (runGit [ "clone", "https://github.com/purescript/registry-index.git", dir ] Nothing) >>= case _ of
-      Left err -> Aff.throwError $ Aff.error err
-      Right _ -> log "Successfully cloned the 'registry-index' repo"
+checkIndexExists dirPath = do
```
f-f, line 711: There have been changes to this function in #443, do we merge them or is this a refactoring?

ci/src/Registry/API.purs
thomashoneyman: I haven’t stored the octokit instance in the environment, but I can. I kept it out so we don’t have to construct one for tests. That said, it’s no trouble to make it in the test environment.

ci/src/Registry/API.purs
thomashoneyman: Refactoring; I was running into circular references and we’ve so far avoided having a Utils module as they turn into a bit of a grab-bag. However, I can reinstate it as the circular references aren’t a problem anymore.

ci/src/Foreign/GitHub.purs
thomashoneyman: Sure!

ci/src/Foreign/GitHub.purs
thomashoneyman: Sure!

ci/src/Registry/API.purs
thomashoneyman: This builds on top of those changes. In #443 Colin made the directory configurable; in this, I pull the most recent registry index in that directory.

ci/src/Foreign/Tar.js
thomashoneyman: Yea, I just reverted the commit. I can apply the formatting changes still.

thomashoneyman:
I believe I've addressed all the feedback on this PR. Thanks for the review, @f-f!

ci/src/Registry/API.purs
f-f: No it's good - I also like trying to not have `Util` modules.

ci/src/Registry/API.purs
f-f: Great!

f-f (approved):

Issue #445, opened 2022-07-03
Should the registry admit duplicate locations?


thomashoneyman:
I noticed while working on #444 that a number of packages in the registry point to the same repository.

For example, we uploaded several releases for the `profunctors` library by `@joneshf`, but it turns out that this package is just a fork of the `profunctor` library using another name. It doesn't have any releases of its own and is listed in the Bower registry as pointing at the `purescript-profunctor` core library. The result: the `profunctors` library in the registry is just a bunch of copies of the `profunctor` library.

Worse, these duplicate libraries we're registering (such as `profunctors`) aren't even installable by Bower, so I'm not sure why they're included in the Bower registry file:

```console
λ bower install purescript-profunctors       
bower ENOTFOUND     Package purescript-profunctors not found
```

Should we allow multiple packages in the registry to point to the same location? For example, should I be able to register a new package, `affy`, which points to `https://github.com/purescript-contrib/purescript-aff`? Right now we allow this and will package up every `aff` release as an `affy` release.

My feeling is that we don't want new packages to be able to target an already-published package location. We definitely don't want a new package to be registered against a package location that is currently in use; possibly, if a location _was_ used and then the package was transferred, it could become eligible for use again, but I think we should deny that as well.

Here is the list of duplicates in the `new-packages.json` and `bower-packages.json` files. This list represents package names that are all registered to the same package location.

```json
{
  "https://github.com/arthurxavierx/purescript-carpenter.git": [
    "carpenter",
    "react-carpenter",
    "react-cedar"
  ],
  "https://github.com/chrisdotcode/purescript-requests.git": [
    "http-request",
    "requests",
    "responseable"
  ],
  "https://github.com/fehrenbach/purescript-unordered-collections.git": [
    "unordered-collections",
    "unordered-containers"
  ],
  "https://github.com/jasonzoladz/purescript-rxjs.git": [
    "rx-observable",
    "rxjs"
  ],
  "https://github.com/joneshf/purescript-path.git": [
    "path",
    "unix"
  ],
  "https://github.com/lumihq/purescript-react-dnd-basic.git": [
    "react-basic-dnd",
    "react-dnd-basic"
  ],
  "https://github.com/mitchdzugan/purescript-impulse.git": [
    "impulse",
    "my-project"
  ],
  "https://github.com/natefaubion/purescript-heterogeneous.git": [
    "heterogeneous",
    "heterogenous"
  ],
  "https://github.com/nsaunders/purescript-dotenv.git": [
[...]

PR #447, opened 2022-07-04
Add checks to verify package names and locations are unique before publishing or transferring

thomashoneyman:
Completes #445 by adding two checks to the registry pipeline:

1. Package additions now check that the package name being registered hasn't already been registered. This is the only operation that registers a new name so it's the only place we check.
2. Package additions and transfers now check that the package location being registered isn't currently in use by any other packages (the check includes the `subdir` in the comparison, to allow monorepo setups).

A bit astonishing we didn't already have this check in place in the pipeline, but that's because we've been relying on the `registry-verify-unique` script until the pipeline is up and running!


ci/src/Registry/API.purs
```diff
@@ -124,15 +124,20 @@ readOperation eventPath = do
...
     let updateData = { packageName, buildPlan, updateRef: newRef }
     -- if the metadata file already exists then we steer this to be an Update instead
     if metadataExists then
       runOperation $ Update updateData
-    else
+    else do
+      -- If this is an addition, we need to verify the package has not been
+      -- registered before we continue.
+      metadataMap <- readPackagesMetadata
+      verifyPackageNameIsUnique packageName metadataMap
```
f-f: We don't need this check - the existence of the metadata file with a certain name should be enough to prove that a package with that name has been registered already (and in fact in that case we divert to making an update to that existing package).

We could add a comment in there to make this clear?

thomashoneyman: Ah, that’s right. I wonder whether we should check the package location right away if it is an addition and throw if the location is different, indicating the user should either do a transfer (if they’re trying to re-register the package at a new location) or that they’re trying to register a package that already exists?

I worry that if you try to register a package that’s already registered that it’ll turn into an attempted update (that fails) and that will be more confusing than rejecting the addition because the package is already registered.

f-f: Oh yeah, good idea

thomashoneyman: See ce33351 for a better breakdown of this error.

f-f (approved):
This is great!

PR #449, opened 2022-07-06
Rename eviefp packages

thomashoneyman:
The proper upstream account for these packages is `@eviefp`.


thomashoneyman:
Note that we'll need to regenerate these packages (delete them from the registry backend, metadata, and registry index, and then re-run the registry importer).

f-f (approved):

f-f:
In normal registry operations we'd just issue a `Transfer`, right?

thomashoneyman:
That's right. Once we get the daily importer up I'd like to expand it to automatically do repo transfers as well.

Issue #344, opened 2022-02-09
Proposal: Publish packages via a separate repository


thomashoneyman:
In the registry call this morning, @f-f and I discussed how GitHub issues are used for all API operations for the registry. However, GitHub issues are _also_ used to track development on the registry and as a communication channel between registry trustees and registry users in the case something goes wrong.

As a registry trustee and registry developer, though, this is problematic. If I watch the repository so I can stay on top of issues, then I am going to be flooded with irrelevant issues about users uploading their packages. I guarantee I'll be forced to unwatch the repository, and then I'm going to miss important issues.

I am proposing that we set up a separate repository for exercising the API from the repository we use to develop it. Either we transfer the current repository to a new one (`registry-dev`?) and use the `registry` name for users, or we keep the repository where it is and create a new repository for users (`registry-api`?).

We can still push commits from one repository to another, so I don't believe this will restrict anything we're currently doing. It'll also be nice to be able to separate user documentation from maintainer documentation!

thomashoneyman:
Of the options here, my preference would be to run registry development from a repository named `registry-dev` or something similar, leaving `registry` to be the place where the package metadata and package sets are located (ie. the place where package managers and users interact with the registry).

To transition, I believe we could take these steps:

1. Transfer this repository to `purescript/registry-dev`, so that the issues and pull requests are preserved
2. Create a new repository named `purescript/registry`
3. Copy over the relevant artifacts, ie. the `metadata` directory, the `sets` directory, user-facing documentation

I can take care of the transition if we agree this is how we'd like to do it. But for now I'm putting the `blocked` flag on, because we almost certainly don't want to do this until we are close to launching the registry alpha.

f-f:
This plan sounds great!

A few more notes:
- I'd like @pacchettibotti to be the only account that can write to the new `registry` repo, and we wouldn't accept PRs
- we shall remember to update the current publishing instructions to point to `registry-dev` for updating the `new-packages.json` file (does that require patching `pulp`?)

thomashoneyman:
As discussed in the registry call this morning, we do need to patch Pulp to point at the `registry-dev` repository when we migrate, and we should leave a note for old Pulp users in the `registry` repository pointing them over to `registry-dev` too.

PR #450, opened 2022-07-06
Migrate metadata to registry repo

thomashoneyman:
We are gearing up for #344, in which the metadata and package sets directories are published from the registry repo, and development happens in the registry-dev repo. To prepare, I've created the https://github.com/purescript/registry-preview repository. It has a copy of the latest package set (as of today) and the contents of the metadata directory from this repository.

_This_ pull request updates the registry API and legacy importer to point at the registry instead of the local metadata directory, and it removes all the metadata files currently stored here. This is similar to how those tools already pointed at the remote registry index instead of a local directory. I'd recommend filtering to only non-JSON files to make the review easier.

I don't really like the hardcoded paths we're using in the codebase, but making it fully configurable (ie. passed via an environment flag or CLI argument) seemed even less appealing.

Regardless: once this is in, then the repository structure is correct for the daily importer to push metadata commit-by-commit.


ci/src/Registry/API.purs
```diff
@@ -157,7 +157,7 @@ runOperation octokit operation = case operation of
         ]
       -- If this is a brand-new package, then we can allow them to register it
       -- so long as they aren't publishing an existing location under a new name
-      Nothing | locationIsUnique newPackageLocation packagesMetadata -> throwWithComment $ String.joinWith " "
+      Nothing | not (locationIsUnique newPackageLocation packagesMetadata) -> throwWithComment $ String.joinWith " "
```
thomashoneyman, line 166: :woman_facepalming:

ci/src/Registry/Cache.purs
```diff
@@ -103,16 +103,11 @@ useCache = do
...
-      case Map.lookup key cache of
-        Just prevValue | prevValue.value == value -> pure unit
-        _ -> do
-          utcTime <- nowUTC
-          let
-            entry = { modified: utcTime, value }
-            jsonEntry = entry { modified = RFC3339String.fromDateTime entry.modified }
-          FS.writeTextFile UTF8 (entryPath key) (Json.stringifyJson jsonEntry)
-          Ref.modify_ (Map.insert key entry) cacheRef
+      utcTime <- nowUTC
```
thomashoneyman, line 106: This change is because we want to always write a new modified time to disk, at least so long as we're manually expiring entries.

thomashoneyman:
Re-running the registry importer with this PR allows me to push these commits:

https://github.com/purescript/registry-preview/commit/8788408506eebf995cac1015156318eac97ebc05
https://github.com/purescript/registry-index/commit/f5b68d346ed96a7378479718f98473ce1fc2c4e7

...both of which look correct. In the future the daily importer will push these one-by-one via `@pacchettibotti`.

ci/src/Registry/RegistryM.purs
```diff
@@ -17,7 +17,8 @@ import Registry.Schema (Metadata)
 type Env =
   { comment :: String -> Aff Unit
   , closeIssue :: Aff Unit
-  , commitToTrunk :: PackageName -> FilePath -> Aff (Either String Unit)
+  , commitMetadataFile :: PackageName -> FilePath -> Aff (Either String Unit)
+  , commitIndexFile :: PackageName -> FilePath -> Aff (Either String Unit)
```
thomashoneyman, line 21: This is the major change here — the `commitToTrunk` function was really about committing a metadata file, and since `pushToMaster` was built to push to _this_ repository it needs to be replaced with a function that pushes to the remote registry repo. Plus, we're going to need to do the same thing to push to the registry-index remote repository, so I went ahead and added that here too.

`pushToMaster` was totally broken. I'm not convinced that these new functions are going to work perfectly once we put them into a CI action, but I can fix that when we get there. For now this is at least less broken than the current state of things.

ci/src/Registry/API.purs
```diff
@@ -733,8 +733,11 @@ fetchRepo address path = FS.exists path >>= case _ of
       Left err -> Aff.throwError $ Aff.error err
       Right _ -> log $ "Pulled the latest from " <> address.repo
   _ -> do
+    pacchettibotti <- liftEffect do
+      Env.lookupEnv "PACCHETTIBOTTI_TOKEN"
+        >>= maybe (throw "PACCHETTIBOTTI_TOKEN not defined in the environment") pure
     log $ "Didn't find the " <> address.repo <> " repo, cloning..."
-    Except.runExceptT (runGit [ "clone", "https://github.com/" <> address.owner <> "/" <> address.repo <> ".git", path ] Nothing) >>= case _ of
+    Except.runExceptT (runGit [ "clone", "https://pacchettibotti:" <> pacchettibotti <> "@github.com/" <> address.owner <> "/" <> address.repo <> ".git", path ] Nothing) >>= case _ of
```
thomashoneyman: It looks like we can push via the `pacchettibotti` credentials so long as we use that account to clone the repository we'll be pushing to. I've added support for that here. I think we could alternately thread the token through to where we call `git push`, if you'd rather do that.

thomashoneyman: See:
https://github.com/purescript/registry-preview/commit/9e6581b3fc67bfb25636784f62e2c586ad663e84

ci/src/Registry/API.purs
```diff
@@ -390,24 +411,30 @@ addOrUpdate octokit { updateRef, buildPlan, packageName } inputMetadata = do
...
-  -- TODO: commit the registry-index
-  -- TODO: commit the metadata
-  liftAff $ Index.insertManifest registryIndexPath manifest
+  -- After a package has been uploaded we add it to the registry index, we
+  -- upload its documentation to Pursuit, and we can now process it for package
+  -- sets when the next batch goes out.
+
+  -- We write to the registry index if possible. If this fails, the packaging
+  -- team should manually insert the entry.
+  writeInsertIndex registryIndexPath manifest >>= case _ of
```
thomashoneyman: With this change, pacchettibotti also writes to the registry index!
https://github.com/purescript/registry-index/commit/96baa2f40c0a5c7051d69a83b238f63ea6cd6234

ci/src/Registry/API.purs
```diff
@@ -850,22 +847,31 @@ gitGetRefTime ref repoDir = do
     Left err -> Aff.throwError $ Aff.error $ "Failed to get ref time: " <> err
     Right res -> pure res
 
-pushToRegistryIndex :: PackageName -> FilePath -> Aff (Either String Unit)
-pushToRegistryIndex packageName registryIndexDir = Except.runExceptT do
-  runGit_ [ "config", "user.name", "PacchettiBotti" ] (Just registryIndexDir)
-  runGit_ [ "config", "user.email", "<pacchettibotti@purescript.org>" ] (Just registryIndexDir)
+configurePacchettiBotti :: Maybe FilePath -> ExceptT String Aff GitHubToken
+configurePacchettiBotti cwd = do
```
thomashoneyman, line 870: This function is now used to configure the two pushing functions, which have been renamed to:

* `pacchettiBottiPushToRegistryIndex`
* `pacchettiBottiPushToRegistryMetadata`

which I think more clearly indicates that this is PacchettiBotti doing the work, and therefore it's more natural to have to provide the PACCHETTIBOTTI_TOKEN value versus during the clone earlier in the process.

thomashoneyman:
@f-f in [97b5aed](https://github.com/purescript/registry/pull/450/commits/97b5aed3c1a6c7eed928f0c83f5d661bde73ca98), [705c2ad](https://github.com/purescript/registry/pull/450/commits/705c2ad2a0a9d0368776e452f38db6fa735f0379), [6819fdb](https://github.com/purescript/registry/pull/450/commits/6819fdb4442ee80181810dc58c053cc378ce5eb1), and [f75032a](https://github.com/purescript/registry/pull/450/commits/f75032ace1aa41978a6c78f921a5a6313be90a91) I added support for PacchettiBotti to push to both the registry and the registry index as part of the pipeline.

I deleted the existing entries for bolson@0.0.7 to test this out and it is working well! Here's what PacchettiBotti committed:
https://github.com/purescript/registry-preview/commit/ad43d58f88b4086bf84ec37d1b8d2fcbcc6ab9d7
https://github.com/purescript/registry-index/commit/8ff25946fa78e7211adf16862a02d22a7bb99612

This is _now_ finally ready for review. Sorry for throwing these extra commits in there — once I noticed `pushToMaster` was totally broken I wanted to go ahead and fix it. I'm going to push one more commit to re-disable committing in the registry importer but that's all.

ci/src/Registry/API.purs
f-f: :smile:

ci/src/Registry/API.purs
f-f: This is great!

ci/src/Registry/API.purs
```diff
@@ -166,12 +168,12 @@ runOperation octokit operation = case operation of
         addOrUpdate octokit { packageName, buildPlan, updateRef: newRef } (mkNewMetadata newPackageLocation)
 
   Update { packageName, buildPlan, updateRef } -> do
-    metadata <- readMetadata packageName { noMetadata: "No metadata found for your package. Did you mean to create an Addition?" }
+    metadata <- readMetadata registryPath packageName { noMetadata: "No metadata found for your package. Did you mean to create an Addition?" }
```
f-f: This bit got me confused - we are supposed to be running this in the CI of the `registry` repo. That means that we can read metadata from the root of the repo, write back to it, etc, _and then_ clone the repo once again in a subfolder, copy the data over, and commit.
It is a slightly more complex procedure than "clone the repo, do everything in there", but it helps minimizing the chance for races: the window for conflicting pipelines to step on each other's toes becomes seconds (the time to clone, copy and push) vs minutes here (since we clone the repo at the beginning of the pipeline and push it at the end)

In any case, I think we should move the `registryPath` to a field of the `RegistryM`, rather than being piped in as an arg

ci/src/Registry/API.purs
```diff
@@ -672,9 +714,9 @@ mkLocalEnv octokit cache packagesMetadata =
 type MetadataMap = Map PackageName Metadata
 type MetadataRef = Ref MetadataMap
 
-mkMetadataRef :: Aff MetadataRef
-mkMetadataRef = do
-  FS.Extra.ensureDirectory metadataDir
```
f-f, line 677: Why remove this? The `metadata` directory might not exist (e.g. if we nuke all the packages and upload them again)

ci/src/Registry/API.purs
```diff
@@ -707,20 +749,32 @@ locationIsUnique location metadata = do
...
-    _ -> do
-      log "Didn't find the 'registry-index' repo, cloning..."
-      Except.runExceptT (runGit [ "clone", "https://github.com/purescript/registry-index.git", dirPath ] Nothing) >>= case _ of
-        Left err -> Aff.throwError $ Aff.error err
-        Right _ -> log "Successfully cloned the 'registry-index' repo"
+-- | Fetch the latest from the given repository. Will perform a fresh clone if
+-- | a checkout of the repository does not exist at the given path, and will
+-- | pull otherwise.
+fetchRepo :: GitHub.Address -> FilePath -> Aff Unit
+fetchRepo address path = FS.exists path >>= case _ of
```
f-f, line 765: We should checkout the `main`/`master` branch before pulling, as the local repo might be on a different branch

ci/src/Registry/API.purs
```diff
@@ -793,21 +847,48 @@ gitGetRefTime ref repoDir = do
...
-    _ -> Left $ String.trim result.stderr
+  let stdout = String.trim result.stdout
+  let stderr = String.trim result.stderr
+  case result.exit of
+    NodeProcess.Normally 0 -> do
+      unless (String.null stdout) (info stdout)
+      pure $ Right stdout
+    _ -> do
+      unless (String.null stderr) (error stderr)
+      pure $ Left stderr
```
f-f, line 910: I wonder if this could accidentally leak Pacchettibotti's token, since we include it in the url :thinking:

ci/src/Registry/API.purs
thomashoneyman: The chance for races is a real problem. The two approaches I've considered are:

1. Do what you've described, by reading from the `metadata` directory and then only calling a `clone` as late as possible.
2. Clone the registry as soon as we need data from it, and ensure we call a `git pull` before pushing (accomplishing the same goal as delaying the clone).

These options aren't much different, because in CI we have to check out the `registry` repository in order to gain access to the `metadata` directory in the first place; GitHub Actions doesn't check out anything for you by default. That means we would clone the registry (via the `checkout` action), then run the registry importer (which would then clone the registry again). I'm not sure why this an improvement over _not_ cloning the registry and leaving it to the registry-importer to clone and pull as needed.

I want to clarify something related to this comment:

> we are supposed to be running this in the CI of the `registry` repo.

We run the legacy importer in the CI of the registry repo on a daily schedule. But we _also_ need to preserve our ability to run the importer in a checkout of `registry-dev` so we can upload a batch of packages (for example: we change a field in the metadata and have to re-upload the whole registry). We should preserve both use cases.

I don't think we need to worry about the "CI environment" and the "local environment" because we can think of this script as always running from the `registry-dev` repository in the `ci` directory. We have to consider race conditions either way, and this doesn't add any complexity to the registry CI (in fact, it makes it a bit simpler!).

The GitHub Action that kicks off this process is in the `registry` repository, but the action doesn't really care what repository you run a workflow over. The `checkout` action can be used to pull whatever repository we need. For example, [here's the action workflow file in the registry preview](https://github.com/purescript/registry-preview/pull/1), which clones the _development_ repo and then runs the importer. The registry preview repo is never cloned by the action (it's all handled by the importer), which means that your comment about being able to read metadata from the root of the repo isn't true unless we specifically clone it first.

thomashoneyman: > In any case, I think we should move the registryPath to a field of the RegistryM, rather than being piped in as an arg

👍 on that!

ci/src/Registry/API.purs
```diff
@@ -672,9 +714,9 @@ mkLocalEnv octokit cache packagesMetadata =
 type MetadataMap = Map PackageName Metadata
 type MetadataRef = Ref MetadataMap
 
-mkMetadataRef :: Aff MetadataRef
-mkMetadataRef = do
-  FS.Extra.ensureDirectory metadataDir
+mkMetadataRef :: FilePath -> Aff MetadataRef
+mkMetadataRef registryDir = do
+  let metadataDir = Path.concat [ registryDir, "metadata" ]
```
```suggestion
  let metadataDir = Path.concat [ registryDir, "metadata" ]
  FS.Extra.ensureDirectory metadataDir
```

ci/src/Registry/API.purs
thomashoneyman: https://github.com/purescript/registry/pull/450/files/c6174d48f5b9d15816471bcdb18fde0197849762#r918989909

ci/src/Registry/API.purs
thomashoneyman: I haven't seen this happen, but that's a legitimate worry. I don't recall the error output exactly.

What do you think of applying a quick replacement pattern here that replaces any incidence of the token with `<PACCHETTIBOTTI_TOKEN>`?

thomashoneyman:
@f-f I've addressed your feedback by:

- Moving the registry and registry index paths to `RegistryM`
- Ensuring that we only use the 'main' or 'master' branches to push
- Adding a `runGitSilent` for sensitive commands that does not produce error logs
- Always calling `git pull` before pushing

Some of these changes were best served by moving code that deals with paths (like `mkMetadataRef`) to be in `RegistryM`. In turn, that meant shuffling some code around so that `runRegistryM` is called earlier. It might look like a lot of code moved but it's generally just a level of indentation.

ci/src/Registry/API.purs
f-f: Yeah, this could be a good stopgap

f-f (approved):

PR #452, opened 2022-07-10
Legacy importer improvements: generate vs. iterate, better stats and error output

thomashoneyman:
I've grown increasingly frustrated working with the existing legacy importer code for a few reasons:

1. The code sprawls over nearly ten files and has been written in an ad-hoc fashion as we learned what exactly we needed. This has lead to some significant problems. For example, the current stats output reports that we processed 1617 packages, but there are 1650 packages listed in the legacy registry files! The code is messy, with random validation passes bolted on to code that should already have been validated, unsatisfied dependency checks happening separately from the main pipeline, disabled packages being filtered at the last minute...it's hard to tell what exactly is going on in the process, and that's led to some subtle errors.
2. The error code is way too complicated and includes a number of errors we don't even use. This is especially problematic because we surface some of these errors to users in the API pipeline in a really verbose format. The same goes for the stats code.
3. If we need to tweak the behavior of the importer -- for example, to decide whether to generate a new batch of packages or to have `@pacchettibotti` commit things one-by-one -- then you have to manually go to the code and change it. This is error-prone and makes it hard to know, for a given run of the tool, whether it used committed code or manually-changed code.
4. Code from the script has leaked into other parts of the registry codebase. We shouldn't be importing any scripts code into the main registry.

These problems have occurred because the importer grew organically as a script. But it's time we rewrote it now that we have a clear picture of what it needs to accomplish. I've done that in this PR and am quite happy with the new version. It contains all the functionality of the previous version, plus several new improvements and additions, and still comes in at -500 lines of code.

Specifically:

1. Code related to parsing legacy manifests has moved into the main registry source tree. We now have a `Legacy` directory for holding legacy types like legacy manifests or legacy package sets.
2. I've removed all errors we don't use, and I've adjusted those we do use so they are more detailed and, at the same time, much more condensed. We no longer have `bower-exclusions` and `unsatisfied-dependencies` files and write out reserved package names to the console; now, all errors that affect an entire package are in `package-failures.json` and all errors that affect individual versions are in `version-failures.json`.
3. Everything related to the registry importer has been rewritten into a single file, which is still only a few hundred lines long. The reported statistics are more accurate than before.
4. There is now a minimal CLI that lets you decide whether to run in `update` mode (only new releases are fetched, and `@pacchettibotti` uploads them one-by-one; in the future, `@pacchettibotti` can just open issues for them) or `generate` mode, in which case the whole batch is written to disk and you can commit the metadata and registry-index directories yourself.

I also made one other change: when we generate a manifest for a Spago project, we typically create a dependency range from the exact version from the package set, so if the dependency is `aff: 7.1.0` then we generate `aff: >=7.1.0 <7.1.1`. But this just about guarantees that these bounds can't be satisfied for a project using a solver, and it's not even what Spago does when generating ranges in a bowerfile. So I've changed it so that we assume caret ranges: `7.1.0` will produce `>=7.1.0 <8.0.0`, `0.7.0` will produce `>=0.7.0 <0.8.0`, and so on.

**Reviewing**
It is, unfortunately, impossible to produce a reasonable diff for this rewrite. However, I think the code isn't too bad to read through on a per-file basis, and you can look at the stats below to verify that they're almost exactly the same (though the new stats are a bit more accurate).

I've also run the tool several times to verify its output. No packages have ended up being removed. I feel confident that the new output is at least an improvement over the old, and the code is much easier to read and verify what's going on.

Here's the new stats output (the total versions processed is a bit lower because disabled packages don't have any versions processed now; they fail early.)

```console
1650 packages processed:
  1179 fully successful
  323 partially successful
  105 reserved (no usable versions)
  41 fully failed
  ---
  34 Cannot Access Repo
  5 Disabled Package
  1 Invalid Package Name
  1 Invalid Package URL

12998 versions processed:
  11320 successful
  1678 failed
  ---
  652 Invalid Manifest (Invalid Dependencies)
  500 Invalid Manifest (Missing License)
  326 Invalid Tag
  240 Unregistered Dependencies
  10 Cannot Access Legacy Manifest
  10 Invalid Manifest (Invalid License)
  4 Disabled Version
```

Compared to the old error output (this run the same day as the above error output):

```console
Packages: 1617 total (1204 totally succeeded, 303 partially succeeded, 110 totally failed)
Versions: 13048 total (11609 totally succeeded, 0 partially succeeded, 1439 totally failed)
Failures by error:
  manifestError: 1424 versions across 374 packages
    badDependencyVersions: 652 versions across 156 packages
    missingLicense: 500 versions across 150 packages
    badVersion: 326 versions across 146 packages
  resourceError: 0 versions across 34 packages
  noDependencyFiles: 14 versions across 12 packages
  invalidGitHubRepo: 0 versions across 1 packages
Constructing self-contained registry index...
Writing unsatisfied dependencies file...
240 manifest entries with unsatisfied dependencies
141 reserved names
```

And here's an example from the errors file (the parser error messages are terrible, but once we make those better these errors will be improved as well).

```json
  "argonaut": {
    "v0.0.1": {
      "reason": "Legacy manifest could not be parsed.",
      "tag": "InvalidManifest",
      "value": "InvalidDependencies ([{ purescript-parsing: * }, Expected '>='.; pos = 0], [{ purescript-control: * }, Expected '>='.; pos = 0)]"
    },
```

Versus the existing bower-exclusions:

```json
{ "manifestError": {
    "argonaut": {
      "tag": "Right",
      "value": {
        "v0.0.1": {
          "tag": "ManifestImportError",
          "value": [
            {
              "tag": "BadDependencyVersions",
              "value": [
                {
                  "dependency": "control",
                  "failedBounds": "*"
                },
                {
                  "dependency": "parsing",
                  "failedBounds": "*"
                }
              ]
            }
          ]
        },
 
```


ci/src/Registry/API.purs
```diff
@@ -344,22 +344,18 @@ addOrUpdate octokit { updateRef, buildPlan, packageName } inputMetadata = do
...
-      liftError = map (lmap ManifestImportError)
-
-      runManifest =
-        Except.runExceptT <<< Except.mapExceptT (liftAff <<< map (lmap Json.printJson))
-
-      gatherManifest :: ExceptT ImportError Aff Manifest
-      gatherManifest = do
-        manifestFields <- Manifest.constructManifestFields octokit cache (RawVersion updateRef) address
-        Except.mapExceptT liftError $ Manifest.toManifest packageName inputMetadata.location version manifestFields
+    legacyManifest <- liftAff (Legacy.Manifest.fetchLegacyManifest octokit cache address (RawVersion updateRef)) >>= case _ of
```
f-f, line 347: `fetchLegacyManifest` is not in `RegistryM`, but maybe it should? So we could fetch these from the environment, to not have to pipe them through explicitly:
```suggestion
    legacyManifest <- liftAff (Legacy.Manifest.fetchLegacyManifest address (RawVersion updateRef)) >>= case _ of
```

f-f: I just noticed that the `addOrUpdate` fn also takes in an `Octokit` instance, but it should fetch it from the env since it's running in `RegistryM`

ci/src/Registry/Scripts/LegacyImporter.purs
```diff
@@ -0,0 +1,625 @@
...
+  metadataRef <- API.mkMetadataRef API.registryPath
+
+  API.fetchRegistryIndex API.registryIndexPath
+  existingRegistry <- do
+    -- To ensure the metadata and registry index are always in sync, we remove
+    -- any entries from the registry index that don't have accompanying metadata
+    metadata <- liftEffect $ Ref.read metadataRef
+    registry <- Index.readRegistryIndex API.registryIndexPath
+    let hasMetadata package version = API.isPackageVersionInMetadata package version metadata
+    pure $ mapWithIndex (Map.filterKeys <<< hasMetadata) registry
```
f-f: Should we write things back to the registry index here?
I.e. if there are superfluous entries in the index we wouldn't want them there, so we'd want to commit the result of this filtering

ci/src/Registry/Scripts/LegacyImporter.purs
```diff
@@ -0,0 +1,625 @@
...
+-- | be fetched in the first place. Otherwise, it will produce errors for all
+-- | versions that don't produce valid manifests, and manifests for all that do.
+buildLegacyPackageManifests
+  :: GitHub.Octokit
+  -> Cache
+  -> Maybe RegistryIndex
+  -> RawPackageName
+  -> GitHub.PackageURL
+  -> Aff (Either PackageValidationError (Map RawVersion (Either VersionValidationError Manifest)))
+buildLegacyPackageManifests octokit cache mbRegistry rawPackage rawUrl = Except.runExceptT do
```
f-f: I find the abbreviation confusing:

```suggestion
buildLegacyPackageManifests octokit cache maybeRegistry rawPackage rawUrl = Except.runExceptT do
```

ci/src/Registry/Scripts/LegacyImporter.purs
```diff
@@ -0,0 +1,625 @@
...
+    }
+
+validatePackageDisabled :: PackageName -> Either PackageValidationError Unit
+validatePackageDisabled package =
+  case Map.lookup (PackageName.print package) disabledPackages of
+    Nothing -> pure unit
+    Just reason -> Left { error: DisabledPackage, reason }
+  where
+  -- These packages have no usable versions, but we've discovered by running the
+  -- pipeline that they produce at least one manifest. To avoid processing these
```
f-f, line 458: Does this mean "we can create a manifest but they have no sources"? If yes, probably worth clarifying this comment

f-f:
Left some minor comments - otherwise this looks great, well done! :clap:

ci/src/Registry/Scripts/LegacyImporter.purs
thomashoneyman: Good point. We can delete the entries on disk as well.

thomashoneyman:
@f-f Thanks! I'd like to finish resolving your feedback on the upstream PR and then I can resolve your feedback here.

PR #457, opened 2022-07-15
Improve handling of dependencies in importer

thomashoneyman:
The legacy importer previously chose to use Spago dependencies sooner than it should have. In this PR I have updated it to choose the most-likely accurate dependencies from either the Spago or Bower manifest.

Specifically, the importer previously would fetch the Bower and Spago manifests for a package. If it could get a Spago manifest, then it took those dependencies and moved on. However, this can be problematic: if the Spago dependencies are invalid, for example, then we should look at the Bower dependencies to see if they're correct. (For several libraries in the latest package set, like `aff-promise` and `css`, the Bower file is correct and the Spago file is not). I've updated it to validate both before making a selection.

Also, it's possible that either the Spago file or the Bower file has fallen out of date, even if the dependency list is well-formed. To handle this scenario, we now check shared dependencies between the two files. If one file specifies higher upper bounds for their packages (ie. `aff >=6.0.0 <7.0.0` has a higher upper bound than `aff >=5.0.0 <6.0.0`), then we assume that is the correct one to use and we use it. In the case of a tie, we use the Bower dependencies.

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -12,7 +12,6 @@ import Control.Monad.Reader (ask, asks)
 import Data.Argonaut.Parser as Argonaut.Core
 import Data.Argonaut.Parser as JsonParser
 import Data.Array as Array
-import Data.Array.NonEmpty as NonEmptyArray
 import Data.DateTime as DateTime
 import Data.Foldable (traverse_)
 import Data.Generic.Rep as Generic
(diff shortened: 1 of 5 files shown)


thomashoneyman:
I've confirmed that packages like `aff-promise@v4.0.0` now succeed, whereas they were failing before.

ci/src/Registry/API.purs
```diff
@@ -768,7 +764,7 @@ fetchRepo address path = FS.exists path >>= case _ of
           [ "Cannot import using a branch other than 'main'. Got: "
           , branch
           ]
-      runGit_ [ "pull", "--rebase" ] (Just path)
+      runGit_ [ "pull", "--rebase", "--autostash" ] (Just path)
```
f-f, line 767: TIL autostash 😄

f-f (approved):
Great work! 🚀

PR #459, opened 2022-07-15
Add timeouts, retries, and exponential backoff to some requests

thomashoneyman:
I've noticed a few circumstances in which a request will hang (in some cases simply until I exit the process, and in others until we get an SSL connection error). However, killing the process and restarting gets things working smoothly again.

Some common examples include uploading many packages to the registry backend, some of the GitHub requests (other than when we're rate limited), and using `wget` to download a few hundred packages rapidly. This isn't really a problem when running the registry locally, but it's a problem in CI where the job will just fail with no one there to restart it.

In these circumstances it's useful to be able to automatically kill a request after a certain amount of time and retry it. To that end, I've implemented a utility, `withBackoff`, with the following properties:

* You can set a timeout for any `Aff` function.
* When the time is up, you can provide a `shouldCancel` callback to determine whether to cancel the `Aff` or to give it more time. For example, I use `shouldCancel` to skip cancelling a GitHub request if we're currently rate-limited.
* When the action has been cancelled, you can decide whether to retry it via a `shouldRetry` callback.

### DIFF
--- a/ci/src/Foreign/GitHub.purs
+++ b/ci/src/Foreign/GitHub.purs
@@ -45,7 +45,7 @@ import Data.String as String
 import Data.String.Base64 as Base64
 import Data.String.CodeUnits (fromCharArray)
 import Data.Time.Duration as Duration
-import Effect.Exception as Aff
+import Effect.Aff as Aff
 import Effect.Exception as Exception
 import Effect.Now as Now
 import Effect.Uncurried (EffectFn1, EffectFn6, runEffectFn1, runEffectFn6)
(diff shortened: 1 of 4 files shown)


f-f (approved):

ci/src/Foreign/GitHub.purs
```diff
@@ -241,17 +241,33 @@ type UncachedRequestArgs args =
...
+  rateLimitBackoff octokit (Promise.toAffE request)
+
+-- | Apply exponential backoff to requests that hang, but without cancelling
+-- | requests if we have reached our rate limit and have been throttled.
+rateLimitBackoff :: Octokit -> Aff (Either (Object Json) Json) -> Aff (Either GitHubAPIError Json)
+rateLimitBackoff octokit action = do
+  maybeResult <- withBackoff
+    { delay: Aff.Milliseconds 5_000.0
+    , action
+    , shouldCancel: \_ -> Except.runExceptT (getRateLimit octokit) >>= case _ of
```
thomashoneyman, line 263: In the absolute worst case we can be rate-limited for 1 hour (though typically it's more like 25 minutes). That means that this `shouldCancel` callback will be run at most 5 times (5 seconds ^ 5 = ~50 minutes), and making the rate limit request doesn't count against our rate limit (I've verified). Technically, `shouldCancel` could accept a new timeout, but that felt a bit much to require from a user of the function.

PR #460, opened 2022-07-15
Implement automatic batch processing for package sets

thomashoneyman:
This PR introduces `processBatch`, which is a function you can use to install the current package set, build it, and then attempt to integrate a batch of new packages to the package set. It reports which packages could be added, which ones couldn't, and returns a new package set with the additions that is ready to be published.

Notably, the compiler does not report compilation errors to `stderr`, so I needed to add another constructor to our `callCompiler` output to capture that case.

Using the 1.0.0 package set from the registry-preview repository this successfully compiles the set. In addition, with the below manual batch, we produce a new package set:

```purs
   let
      testPackages = Map.fromFoldable
        [ mkEntry "datetime" "6.1.0"
        ]
        where
        mkEntry name version =
          Tuple (unsafeFromRight (PackageName.parse name)) (unsafeFromRight (Version.parseVersion Version.Lenient version))
```

Once this is merged we should be able to use this to bridge the gap between finding packages to upload and actually publishing the new set.

### DIFF
--- a/ci/src/Registry/Scripts/PublishPackageSet.purs
+++ b/ci/src/Registry/Scripts/PublishPackageSet.purs
@@ -6,6 +6,7 @@ import Affjax as Http
 import Control.Monad.Reader (asks)
 import Data.Array as Array
 import Data.Array.NonEmpty as NonEmptyArray
+import Data.Bitraversable (ltraverse)
 import Data.DateTime as DateTime
 import Data.Int as Int
 import Data.Map as Map
@@ -34,7 +35,7 @@ import Registry.Cache as Cache
 import Registry.Json as Json
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
-import Registry.RegistryM (Env, RegistryM, readPackagesMetadata, runRegistryM)
+import Registry.RegistryM (Env, readPackagesMetadata, runRegistryM)
 import Registry.Schema (PackageSet(..))
 import Registry.Version (Version)
 import Registry.Version as Version
@@ -148,7 +149,7 @@ main = Aff.launchAff_ do
     liftEffect $ Console.log "Found the following uploads eligible for inclusion in package set:"
     liftEffect $ Console.log (show uploads)
 
-    buildPackageSet tmpDir (unsafeFromRight packageSetResult) >>= case _ of
+    liftAff $ buildPackageSet tmpDir (unsafeFromRight packageSetResult) >>= case _ of
       Left MissingCompiler -> logShow "Missing compiler"
       Left (CompilationError err) -> log $ "Failed compilation: " <> err
       Left (UnknownError err) -> log $ "Unknown error: " <> err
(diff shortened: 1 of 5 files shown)


ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -142,4 +149,73 @@ main = Aff.launchAff_ do
     liftEffect $ Console.log "Found the following uploads eligible for inclusion in package set:"
     liftEffect $ Console.log (show uploads)
 
+    liftAff $ buildPackageSet tmpDir (unsafeFromRight packageSetResult) >>= case _ of
```
colinwahl: `buildPackageSet` here is already called from within `tmpDir` (since we chdir above) - does passing it here mean that the `ensureDirectory` below is actually creating a `tmpDir` directory within the temporary directory we are already in?

thomashoneyman:
In the latest commit, you can test uploading a batch by replacing the 'uploads' manually with the following:

```purs
   let
      testPackages = Map.fromFoldable
        [ mkEntry "datetime" "6.1.0"
        ]
        where
        mkEntry name version =
          Tuple (unsafeFromRight (PackageName.parse name)) (unsafeFromRight (Version.parseVersion Version.Lenient version))
```

(and throw in any others you feel like)

ci/src/Registry/Scripts/PublishPackageSet.purs
thomashoneyman: That's a good point. I'm not exactly sure what would happen; I've just removed that now so that everything runs with the understanding we're in our working dir.

thomashoneyman:
Note: I'm regenerating the registry, so you may not actually be able to run a batch with the new package set for a few hours. Code review only for now!

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -142,4 +153,188 @@ main = Aff.launchAff_ do
...
+-- | Attempt to add or update a collection of packages in the package set. This
+-- | operation will be rolled back if the addition fails.
+-- |
+-- | NOTE: You must have previously built a package set with
+-- | `buildPackageSet` before running this function.
+tryBatch :: PackageSet -> Map PackageName Version -> Aff (Either CompilerFailure PackageSet)
+tryBatch (PackageSet set) packages = do
+  let backupDir = "output-backup"
+  let outputDir = "output"
+  FSE.copy { from: outputDir, to: backupDir }
```
thomashoneyman: I should probably replace this with something that preserves timestamps, because right now the compiler is not picking this up -- it's recompiling from scratch.

colinwahl (approved):
👏 Awesome!

ci/src/Registry/Scripts/PublishPackageSet.purs
colinwahl: Oh, I didn't think of that - good idea.

colinwahl: It looks like the `copy` implementation we use has a `preserveTimestamps` option, which is defaulted to false: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/copy.md

thomashoneyman:
I'm going to wait until I can test this again, at which point I'll merge if it runs successfully.

thomashoneyman:
Works! The set of new commits are from rebasing.

PR #461, opened 2022-07-15
Automated Package Sets: Compute new package set version

colinwahl:
Part of https://github.com/purescript/registry/issues/387

Adds a utility which computes the new package set version via inspecting the old package set & the validated updates.

The method for deciding how to modify the previous package set version to obtain the new one should match what is in the spec at https://github.com/purescript/registry/blob/master/SPEC.md#automatic-package-sets-process - specifically:

* If the highest SemVer upgrade in the set was a major version, then the package set increments a major version.
* If the highest SemVer upgrade in the set was a minor version, or any new packages were added to the package set, then the package set increments a minor version.
* If the highest SemVer upgrade in the set was a patch version, then the package set increments a patch version.

### DIFF
--- a/ci/src/Registry/Scripts/PublishPackageSet.purs
+++ b/ci/src/Registry/Scripts/PublishPackageSet.purs
@@ -143,3 +143,38 @@ main = Aff.launchAff_ do
     liftEffect $ Console.log (show uploads)
 
     pure unit
+
+-- | Computes new package set version from old package set and version information of successfully added/updated packages.
+-- | Note: this must be called with the old `PackageSet` that has not had updates applied.
+computeVersion :: PackageSet -> Map PackageName Version -> Version
+computeVersion (PackageSet { packages, version: packageSetVersion }) updates =
+  updateVersion packageSetVersion
+  where
+  updateVersion =
+    if major then
+      Version.bumpMajor
+    else if minor then
+      Version.bumpMinor
+    else if patch then
+      Version.bumpPatch
+    else
+      identity
+
+  -- Check for major version bumps for existing packages
+  major :: Boolean
+  major = Map.toUnfoldable updates # Array.any \(Tuple package version) -> fromMaybe false do
+    prevVersion <- Map.lookup package packages
+    pure (version >= Version.bumpMajor prevVersion)
+
+  -- Check for minor version bumps for existing packages or package introductions
+  minor :: Boolean
+  minor = Map.toUnfoldable updates # Array.any \(Tuple package version) -> fromMaybe true do
+    prevVersion <- Map.lookup package packages
+    pure (version >= Version.bumpMinor prevVersion)
+
+  -- Check for patch version bumps for existing packages
+  patch :: Boolean
+  patch = Map.toUnfoldable updates # Array.any \(Tuple package version) -> fromMaybe false do
+    prevVersion <- Map.lookup package packages
+    pure (version >= Version.bumpPatch prevVersion)
+


thomashoneyman (dismissed):

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -143,3 +143,38 @@ main = Aff.launchAff_ do
...
+
+  -- Check for minor version bumps for existing packages or package introductions
+  minor :: Boolean
+  minor = Map.toUnfoldable updates # Array.any \(Tuple package version) -> fromMaybe true do
+    prevVersion <- Map.lookup package packages
+    pure (version >= Version.bumpMinor prevVersion)
+
+  -- Check for patch version bumps for existing packages
+  patch :: Boolean
+  patch = Map.toUnfoldable updates # Array.any \(Tuple package version) -> fromMaybe false do
```
thomashoneyman, line 177: It looks like all of these use the same pattern, just with the choice of comparing on bumpMajor, bumpMinor, or bumpPatch. Could we pull the shared bit into a helper function?

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -143,3 +143,38 @@ main = Aff.launchAff_ do
     liftEffect $ Console.log (show uploads)
 
     pure unit
+
+-- | Computes new package set version from old package set and version information of successfully added/updated packages.
+-- | Note: this must be called with the old `PackageSet` that has not had updates applied.
+computeVersion :: PackageSet -> Map PackageName Version -> Version
+computeVersion (PackageSet { packages, version: packageSetVersion }) updates =
+  updateVersion packageSetVersion
```
thomashoneyman, line 151: We also need to handle the case in which a package has been added to the package set. That corresponds with a minor version bump. We can check by looking up each updated package in the previous package set; if there are any that weren’t there before then we know it was added.

Technically we could add the condition in which a package is dropped from the set, too — that would be a major version change — in anticipation of when we have the package sets API , since we’ll probably reuse this helper function. What do you think of adding that as well? Want to wait, or go ahead?

ci/src/Registry/Scripts/PublishPackageSet.purs
colinwahl: Sure, I can do that.

ci/src/Registry/Scripts/PublishPackageSet.purs
colinwahl: It is a bit subtle, but in the computation of `minor`, I set `fromMaybe true`, so if the map lookup fails (meaning the package isn't in the previous set) then it marks the package set as eligible for a minor version bump.

I'll add an argument for package removals - we can just set to it empty for this and use it in the API later.

thomashoneyman: Oh, sorry, I missed that! :+1:

thomashoneyman: If that makes it feel off to pull things out into a helper function, then feel free to pass on that. I hadn't noticed that difference.

thomashoneyman (approved):

ci/src/Registry/Scripts/PublishPackageSet.purs
colinwahl: Okay, I'll keep them separate for now then.

PR #462, opened 2022-07-16
Improve error output for callCompiler

thomashoneyman:
This PR improves how we report compiler errors in `callCompiler`. Specifically, we pass the `--json-errors` flag, which ensures that error and warning output is written to stdout, including in the case of success. For non-compilation commands this flag has no effect (ie. printing the version is unaffected).

Accordingly, we now parse the compiler errors and then print them later. This now has significantly improved output for reporting to users in the API pipeline or for seeing why a package couldn't be added to the set. For example:

```
Error 1 of 1

  Module: Deku.Core
  File: packages/deku/src/Deku/Core.purs
  Message:

  Cannot import value fixed from module Bolson.Core
  It either does not exist or the module does not export it.

  Error details:
  https://github.com/purescript/documentation/blob/master/errors/UnknownImport.md
```

Instead of the previous error, which was just a bunch of  `[129 of 417] Compiling ...` suddenly terminating.

### DIFF
--- a/ci/test/Main.purs
+++ b/ci/test/Main.purs
@@ -523,7 +523,7 @@ compilerVersions = do
           MissingCompiler ->
             Assert.fail "MissingCompiler"
           CompilationError errs ->
-            Assert.fail ("CompilationError: " <> API.printCompilerErrors errs)
+            Assert.fail ("CompilationError:\n" <> API.printCompilerErrors errs)
           UnknownError err' ->
             Assert.fail ("UnknownError: " <> err')
         Right stdout ->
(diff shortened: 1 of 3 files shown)


ci/test/Main.purs
```diff
@@ -522,8 +522,8 @@ compilerVersions = do
         Left err -> case err of
           MissingCompiler ->
             Assert.fail "MissingCompiler"
-          CompilationError err' ->
-            Assert.fail ("CompilationError: " <> err')
+          CompilationError errs ->
+            Assert.fail ("CompilationError: " <> API.printCompilerErrors errs)
```
colinwahl:
```suggestion
            Assert.fail ("CompilationError:\n" <> API.printCompilerErrors errs)
```

So that there's a newline before the errors.

colinwahl (approved):

PR #463, opened 2022-07-16
Automatic Package Sets: Write package set files to upstream repo

thomashoneyman:
Partially addresses #387 by writing the new package set to disk and then committing that result upstream to the registry repo.

See: https://github.com/purescript/registry-preview/commit/e0f65691f213d9810dedab291c5f724a2c5b1cba

### DIFF
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -57,7 +57,7 @@ import Registry.PackageName as PackageName
 import Registry.PackageUpload as Upload
 import Registry.RegistryM (Env, RegistryM, closeIssue, comment, commitIndexFile, commitMetadataFile, deletePackage, readPackagesMetadata, runRegistryM, throwWithComment, updatePackagesMetadata, uploadPackage)
 import Registry.SSH as SSH
-import Registry.Schema (AuthenticatedData(..), AuthenticatedOperation(..), BuildPlan(..), Location(..), Manifest(..), Metadata, Operation(..), UpdateData, addVersionToMetadata, isVersionInMetadata, mkNewMetadata, unpublishVersionInMetadata)
+import Registry.Schema (AuthenticatedData(..), AuthenticatedOperation(..), BuildPlan(..), Location(..), Manifest(..), Metadata, Operation(..), PackageSet(..), UpdateData, addVersionToMetadata, isVersionInMetadata, mkNewMetadata, unpublishVersionInMetadata)
 import Registry.Types (RawPackageName(..), RawVersion(..))
 import Registry.Version (ParseMode(..), Range, Version)
 import Registry.Version as Version
@@ -672,6 +672,7 @@ mkEnv octokit cache packagesMetadata issue =
       Right _ -> pure unit
   , commitMetadataFile: pacchettiBottiPushToRegistryMetadata
   , commitIndexFile: pacchettiBottiPushToRegistryIndex
+  , commitPackageSetFile: pacchettiBottiPushToRegistryPackageSets
   , uploadPackage: Upload.upload
   , deletePackage: Upload.delete
   , packagesMetadata
@@ -703,6 +704,9 @@ mkLocalEnv octokit cache packagesMetadata =
       pure (Right unit)
   , commitIndexFile: \_ _ -> do
       log "Skipping committing to registry index..."
+      pure (Right unit)
+  , commitPackageSetFile: \_ _ -> do
+      log "Skipping committing to registry package sets..."
       pure (Right unit)
   , uploadPackage: Upload.upload
   , deletePackage: Upload.delete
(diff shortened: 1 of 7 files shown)


ci/src/Registry/RegistryM.purs
```diff
@@ -59,15 +60,20 @@ closeIssue = asks _.closeIssue >>= liftAff
 throwWithComment :: forall a. String -> RegistryM a
 throwWithComment body = comment body *> Aff.throwError (Aff.error body)
 
-commitMetadataFile :: PackageName -> FilePath -> RegistryM (Either String Unit)
-commitMetadataFile packageName path = do
-  f <- asks _.commitMetadataFile
-  liftAff $ f packageName path
+commitMetadataFile :: PackageName -> RegistryM (Either String Unit)
```
thomashoneyman, line 63: For each of these functions, we _know_ we want the provided `FilePath` to be a specific value from the environment. I mistakenly provided the path to the file to commit while working on this, and it's an easy mistake to make. So I've adjusted these to read the path from the environment and provide it directly.

ci/src/Registry/API.purs
```diff
@@ -895,6 +899,16 @@ pacchettiBottiPushToRegistryMetadata packageName registryDir = Except.runExceptT
   let origin = "https://pacchettibotti:" <> token <> "@github.com/purescript/registry-preview.git"
   void $ runGitSilent [ "push", origin, "main" ] (Just registryDir)
 
+pacchettiBottiPushToRegistryPackageSets :: PackageSet -> FilePath -> Aff (Either String Unit)
+pacchettiBottiPushToRegistryPackageSets (PackageSet set) registryDir = Except.runExceptT do
+  GitHubToken token <- configurePacchettiBotti (Just registryDir)
+  runGit_ [ "pull", "--rebase", "--autostash" ] (Just registryDir)
+  runGit_ [ "add", Path.concat [ "package-sets", Version.printVersion set.version <> ".json" ] ] (Just registryDir)
+  let message = "Release " <> Version.printVersion set.version <> " package set."
```
thomashoneyman, line 923: As a followup task, it would be really nice to commit the changes as well. Something like:

```
Release 1.0.1 package set

Additions:
  css 1.0.0

Updates:
  aff 7.0.0 -> 7.1.0

Removals:
  argonaut 8.1.0
```

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -267,12 +267,14 @@ tryBatch (PackageSet set) packages = do
   let backupDir = "output-backup"
   let outputDir = "output"
   FSE.copy { from: outputDir, to: backupDir, preserveTimestamps: true }
+  removePackages (Map.keys packages)
```
thomashoneyman, line 271: I had to add these calls to `removePackage(s)`because if we're processing a _new_ package for the set, then we can't revert to the old set by reinstalling (because the package wasn't in the old set). We need to always explicitly remove packages.

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -150,11 +145,29 @@ main = Aff.launchAff_ do
...
+      let format name version = PackageName.print name <> "@" <> Version.printVersion version
+      log "Found the following uploads eligible for inclusion in package set:"
+      forWithIndex_ uploads (\name version -> log $ format name version)
+      processBatch prevPackageSet uploads >>= case _ of
+        Nothing -> do
+          log "No packages could be added to the set. All packages failed:"
+          forWithIndex_ uploads (\name version -> log $ format name version)
+        Just { success, fail, packageSet } -> do
+          unless (Map.isEmpty fail) do
+            log "Some packages could not be added to the set:"
```
colinwahl: In the case that there are failures and successes, I think this could benefit from more visual distinction between the failures reporting message and the successes message, what do you think?

colinwahl: In particular, maybe a newline before the "New packages were added to the set!" line so that you don't glance right past it.

ci/src/Registry/Scripts/PublishPackageSet.purs
colinwahl: Good catch.

ci/src/Registry/Scripts/PublishPackageSet.purs
```diff
@@ -182,10 +195,19 @@ processBatch prevSet@(PackageSet { compiler, packages }) batch = do
       throwError $ Aff.error $ "Starting package set must compile in order to process a batch."
     Right _ -> pure unit
 
+  let
+    mkNewPackageSet :: PackageSet -> Map PackageName Version -> RegistryM PackageSet
+    mkNewPackageSet (PackageSet new) successes = do
```
colinwahl: Given this name, I thought the function should also update `packages` until I realized that's handled elsewhere (in `tryBatch` / `tryPackage`) - maybe we could call it `updatePackageSetMetadata` or similar instead?

colinwahl (approved):

PR #464, opened 2022-07-16
Legacy Importer: Refer to package set dependencies for packages with Spago files

thomashoneyman:
There are many packages imported into the registry which have issues with their Spago dependencies. If those packages have valid Bower dependencies then we take them, but there are quite a few (~70) package versions where there is no valid Bowerfile -- including some popular packages like `halogen-hooks`. Many of those packages have issues because the Spago file is used for local development only, with the "production" version hosted in the package-sets repository.

Six packages are barred from the registry altogether because of their local Spago files, even though they are in the current package sets (ocarina, rito, envparse, etc). Others have some versions affected (halogen-hooks, etc.).

But there are other issues, too. Some packages have technically valid Spago.dhall files, but if you go to resolve their dependencies you find they're pointing to forks or otherwise can't actually be used (bolson, deku, for example).

Many of these issues can be resolved by the following process (which this PR does):

1. If a package has a spago.dhall file, then attempt to look up the package version in the package sets.
2. If the version is there, then use those dependencies (they've been validated and checked in CI and, for the community, serve as the de facto "real" dependencies of the package).
3. If the version isn't there, then use the dependencies listed in the spago.dhall + packages.dhall for the package.

Of course, if the package has a valid Bower file that's up-to-date, then we use that as the dependency list. Still, implementing the process described above produces improved manifests for a lot of packages.

### DIFF
--- a/ci/src/Registry/Legacy/Manifest.purs
+++ b/ci/src/Registry/Legacy/Manifest.purs
@@ -25,6 +25,7 @@ import Foreign.Tmp as Tmp
 import Node.FS.Aff as FSA
 import Node.Path as Path
 import Registry.Cache as Cache
+import Registry.Hash (sha256String)
 import Registry.Json ((.:), (.:?))
 import Registry.Json as Json
 import Registry.Legacy.PackageSet (LegacyPackageSet(..), LegacyPackageSetEntry(..))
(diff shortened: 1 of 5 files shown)


ci/src/Registry/Legacy/Manifest.purs
```diff
@@ -276,3 +291,81 @@ fetchBowerfile address (RawVersion ref) = do
...
+        -- Package sets are only valid if all packages in the set have dependencies that are also in
+        -- the set, so this (should) be safe.
+        resolveDependencyVersion dep = unsafeFromJust do
+          LegacyPackageSetEntry entry <- Map.lookup dep entries
+          pure entry.version
+        resolveDependencyVersions =
+          Array.foldl (\m name -> Map.insert name (resolveDependencyVersion name) m) Map.empty
+      Map.singleton version (resolveDependencyVersions dependencies)
+
+  legacySets <- do
```
thomashoneyman, line 318: This aggressive caching is important for performance reasons, especially since the legacy package sets are read on every API operation.

thomashoneyman:
I've uploaded the full registry using this PR and it's looking good.

ci/src/Registry/Legacy/Manifest.purs
f-f: Should this be a comment in the code?

thomashoneyman:
In the latest commit, following a brief discussion on Discord, I've updated the code so that we always attempt to use package sets dependencies, and then fall back to the manifests (first Bower, then Spago).

ci/src/Registry/Legacy/Manifest.purs
```diff
@@ -66,22 +64,20 @@ fetchLegacyManifest packageSetsDeps address ref = do
         let printRange version = Array.fold [ ">=", fixed, " <", bump version ]
         RawVersionRange $ Either.either (const fixed) printRange parsedVersion
 
+      validatePackageSetDeps = case packageSetsDeps of
+        Nothing ->
+          Left { error: InvalidDependencies [], reason: "No package sets dependencies." }
```
thomashoneyman, line 69: This needs to return an error compatible with the other validation functions, but it won't ever arise in practice; we always attempt the next manifest if this fails.

ci/src/Registry/Legacy/Manifest.purs
```diff
@@ -300,20 +296,7 @@ fetchLegacyPackageSets = do
...
-          tagVersion <- hush $ Version.parseVersion Version.Lenient trimmed
-          -- The package sets did not have Spago-compatible sets until after
-          -- the 0.12.3 compiler sets. We only refer to the package sets when
-          -- looking at Spago packages, so those are all we need.
-          minimumVersion <- hush $ Version.parseVersion Version.Lenient "0.12.4"
-          guard $ tagVersion >= minimumVersion
-          pure name
-
-      pure $ filterMap pred tags
+    Right tags -> pure $ map _.name tags
```
thomashoneyman, line 299: Since we're now always looking at the package sets, there's no point in filtering to only package sets since Spago was released.

thomashoneyman:
I’ve done _another_ registry upload using this change.

colinwahl (approved):
Code seems to do what you've described - as long as it's been discussed with @f-f , I think it should be good to go.

Issue #300, opened 2022-01-19
List package versions from the Bower registry which will not be in the PureScript registry


thomashoneyman:
A number of package versions that exist in the Bower registry today will not be accepted into the PureScript registry because they have unsolvable issues. Some examples, in no particular order:

* Some package versions have no license files
* Some package versions no longer have reachable GitHub repositories
* Some package versions have malformed Bowerfiles or we otherwise can't produce a manifest for them
* Some package versions do not use the `X.Y.Z` version format we restrict to in the registry (ie. they use prerelease identifiers)

These package versions will be dropped from the registry. This issue tracks ensuring we have a list of these package versions so that users and authors can see what was dropped and for what reason.

f-f:
Is this now solved?

thomashoneyman:
I’m not considering this solved until we actually do the legacy import — until then we can only estimate

thomashoneyman:
These are the removed packages: [removed-packages.json](https://github.com/user-attachments/files/24968751/removed-packages.json). Where should we put this, @f-f?

f-f:
Let's create an `archive` folder in the `registry` repo, where we put that one, and we also move `bower-packages` and `new-packages`. (those two will need a patch here first, since we sync to them when adding a new package)

f-f:
This happened in https://github.com/purescript/registry/pull/533 so this can now be closed 🎉

PR #465, opened 2022-07-18
Commit the package and version failures

thomashoneyman:
Solves #300 by committing the files that list the package and version failures from what is hopefully our final full registry import. These files can be used by people who are curious why particular packages or versions are not in the registry.


thomashoneyman:
Technically, these files can change over time because we run a daily legacy import. One option is to have the importer generate and commit these files as part of the daily import. Another is to remark that these files reflect the mass import, and don't accommodate changes after X date.

Or, we can leave this until we launch the registry alpha, at which point we will be turning off the daily importer and really _will_ have a final set of failures.

Regardless of our decision, I wanted to open this PR so that y'all can take a look at these failures and see if anything looks off. I'm happy to close this PR if we want to wait until the importer is turned off (for example).

PR #469, opened 2022-07-23
Add manual package set batch update API

### LINKED ISSUE #438: API endpoint for manual package set interventions
The new registry includes support for automatically releasing new package sets each day. However, there are several commonplace scenarios in which packages will not automatically be added to the sets, and manual intervention is required.

For example, the automatic sets have a rule: no package version will ever be added to the package sets if it would cause another package to be dropped. This happens every month or so: a widely-used package has a breaking change, and it takes some time before the rest of the packages in the package set update. However, the automatic sets only process packages uploaded in the last 24 hours, so this leads to a situation in which _none_ of these packages can be updated in the sets even if they all _could_ be updated in a batch.

So far we've just waved our hands and said we'll do it manually. But how, exactly, should we do this manual intervention? There are some options — here are a few, though there are likely more:

1. We support manual edits to the `packages.json` file via pull requests from members of the `@purescript/packaging` team or the community. This is how we do things in the `package-sets` repo and it's simple. But it has downsides: our plan has been to disallow pull requests to the `registry` repo altogether, but this would require allowing them. Also, we wanted only the `@pacchettibotti` user to be able to write to the repository, but now others could commit via a pull request.
2. We allow pull requests to some file in the `registry-dev` repository instead. This sidesteps the issue of write access to the registry, but increases friction because the implementation is split between repositories, and community members now have to know about the `registry-dev` repository.
3. We add a new API endpoint to the registry, `UpdatePackageSet`, and manual package set updates are now part of the registry CI along with package additions, updates, transfers, unpublishing, and so on.

This issue is my proposal for the third option, which I think is the best. The advantages are:

* We can continue to disallow pull requests and only allow `@pacchettibotti` to write to the registry, while retaining the ability for packaging team and regular community members to suggest updates to the package sets.
* The mechanism is the same as the one used for other registry operations, ie. a payload in an issue corresponding with an API "endpoint", so we stay consistent. The same  goes for the implementation: it goes in the API file and is triggered the same way as other API operations. We don't have to build anything new.

**`UpdatePackageSet` Payload**

This API needs to support several operations on a package set:

* Adding a new package to the package set
* Updating the version of a package in the package set
* Removing a package from the package set

My proposed payload: the user submits a batch of packages to update together. This batch is represented in JSON with key/value pairs, where keys are package names and values are either versions (representing an addition or update) or `null` (representing removal). Unmentioned packages are unchanged.

```json
{
  "aff": "7.0.0",  // addition or update
  "argonaut": null // removal
}
```

In addition, I think that community members should be able to propose a batch, but only so long as it corresponds with at most a minor version change for the package set. In other words, no removals and no version downgrades. Only the packaging team can take those potentially disruptive actions (and we may want to disallow package downgrades altogether).

**Managing Permissions**

How are we to determine whether someone suggesting a batch is a member of the packaging team?

First, the issue creation / comment event includes the issue creator's / commenter's GitHub username and user ID:
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#issues

The teams endpoint of the GitHub API returns the members of the team you ask for (so long as your account has access to that information, which we do have), including their usernames and user IDs. We can use that to check the members of the packaging team and cross-reference that against the issue creator:
https://docs.github.com/en/rest/teams/members#list-team-members

**Example Workflow**

I open a new issue:

```json
{
  "packageSetBatch": {
    "aff": "8.0.0",
    "argonaut": "8.1.1",
    "halogen-hooks": null
  }
}
```

This triggers the `UpdatePackageSet` API operation to kick off. The operation parses the payload and ensures that it's valid. If the payload is invalid (malformed version, etc.) then a comment is left on the issue indicating the error.

Next, if the suggested batch includes a removal (or downgrade, if we want to support that), then my permissions are checked. If I don't have sufficient permissions, the error is left on the issue and the issue is closed.

Then, the batch is submitted to the automatic package set upgrade process, except that the restriction that packages can't be dropped is lifted. Otherwise, it works the same way: packages are ordered by their dependencies and tried as a batch, then one-by-one. (We may have to adjust this to accommodate the various ways that dropping a package can have effects on the rest of the batch).

If any packages in the batch could not be added, then the error is reported to the user who opened the issue, but the rest of the packages are added to the package set. (Alternately, since these are manual package set update attempts, we could just accept/reject the batch as a whole).

**Potential Issues**

The main issue I see here is that of race conditions: two users submit manual batches at the same time. There's not a good way that I know of to ensure that only one of these operations runs at a time. If this is an issue, then it's possible that we would need to use a different method or push this work off to our own server, where we can implement a queue.

thomashoneyman:
Fixes #438 by implementing an API operation for package set batches. This implementation follows the package sets section of the spec, with some minor tweaks that are included in this PR. I'd recommend reading the manual package set interventions part of the spec before reviewing the code.

The way the new API works is that a packaging team member (an admin) can open an issue with this body:

```jsonc
{ "compiler": "0.15.3", // optional field. version cannot be downgraded. only admins can upgrade compiler versions.
  "packages": { // required field listing the package set updates to make
    "aff": "7.0.0", // adds or updates 'aff'. versions cannot be downgraded.
    "argonaut": null, // removes 'argonaut'. only admins can remove packages.
  }
}
```

However, it will fail if you are not on the packaging team, because this includes restricted actions. A normal community member can also suggest a batch, but they are only allowed to provide a body like this:

```jsonc
// the user can provide the "compiler" field if they wish, but they must set it to the same compiler
// version that is already in use, if so.
{ "packages": {
    "aff": "7.0.0",
    "argonaut": "8.1.0"
  }
}
```

To accomplish the new API I've had to shuffle a few files around:

1. The `PublishPackageSet` script is now `PackageSetUpdater`, and almost all of the package set updating code has been moved out of this module since it is also required by the API.
2. There is now a `Registry.PackageSet` module for updating the package sets. It's largely the exact code that was previously in the `PublishPackageSet` script.
3. The `callCompiler` and `wget` utilities are causing a cycle in dependencies in the API module, so they have been moved to `Foreign.Purs` and `Foreign.Wget` respectively.
4. I've updated our fixtures to reflect that [every GitHub Actions workflow trigger includes who triggered it](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties) via the `sender` property. We now read this and propagate it to the operations so that we can look up the packaging team if a user tries to take a packaging-team-only action in this endpoint.

The bulk of the changes in this PR, though, are located in the new operation in the `runOperation` function. The rest is organizational work necessary to get everything to compile.


flake.nix
```diff
@@ -108,13 +108,13 @@
               spago run -m Registry.Scripts.LegacyImporter --node-args $1
             '';
 
-            registry-package-set = ''
+            registry-package-set-updater = ''
```
thomashoneyman, line 111: https://github.com/purescript/registry-preview/pull/2

colinwahl (approved):
Thanks for doing the reorganization work - looks good

PR #472, opened 2022-07-25
Don't include empty entries after checking unsatisfied dependencies

thomashoneyman:
The legacy importer validates packages and then uploads all publishable versions. If a package has no usable versions, then we write an empty entry into the metadata directory. However, if a package passes all validation _except_ the self-contained check, and all versions fail because of that check, then we don't write any metadata for it.

When we check for self-contained dependencies in the package graph we are not correctly filtering packages out of the registry index if they no longer have any usable versions. This PR fixes that bug.

These packages will have empty metadata written once this change goes in:

- couchdb-aff-coroutines
- fedger
- lnforum-api
- lumi-components
- mtproto
- stuff
- tiny-emitter


colinwahl (approved):
Good catch

PR #473, opened 2022-07-26
Automatically process legacy package transfers

### LINKED ISSUE #471: Add support for transfer operations in legacy importer
The legacy importer runs each day until the end of the registry alpha, fetching and publishing new releases for any packages listed in the `new-packages.json` and `bower-packages.json` file. So far it's been working well. However, there's still a manual step: if a package changes location, we continue to publish from the old location. Eventually someone tells us or we notice and we update the entry in the `new-packages.json` or `bower-packages.json` file, and then delete all the existing uploads, and re-run the importer to re-publish everything.

Ideally we could have the `Transfer` operation run automatically, just the way we already have the `Addition` and `Update` operations run automatically. There's an extra benefit besides the automation: right now we've never exercised this part of the API, and running it automatically means we can get some real-world testing on it.

In order to accomplish this we need to take two steps: determine when a repository has changed location, and initiate a Transfer operation.

### Identifying repository location changes

When we run the registry importer we begin by reading in the `new-packages.json` and `bower-packages.json` files. Then, for every GitHub address listed, we fetch the released tags. This is the earliest point at which we can determine whether a repository has changed location.

The easiest way to check is by parsing more of the response to the tags request so that we get the URL pointing to the commit. If the URL contains a location that is different from the GitHub address we requested, then we know the repo needs to be transferred. The response to the request looks like this:
https://docs.github.com/en/rest/repos/repos#list-repository-tags

Here's an excerpt:

```json
  {
    "name": "v0.1",
    "commit": {
      "sha": "c5b97d5ae6c19d5c5df71a34c7fbeeda2479ccbc",
      "url": "https://api.github.com/repos/octocat/Hello-World/commits/c5b97d5ae6c19d5c5df71a34c7fbeeda2479ccbc"
    }
  }
```

A more straightforward solution would be to check for a 301 redirect when fetching the address listed in the bower-packages and new-packages files, but this isn't possible with Octokit, and I don't want to make any more requests to GitHub than necessary (ie. by using something like `curl` to check for redirects). It's easy to get the info we need via the retrieved tags.

It's possible that we will miss repositories that have no published tags and then changed location, but since nothing is ever uploaded to the registry in that case it doesn't feel like a big deal.

### Initiating a transfer

We can run transfers as part of the registry importer or as a separate script before or after the importer runs (similar to what we do for package set updates). It doesn't matter in the sense that transferring before, during, or after won't affect the importer — the addition and update operations will follow the redirect, if there is one, and upload the package all the same. All that matters is that we're able to run `listTags` over the new-packages and bower-packages files.

After we've determined if any packages need to be transferred we can move on to producing `Transfer` operations for each package to transfer. A `Transfer` is an authenticated operation:

https://github.com/purescript/registry/blob/8408a993aa93ff2e9403fe42f0306793d520b7a6/ci/src/Registry/Schema.purs#L220-L227

where the payload is transfer data:

https://github.com/purescript/registry/blob/8408a993aa93ff2e9403fe42f0306793d520b7a6/ci/src/Registry/Schema.purs#L262-L265

and the "signature" is an SSH signature. The JSON representation of a transfer of the `math` package to the `purescript-deprecated` organization might look like this:

```json
{
  "payload": { 
    "packageName": "math",
    "newPackageLocation": { 
      "githubOwner": "purescript-deprecated",
      "githubRepo": "purescript-math"
    }
  },
  "signature": "-----BEGIN SSH SIGNATURE-----",
  "email": "pacchettibotti@purescript.org"
} 
```

We know the package name and the new package location already, because we needed that to know that we needed to initiate a transfer. We can use the `pacchettibotti` email address, and use the `pacchettibotti` SSH keys to produce the SSH signature. This would require putting the `pacchettibotti` SSH keys into the environment for CI to read.

We can produce the SSH signature according to this article:
https://www.agwa.name/blog/post/ssh_signatures

...and we already have some utilities in place in the [Registry.SSH](https://github.com/purescript/registry/blob/master/ci/src/Registry/SSH.purs) module, which are currently targeted for verifying signatures but could be tweaked to generate them as well.

thomashoneyman:
Fixes #471 by adding a script that transfers packages using the `@pacchettibotti` account. The most notable part of this is that authenticated operations can now all be signed by `@pacchettibotti` as the representative of the registry trustees, which works because only registry trustees have access to the PacchettiBotti keys.

I've already used this in testing to implement transfers for a number of packages that have moved since being registered. See, for example:
https://github.com/purescript/registry-preview/commit/2d5c727cf04a19118111f2781ee8a1e33ba58be9

There is one more change here that's causing a big diff: the new-packages and bower-packages files are regenerated when we have packages to transfer. When we do this the new files are sorted alphabetically. Right now they're in an arbitrary order; sometimes people insert their packages at the top, or the bottom, or next to another of their packages; after this change, packages will always be sorted alphabetically. Each day when the transferrer runs the package files will be re-sorted. Hopefully that's OK; it's a bit more trouble to avoid this.

With this in place we should no longer have to manually wait for package authors to notify us to change the packages they've registered.


thomashoneyman:
Uh oh, looks like I borked my shell a bit and committed under the `@pacchettibotti` account. Luckily this doesn't cause any problems.

ci/src/Foreign/Git.purs
```diff
@@ -0,0 +1,32 @@
+module Foreign.Git where
```
thomashoneyman, line 1: This just moves the git commands out of the API file and makes no other changes.

ci/src/Registry/API.purs
```diff
@@ -300,63 +300,61 @@ runOperation operation = case operation of
...
             , "Please publish a package version with your SSH public key in the owners field."
             , "You can then retry unpublishing this version by authenticating with your private key."
             ]
-        Just owners ->
-          liftAff (SSH.verifyPayload owners auth) >>= case _ of
-            Left err ->
-              throwWithComment $ String.joinWith "\n"
-                [ "Failed to verify package ownership:"
-                , "  " <> err
+        Just owners -> liftAff (SSH.verifyPayload owners auth) >>= case _ of
```
thomashoneyman, line 316: The changes in this file are almost all indentation-related. The only thing I've changed here is inserting pacchettibotti into the owners if the email matches.

ci/src/Registry/SSH.purs
```diff
@@ -36,8 +44,34 @@ verifyPayload owners (AuthenticatedData { email, signature, rawPayload }) = do
...
     pure $ bimap String.trim String.trim case result.exit of
       NodeProcess.Normally 0 -> Right result.stdout
       _ -> Left result.stderr
+
+type SignAuthenticated =
+  { pacchettiBottiPrivateKey :: String
+  , rawPayload :: String
+  }
+
+signPacchettiBotti :: SignAuthenticated -> Aff (Either String (Array String))
```
thomashoneyman, line 58: This is safe to have in our codebase because this can only be run if you have the private key for the pacchettibotti account, and that's only pulled from the environment. We do have to be careful to not arbitrarily sign things in the codebase, but that's not much different from how we need to be careful not to e.g. disable verification checks.

ci/src/Registry/Scripts/PackageTransferrer.purs
```diff
@@ -0,0 +1,212 @@
...
+  locations <- latestLocations packages
+  newPackages <- transferAll packages (Map.catMaybes locations)
+  liftAff $ writeLegacyRegistryFile legacyFile packages
+  liftAff (commitLegacyRegistryFile legacyFile) >>= case _ of
+    Left err -> throwWithComment err
+    Right _ -> pure unit
+
+transferAll :: Map String GitHub.PackageURL -> Map String PackageLocations -> RegistryM (Map String GitHub.PackageURL)
+transferAll packages packageLocations = do
+  pacchettiBottiPrivateKey <- liftEffect do
```
thomashoneyman, line 79: The key is stored in the GitHub secrets.

ci/src/Registry/Scripts/PackageTransferrer.purs
```diff
@@ -141,11 +141,9 @@ latestLocations packages = forWithIndex packages \package location -> do
               | otherwise -> pure $ Just locations
   where
   -- The eq instance for locations has case sensitivity, but GitHub doesn't care.
-  -- We require that locations are lowercase, so we can compare that directly. But
-  -- owner names can be case-insensitive; we don't enforce this.
   locationsMatch :: Location -> Location -> Boolean
   locationsMatch (GitHub location1) (GitHub location2) =
-    location1.repo == location2.repo
+    (String.toLower location1.repo == String.toLower location2.repo)
```
thomashoneyman, line 146: I needed to test this on a few packages, so I used case sensitivity in the repo name as one reason to transfer a package. I've now reverted it as it doesn't actually matter.

new-packages.json
```diff
@@ -1,225 +1,225 @@
...
-  "purescript-react-icons": "https://github.com/andys8/purescript-react-icons.git",
-  "purescript-web-router": "https://github.com/robertdp/purescript-web-router.git"
+  "purescript-whatwg-html": "https://github.com/ursi/purescript-whatwg-html.git",
+  "purescript-which": "https://github.com/maxdeviant/purescript-which.git",
+  "purescript-wire": "https://github.com/robertdp/purescript-wire.git",
+  "purescript-workly": "https://github.com/cakekindel/purescript-workly.git",
+  "purescript-yoga-fetch": "https://github.com/rowtype-yoga/purescript-yoga-fetch.git",
+  "purescript-yoga-json": "https://github.com/rowtype-yoga/purescript-yoga-json.git",
+  "purescript-yoga-postgres": "https://github.com/rowtype-yoga/purescript-yoga-postgres.git"
 }
```
thomashoneyman, line 225: In lieu of manually reviewing these files, you can at least take solace in that they contain the same number of packages.

ci/src/Registry/Scripts/LegacyImporter.purs
```diff
@@ -135,11 +135,18 @@ main = launchAff_ do
...
+      metadataMap <- liftEffect $ Ref.read metadataRef
+      case Map.lookup package metadataMap of
+        Nothing -> do
+          let metadata = { location, owners: Nothing, published: Map.empty, unpublished: Map.empty }
+          liftAff $ Json.writeJsonFile (API.metadataFile registryPath package) metadata
+          liftEffect $ Ref.modify_ (Map.insert package metadata) metadataRef
+          commitMetadataFile package >>= case _ of
+            Left err -> throwWithComment err
+            Right _ -> pure unit
+        Just _ -> pure unit
```
f-f, line 149: This change looks correct, but the previous implementation does too - was it buggy?

ci/src/Registry/Scripts/PackageTransferrer.purs
f-f: How can the trustees sign with this if it's only stored in Github secrets?

ci/src/Registry/Scripts/LegacyImporter.purs
thomashoneyman: The only difference is that we now commit metadata for reserved packages when in 'update' mode, which is what gets run in CI. Otherwise the metadata is written, but it doesn't get committed.

When running locally we generally use 'generate' mode, and then commit the resulting batch of changes; this doesn't affect that use case. The call to `commitMetadataFile` in 'generate' mode will just print "Not committing..." and succeed.

ci/src/Registry/Scripts/PackageTransferrer.purs
thomashoneyman: Yikes, I left that part out!

I would like to make pacchettibotti and/or the registry trustees "owners" of a package for the purpose of signing authenticated operations. That way when we take an action on behalf of a package (ie. unpublish) it is public and goes through the same pipeline as everything else. I know in the spec that we only consider unpublishing an action trustees can take, but I'd like to suggest that at least for legacy packages we allow trustees to transfer packages, too.

To accomplish this, we have some options:

1. We make the registry trustees "owners" of a package for the purpose of signing authenticated actions. We maintain a list of email + public key for the registry trustees somewhere, and the signature must match either the 'owners' field of the package manifest or the registry trustees list.
2. We make pacchettibotti the "owner" of a package for the purpose of signing actions. The key can be read in CI.

The only reason I've shied away from option (1) is that I'd rather not maintain a list of email + public keys for registry trustees somewhere. We'll likely forget to change it when adding or removing registry trustees. If there's a way to do this without us having to manage a file then I'd prefer option (1).

In option (2), my thinking was essentially that:

1. We rely on the GitHub API to look up who triggered the Registry API operation and see if they are a member of the packaging team. That means we can use team membership to gate sensitive operations, and not a list of email + public key that we have to maintain.
2. When we receive an authenticated operation we could look up who opened it. If it's a registry trustee and the email field is set to 'pacchettibotti@purescript.org', then we could re-sign the payload using the pacchettibotti private key from the CI environment and then process the operation. Packaging team members don't necessarily ever know what the key is (though it is stored in 1Password).

I'm not really sure the best way to do this, though. Do you have any suggestions?

f-f: Yes, this sounds great!
It is not implemented yet here, right?

thomashoneyman: It is not implemented here. Are you referring to option (2) as the one to go with? If so, then I can either add it to this PR or in a followup (whichever you prefer).

f-f: Yeah, option 2 is also how I would implement it, so it makes perfect sense to me.
It feels like this could go here in the same PR, since it's really about the same "feature"

thomashoneyman: @f-f you can see this implemented in https://github.com/purescript/registry/pull/473/commits/7d82be94cbb0c735c03e55ade1cf03306ac9278e, with a fixed documentation comment in https://github.com/purescript/registry/pull/473/commits/fd325a29452c6f904c8098b93328516370801d1b

f-f (approved):
Looks great, thanks!

PR #475, opened 2022-07-27
Add support for generating legacy package sets

### LINKED ISSUE #329: Package sets naming convention and mirroring
As [mentioned in the spec](https://github.com/purescript/registry#package-sets), this repo will be the source of truth for package set.

However, we would like to support (ideally for a long time) folks using them [from the current place](https://github.com/purescript/package-sets/releases) - this is the vast majority of the ecosystem right now, and they should not be forced to migrate all at once.

So we'll need to mirror them from [their new location](https://github.com/purescript/registry/tree/master/v1/sets) to the current one.

In practice that means that:
- we can clear out most of the `package-sets` repo. The only files left there will be a `README` (explaining the situation), the `packages.json` (for psc-package users), and the `src/packages.dhall` file (for Spago users).
- for every new package set published here, we will:
  - mirror the content to these two files
  - and publish a release in `package-sets` (note: this process copies the `src/packages.dhall` into a file in the release itself. This is important because for historical reasons Spago points to this one)

### Doing the conversion

We'll need to go from a bunch of
```dhall
...
aff = Address.Registry "v5.1.2"
...
```
to
```dhall
aff = { repo = "https://github.com/purescript-contrib/purescript-aff", dependencies = [...], version = "v5.1.2" }
```

We already have the data for the `version`, we'll need to get the dependencies and the URL from the `registry-index`.

### Naming convention

Our existing naming convention for package sets comes from the compiler version and the date the set it was published.
I always found this scheme incomplete, because it does not allow us to provide info about breakage - e.g. in the same way that [LTS Haskell](https://github.com/commercialhaskell/lts-haskell#readme) does - as the name of the set does not reflect the fact that we are dropping packages from the set.

So I would like to switch to a more informative naming scheme, but:
- I'm not sure how it would look like
- I'm not sure how to best do it in a non-breaking way

thomashoneyman:
Closes #329 by adding the ability to generate legacy package package sets from a new-style package set. Specifically, given a package set, we can generate a `packages.json` file and a `packages-dhall` file which each conform to the existing formats.

With this and the automatic package sets publishing we can officially migrate repositories, deprecate the existing package-sets publishing process, and replace it with mirrors from the registry proper.

### DIFF
--- a/ci/src/Registry/Legacy/PackageSet.purs
+++ b/ci/src/Registry/Legacy/PackageSet.purs
@@ -83,7 +83,7 @@ fromPackageSet index metadata (PackageSet { compiler, packages, published, versi
 
     pure
       { version: RawVersion metadataEntry.ref
-      , dependencies: Array.sort $ Array.fromFoldable $ Map.keys $ manifest.dependencies
+      , dependencies: Array.fromFoldable $ Map.keys $ manifest.dependencies
       , repo
       }
     where
--- a/ci/src/Registry/Legacy/PackageSet.purs
+++ b/ci/src/Registry/Legacy/PackageSet.purs
@@ -106,7 +106,7 @@ printDhall (LegacyPackageSet entries) = do
     Dodo.Common.jsCurlies
       $ Dodo.foldWithSeparator Dodo.Common.leadingComma
       $ map printPackage packages
-  
+
   quoteReserved :: String -> String
   quoteReserved input = do
     -- We can't use any reserved keywords:
--- a/ci/src/Registry/Legacy/PackageSet.purs
+++ b/ci/src/Registry/Legacy/PackageSet.purs
@@ -106,13 +106,20 @@ printDhall (LegacyPackageSet entries) = do
     Dodo.Common.jsCurlies
       $ Dodo.foldWithSeparator Dodo.Common.leadingComma
       $ map printPackage packages
+  
+  quoteReserved :: String -> String
+  quoteReserved input = do
+    -- We can't use any reserved keywords:
+    -- https://docs.dhall-lang.org/references/Built-in-types.html#built-in-types-functions-and-operators
+    if input `Array.elem` [ "assert", "let", "using", "missing", "merge", "if", "then", "else" ] then
+      Array.fold [ "`", input, "`" ]
+    else
+      input
 
   printPackage :: forall a. Tuple PackageName LegacyPackageSetEntry -> Dodo.Doc a
   printPackage (Tuple name entry) =
     Array.fold
-      [ Dodo.text do
-          let nameStr = PackageName.print name
-          if nameStr == "assert" then "`assert`" else nameStr
+      [ Dodo.text (quoteReserved (PackageName.print name))
       , Dodo.text " ="
       , Dodo.break
       , Dodo.indent (printEntry entry)
--- a/ci/src/Registry/Legacy/PackageSet.purs
+++ b/ci/src/Registry/Legacy/PackageSet.purs
@@ -2,8 +2,23 @@ module Registry.Legacy.PackageSet where
 
 import Registry.Prelude
 
+import Data.Array as Array
+import Data.Compactable (separate)
+import Data.Formatter.DateTime (FormatterCommand(..))
+import Data.Formatter.DateTime as Formatter.DateTime
+import Data.FunctorWithIndex (mapWithIndex)
+import Data.List as List
+import Data.Map as Map
+import Data.String as String
+import Dodo as Dodo
+import Dodo.Common as Dodo.Common
+import Registry.Index (RegistryIndex)
 import Registry.Json as Json
 import Registry.PackageName (PackageName)
+import Registry.PackageName as PackageName
+import Registry.Schema (Location(..), Manifest(..), Metadata, PackageSet(..))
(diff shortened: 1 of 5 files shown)


ci/src/Registry/Legacy/PackageSet.purs
```diff
@@ -15,16 +30,117 @@ instance RegistryJson LegacyPackageSet where
...
+
+fromPackageSet :: RegistryIndex -> Map PackageName Metadata -> PackageSet -> Either String ConvertedLegacyPackageSet
+fromPackageSet index metadata (PackageSet { compiler, packages, published, version }) = do
+  converted <- case separate $ mapWithIndex convertPackage packages of
+    { left, right } | Map.isEmpty left -> Right right
+    { left } -> do
+      let toValues = Array.fromFoldable <<< Map.values
+      Left $ String.joinWith "\n" $ Array.cons "Failed to convert packages:" $ toValues left
+
+  let converted' = Map.insert (unsafeFromRight (PackageName.parse "metadata")) metadataPackage converted
```
thomashoneyman, line 53: The metadata package is not registered in the new registry. However, this package is used in the legacy package sets to indicate the compiler version (though I'm hazy on the details and whether or not this is necessary anymore). When generating a legacy package set we always insert metadata using the package set compiler version.

ci/src/Registry/Legacy/PackageSet.purs
```diff
@@ -15,16 +30,117 @@ instance RegistryJson LegacyPackageSet where
...
+    versionStr = Version.printVersion packageVersion
+    noIndexPackageError = "No registry index entry found for " <> nameStr
+    noIndexVersionError = "Found registry index entry for " <> nameStr <> " but none for version " <> versionStr
+    noMetadataPackageError = "No metadata entry found for " <> nameStr
+    noMetadataVersionError = "Found metadata entry for " <> nameStr <> " but no published version for " <> versionStr
+    usesSubdirError = "Package " <> nameStr <> " uses the 'subdir' key, which is not supported for legacy package sets."
+
+printDhall :: LegacyPackageSet -> String
+printDhall (LegacyPackageSet entries) = do
+  let format = Dodo.twoSpaces { pageWidth = 80 }
```
thomashoneyman, line 100: This mimics the existing Dhall formatting. If the dependencies array, for example, fits within 80 characters then it will be formatted on one line. Otherwise, it will break to new lines.

ci/test/Registry/PackageSet.purs
```diff
@@ -0,0 +1,210 @@
...
+    ]
+
+  metadata = Map.fromFoldable
+    [ unsafeMetadataEntry assert
+    , unsafeMetadataEntry console
+    , unsafeMetadataEntry effect
+    , unsafeMetadataEntry prelude
+    ]
+
+legacyPackageSetJson :: String
```
thomashoneyman, line 92: This is what we produce from the above package set to conform to the packages.json format.

ci/test/Registry/PackageSet.purs
```diff
@@ -0,0 +1,210 @@
...
+    "version": "v0.15.2"
+  },
+  "prelude": {
+    "dependencies": [],
+    "repo": "https://github.com/purescript/purescript-prelude.git",
+    "version": "v6.0.0"
+  }
+}"""
+
+legacyPackageSetDhall :: String
```
thomashoneyman, line 131: This is what we produce from the above package set to conform to the existing packages.dhall format.

ci/src/Registry/Legacy/PackageSet.purs
thomashoneyman: @f-f do you happen to know if we still need to do this for the legacy package sets? I don't really know the details.

f-f: We still need the `metadata` package, as that's how Spago knows which PureScript version the package-set is compatible with.

thomashoneyman: Great. In that case this should be working properly. We always insert the metadata package with the compiler version used to produce the package set.

ci/src/Registry/Legacy/PackageSet.purs
```diff
@@ -15,16 +30,117 @@ instance RegistryJson LegacyPackageSet where
...
+    Dodo.Common.jsCurlies
+      $ Dodo.foldWithSeparator Dodo.Common.leadingComma
+      $ map printPackage packages
+
+  printPackage :: forall a. Tuple PackageName LegacyPackageSetEntry -> Dodo.Doc a
+  printPackage (Tuple name entry) =
+    Array.fold
+      [ Dodo.text do
+          let nameStr = PackageName.print name
+          if nameStr == "assert" then "`assert`" else nameStr
```
colinwahl: Are there any other reserved words we should be on the lookout for?

thomashoneyman: Technically, these are all reserved keywords:

* `assert`
* `let`
* `using`
* `missing`
* ~~`toMap`~~ (we disallow case-sensitivity in package names)
* `merge`
* `if`
* `then`
* `else`

that I suppose we should backtick if found.

colinwahl (approved):

colinwahl:
Looks like a module isn't formatted

PR #476, opened 2022-07-28
Add removals to batch package set API

### LINKED ISSUE #438: API endpoint for manual package set interventions

thomashoneyman:
Closes #438 by adding the last piece: support for package removals. To that end:

1. I've modified our existing package sets processing code to work on a `Map PackageName (Maybe Version)` instead of always having package versions available. No version indicates that a package is being removed.
2. I have added validation (with @colinwahl) for package removals. The rule: you can remove a package only so long as nothing in the package set depends on it, _unless_ that package is also being removed.
3. I've beefed up our error reporting so we can be clear about why exactly an addition, update, or removal did not go through.
4. I have split the batch processing function, `processBatch`, into two separate functions: `processBatchAtomic`, which fails if the entire batch can't go through, and `processBatchSequential`, which attempts to find the maximal set of packages from the candidates that can go into the next set. The batch API should fail if the whole suggested batch doesn't work, so that you can manually fix it. No one is manually fixing the automatic package sets, so those should find as many packages as possible.

Also, I fixed a bug in which we were reusing some logic from the automatic package sets and only considering packages from the last 24 hours. There's no reason to do that in the batch API — someone is specifically asking for _these_ packages to be considered, regardless of when they were published.

With this in place the batch API should be complete.

### DIFF
--- a/ci/src/Foreign/GitHub.purs
+++ b/ci/src/Foreign/GitHub.purs
@@ -196,7 +196,7 @@ createComment octokit issue body = do
 -- | https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/v5.16.0/docs/issues/update.md
 closeIssue :: Octokit -> IssueNumber -> ExceptT GitHubError Aff Unit
 closeIssue octokit issue = do
-  let args = { status: "closed" }
+  let args = { state: "closed" }
   _ <- Except.withExceptT APIError $ uncachedRequest { octokit, route, headers: Object.empty, args }
   pure unit
   where
--- a/ci/src/Foreign/GitHub.purs
+++ b/ci/src/Foreign/GitHub.purs
@@ -196,7 +196,7 @@ createComment octokit issue body = do
 -- | https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/v5.16.0/docs/issues/update.md
 closeIssue :: Octokit -> IssueNumber -> ExceptT GitHubError Aff Unit
 closeIssue octokit issue = do
-  let args = { status: "closed " }
+  let args = { status: "closed" }
   _ <- Except.withExceptT APIError $ uncachedRequest { octokit, route, headers: Object.empty, args }
   pure unit
   where
--- a/ci/src/Registry/API.purs
+++ b/ci/src/Registry/API.purs
@@ -256,16 +256,27 @@ runOperation operation = case operation of
     -- a version that isn't supported by the registry then an 'unsupported
     -- compiler' error will be thrown.
     registryIndex <- liftAff $ Index.readRegistryIndex registryIndexPath
-    candidates <- PackageSet.findPackageSetCandidates registryIndex latestPackageSet
+    let candidates = PackageSet.validatePackageSetCandidates registryIndex latestPackageSet packages
 
-    if Map.isEmpty candidates then do
-      comment "No new package versions eligible for inclusion in the package set."
-      closeIssue
+    when (not Map.isEmpty candidates.rejected) do
+      comment $ String.joinWith "\n"
+        [ "One or more packages in the suggested batch cannot be processed:"
+        , PackageSet.printRejections candidates.rejected
+        ]
+
+    if Map.isEmpty candidates.accepted then do
+      throwWithComment "No packages in the suggested batch can be processed; all failed validation checks."
     else do
-      PackageSet.processBatch registryIndex latestPackageSet compiler candidates >>= case _ of
+      -- TODO: Removals.
+      let updates = Map.catMaybes candidates.accepted
+      PackageSet.processBatch registryIndex latestPackageSet compiler updates >>= case _ of
         Nothing -> do
-          comment "No packages could be added or updated in the set. All candidates failed."
-          closeIssue
+          throwWithComment "No packages from the suggested batch could be added, updated, or removed. All failed compilation."
+        Just { fail } | not Map.isEmpty fail -> do
+          throwWithComment $ String.joinWith "\n"
+            [ "Some packages from the suggested batch could not be added, updated, or removed."
+            -- TODO: Print failure error messages?
+            ]
         Just { packageSet } -> do
           newPath <- PackageSet.getPackageSetPath (un PackageSet packageSet).version
           liftAff $ Json.writeJsonFile newPath packageSet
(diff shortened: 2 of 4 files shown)


thomashoneyman:
Voila! Here are the results of testing this out on the `registry-preview` repository.

**Package Removal**
This payload should remove the `slug` package:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198535618

Result:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198541198
https://github.com/purescript/registry-preview/commit/39d25e3e278500ba91a6b15069c2af9ec1f30b57

**Failed Multi-Package Removal**
This payload should fail to remove halogen-hooks and halogen-store, because halogen-hooks-extra also relies on halogen-hooks:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198556329

_Note: I had to edit this one because it uncovered bugs, which are fixed by commits 425bd54 and f52a036._

Result:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198562308

**Successful Multi-Package Removal**
Alright! Now that CI has told me I missed a package that also must be removed, I can remove it:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198562798

Result:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198566575
https://github.com/purescript/registry-preview/commit/6b48e957e7fb2dab81fa383e9b9065f947f4f0f5

**Successful Multi-Package Addition**
And now, finally, I can add all these packages back into the package set:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198575138

_Note: I found a typo in the close issue function, which I fixed in [5f8b100](https://github.com/purescript/registry/pull/476/commits/5f8b100f9fd51e428901fb910645d16977870fb4). Issues should be closed when a batch processes without error._

Result:
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198586326
https://github.com/purescript/registry-preview/commit/02b1e80aafabc49f63e1c37000838d6efc39c354

---

This one's a bonus: I accidentally swapped two package versions, but still got a nice error from it.
https://github.com/purescript/registry-preview/issues/4#issuecomment-1198574498

colinwahl (approved):

f-f:
Great work! :clap:

PR #479, opened 2022-08-05
Implement version solver

### LINKED ISSUE #453: Implement a solver for dependency ranges
The registry currently does not produce build plans for legacy packages when they are imported. Because of this, we also don't attempt to publish their documentation to Pursuit. We simply produce a bogus build plan and avoid compiling the package altogether.

However, when the registry launches we will want to verify that a package version compiles properly before admitting it to the registry. We currently do that by requiring that package managers submit a build plan as part of the registry operation. This also allows us to upload documentation for the package to Pursuit.

I would like to be able to fully process legacy packages and make build plan submission optional by implementing our own solver in the registry that can produce the build plan itself. As a reference, this is Elm's solver, which is only a few hundred lines:

https://github.com/elm/compiler/blob/0.19.1/builder/src/Deps/Solver.hs

We have simple version ranges (ie. only `>=MA.MI.PA <MA.MI.PA`, where the lhs must be less than the rhs), which helps make it a little easier to implement the solver. We also have all manifests already available via the registry index.

This issue tracks:

1. Implementing a solver for the registry
2. Using the solver to implement build plans for legacy packages

As a bonus, the registry solver will be part of the registry-as-a-library, which means that authors of package managers can reuse our solver if they would like to.

MonoidMusician:
Fixes #453.


ci/src/Registry/Version.purs
```diff
@@ -166,6 +167,16 @@ printRange range =
...
 
+intersect :: Range -> Range -> Maybe Range
+intersect (Range r1) (Range r2)
+  | r1.lhs > r2.rhs || r2.lhs < r1.rhs = Nothing
+  | otherwise = Just $ Range
+    { lhs: max r1.lhs r2.lhs
+    , rhs: min r1.rhs r2.rhs
+    , mode: Lenient
+    , raw: r1.raw <> " && " <> r2.raw
+    }
```
MonoidMusician: We should discuss if this looks okay. For the solver, `mode` and `raw` don't really matter, but I just figured it would be better to keep around the information of the individual bits that went in (in `raw`) if possible.

thomashoneyman: We may ultimately want to separate Range into Range and ParsedRange, where the former has “mode” and “raw” fields and the latter does not, and where intersection is only for ranges without parsing information. Ultimately we use ranges with parsing information for error reporting and to recover the original, but of course that’s not needed in something like the solver.

thomashoneyman: (Same can be true of the Version type)

natefaubion:
@thomashoneyman pinged us at Arista, so we looked at this during our team pairing session and had a few questions:

* It looks like this will fail on the first unsolvable constraint. Would it be worth noting a failed package solution, and continuing with other packages so that all failured could be reported?
* By using `oneOfMap` on a `Set Version`, will this result in a bias for older versions given multiple compatible versions?
* What do you think about using something like `RWSET` (`type SolverT = RWSET Dependencies Unit State Unit`) from uncurried-transformers? This would give you straightforward stack-safety and performance.

MonoidMusician:
Yup I just fixed item 2 :)

It looks like uncurried-transformers isn't in the 0.14.5 package set. Is it enough to make the base monad `Trampoline` instead of `Identity` for this?

Regarding the first: I'm not sure what good errors will look like. Right now it doesn't have any errors. It just backtracks until it cannot backtrack anymore.

thomashoneyman:
With regards to getting access to `uncurried-transformers`, it looks like everything it builds with is available in 0.14.5 and it doesn't appear to use anything 0.15-specific. So we could add an override for it to the spago.dhall file if you want it:
https://github.com/PureFunctor/purescript-uncurried-transformers/blob/main/spago.dhall

natefaubion:
> Regarding the first: I'm not sure what good errors will look like. Right now it doesn't have any errors. It just backtracks until it cannot backtrack anymore.

Maybe one option would be to compute the disjoint ranges that would be required. So if you hit a conflict you essentially create two buckets for the conflicts and the packages that require it. As you continue to solve you put the range in the any of the buckets that match with the package that demanded it. A successful build plan is the one where every package has exactly one bucket, and those with multiples are the conflicts and where those requirements come from.

MonoidMusician:
Thanks for the idea, I'll investigate it. I think it's a little more complicated because the other mode of failure is “version range was coherent, but did not exist in package index”. That error will be visible in `getRelevantVersions`.

I suppose I could just go full monoid mode. That is my specialty after all 😉

MonoidMusician:
I guess you would end up with a tree of errors, if you keep all of the information:
```purescript
type ErrorTree = Free Layer Conflict
-- empty map indicates no suitable versions in registry
data Layer a = Layer PackageName (Map Version a)
data Conflict
  = OutOfRange Range Version
  | DisjointRanges Range Range
```

But I'm not sure if you want to actually collect all of that information and print it. I would look something like
```
Dependency `arrays` not satisfied:
- 1.0.0:
  - Dependency `mysterious` not satisfied:
    - No matching versions
- 1.1.1:
  - Dependency `prelude` not satisfied:
    - 2.0.0: Out of range
    - 3.0.0: Out of range
```

natefaubion:
I guess I wouldn't be interested in the full tree of error, and I think that would imply you need to observe the entire solution space, which probably is not ideal. I think I would only be interested in the top-level goals that can't be satisfied, and the paths from other top-level goals that created those conflicts. That is, you're doing the same amount of work for each individual package as if you bailed early, but instead of bailing on all solving, you just move on to the next package. If when solving a subsequent package, you uncover another conflict in a failed package, record that too.

thomashoneyman:
One other thing to consider is that the solver will operate on a provided manifest, where a manifest lists out dependencies as key-value pairs of a package name and a dependency range. Notably, it's possible that the manifest omits transitive dependencies.

This matters because we're solving on a manifest in order to pass the result to the compiler; we're ultimately producing a [`BuildPlan`](https://github.com/purescript/registry/blob/f464fe29cc36ae837be0653f14f34be954cf6e1b/ci/src/Registry/Schema.purs#L102-L105). At some point we have to install the transitive dependencies, if there are any, so we'll have to resolve those as well.

The registry index, which this implementation uses to find all "relevant" versions for the package, also lists the dependency ranges for all of its dependencies at that version. Is it worth tracking that information in the solver?

In other words, say we find a solution:

```
console >=1.0.0 <3.0.0  -> 2.5.1
effect >=1.0.0 <2.0.0   -> 1.6.0
```

Then, we identify that these both depend on `prelude`, so we need to solve for a specific `prelude` version. But our solution no longer works: `console` versions beyond v2 use an incompatible version of prelude with the `effect` version we solved. We can either error here, or we can backtrack and discover a different solution that works, including the transitive dependencies:

```
console >=1.0.0 <3.0.0  -> 1.5.1
effect >=1.0.0 <2.0.0   -> 1.6.0
prelude (transitive)    -> 1.5.0
```

Perhaps this is too complicated to add to the solver in the first version — but we're going to have to find specific versions for transitive dependencies somehow. What do y'all think @natefaubion @MonoidMusician?

MonoidMusician:
It already solves for transitive dependencies from the registry index, see `getConstraints` and `addConstraint`. Is there more that you think needs to be added?

natefaubion:
I think maybe the `Dependencies` name is a bit misleading (or at least warrants documentation about your assumptions), as it can be read as only data for immediate dependencies.

thomashoneyman:
Ah, I misread. So long as we are indeed picking up transitive dependencies :+1:

MonoidMusician:
It doesn't matter whether `Dependencies` is transitive or not – during the course of solving, when trying a particular version of a package, the dependencies of that particular version are also recorded as needing to be solved, which takes care of transitive dependencies eventually even if they aren't immediately listed.

MonoidMusician:
I added basic errors. It won't report multiple errors. If you think it should lmk what you'd like to see without it getting overwhelming.

thomashoneyman:
@MonoidMusician I’ll check on those errors later tonight. In the meantime the codebase is now on 0.15 if you’d like to install uncurried-transformers.

ci/src/Registry/Solver.purs
```diff
@@ -0,0 +1,173 @@
...
+import Registry.Version (Range, Version, intersect, rangeIncludes)
+
+type Dependencies = Map PackageName (Map Version (Map PackageName Range))
+
+data SolverPosition
+  = SolveRoot
+  | Solving PackageName Version SolverPosition
+
+derive instance eqSolverPosition :: Eq SolverPosition
+derive instance ordSolverPosition :: Ord SolverPosition
```
thomashoneyman: We standardized on not naming instances in the codebase – do you mind trimming these (and other instances in this file)?

```suggestion
derive instance Eq SolverPosition
derive instance Ord SolverPosition
```

ci/src/Registry/Solver.purs
```diff
@@ -0,0 +1,173 @@
...
+
+instance showSolverPosition :: Show SolverPosition where
+  show SolveRoot = ""
+  show (Solving name version pos) = " while solving " <> show name <> "@" <> show version <> show pos
```
thomashoneyman: We only use show instances for testing's sake, but not when we explicitly want to print a value for the sake of e.g. error messages. For printing we generally write `print*` functions, ie.

```suggestion
printSolverPosition :: SolverPosition -> String
printSolverPosition = case _ of
  SolveRoot -> ""
  Solving name version pos -> " while solving " <> show name <> "@" <> show version <> show pos
```

and, if this is solely used for printing the error messages out, you could fold it into a single

```purs
printSolverError :: SolverError -> String
printSolverError = case _ of
  NoVersionsInRange -> ...
  where
  printSolverPosition = case _ of
    ...
```

so that we're not arbitrarily using this.

ci/test/Registry/Solver.purs
```diff
@@ -0,0 +1,169 @@
...
+import Registry.Version as Version
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+spec :: Spec.Spec Unit
+spec = do
+  let
+    p = unsafeFromRight <<< PackageName.parse
+    v = unsafeFromRight <<< Version.parseVersion Strict <<< (_ <> ".0.0") <<< show
+    r l u = unsafeFromRight <<< Version.parseRange Strict
```
thomashoneyman: You can also parse lenient ranges, ie `Version.parseRange Lenient "^1"` is equivalent to `>=1.0.0 <2.0.0`. What you have here works, but just so you know – you can write flexible ranges so long as you're in lenient mode!

ci/test/Registry/Solver.purs
```diff
@@ -0,0 +1,169 @@
...
+import Registry.Version (ParseMode(..))
+import Registry.Version as Version
+import Test.Spec as Spec
+import Test.Spec.Assertions as Assert
+
+spec :: Spec.Spec Unit
+spec = do
+  let
+    p = unsafeFromRight <<< PackageName.parse
+    v = unsafeFromRight <<< Version.parseVersion Strict <<< (_ <> ".0.0") <<< show
```
thomashoneyman: I've been having a little trouble scanning through the code with these single-letter names; would you be OK with renaming these to say `unsafePackage`, `unsafeVersion`, `unsafeRange`, or at least `package`, `version`, `range`? A find-replace would do.

ci/src/Registry/Solver.purs
```diff
@@ -0,0 +1,173 @@
...
+type Solved = Map PackageName Version
+type State =
+  { pending :: Goals
+  , solved :: Map PackageName Version
+  }
+
+oneOfMap1 :: forall f a b. Alt f => (a -> f b) -> NonEmptyArray a -> f b
+oneOfMap1 = alaF Alternate foldMap1
+
+solve :: Dependencies -> Map PackageName Range -> Either SolverError Solved
```
thomashoneyman: I'd like to be able to report a list of errors if possible, though it's hard for me to judge how much additional work that is. So, for example, if two ranges have no valid versions in the index, such as:

```
prelude: >=15.0.0 <16.0.0
effect: >=15.0.0 <16.0.0
```

I'd like to be able to report errors along the lines of:

```
Some dependencies could not be solved:

- prelude >=15.0.0 <16:0.0: No versions available in this range in the index
- effect >=15.0.0 <16:0.0: No versions available in this range in the index
```

Perhaps this would entail a function of the form:

```purs
solve :: Dependencies -> Map PackageName Range -> { failed :: Map PackageName SolverError, success :: Solved }
```

...which is similar to validation functions we use elsewhere. I'm curious what you think about this and whether it seems achievable (whether in this PR or in a later one)!

thomashoneyman (commented):
I have a few minor comments, but on the whole this is looking really nice. The error messages read well. Really appreciate your work on this!

MonoidMusician:
Thanks Thomas, I'll address it in the morning! I will also push what I have for multiple errors. It actually won't report errors for multiple packages. What it will do instead is report errors for each version of the package it tried, since it needs all of those to fail before it fails of course. That was the easiest to adapt in the code, but it's possible that something can be done for multiple packages. As Nate was alluding to, we can recover on errors and try to solve the remaining constraints before ultimately failing with the accumulated errors.

thomashoneyman:
Yea, recovering on errors and progressing to the remaining constraints sounds like the right thing to do. Please let me know if that's something I should consider as part of this PR or whether you would prefer it be considered separately in a followup.

MonoidMusician:
Yeah, let me give it a quick try today and if it doesn't work smoothly I'll save it for another PR.

ci/src/Registry/Solver.purs
MonoidMusician: You have betrayed all that we once stood for 😱😛

thomashoneyman: 😇 I have moved beyond

MonoidMusician:
`uncurried-transformers` does not do the backtracking we need, unfortunately 😕 I think because it doesn't include a new error continuation in the success callback? Not sure.

MonoidMusician:
Debrief on what the issue was that forced this change:

```diff
-        oneOfMap1 (addVersion pos name) =<< getRelevantVersions pos name constraint
-        exploreGoals
+        versions <- getRelevantVersions pos name constraint
+        versions # oneOfMap1 \version ->
+          addVersion pos name version *> exploreGoals
```

Because the CPS monad transformer had an error callback `(e -> m r)` and _also_ requested that error callback in the `pure` callback `((e -> m r) -> a -> m r)` it was able to do non-local control flow/backtracking. (Like `callCC` or something, idk.)

Even though `oneOfMap1 (addVersion …)` was called first, and then `exploreGoals` was bound after that, `exploreGoals` was allowed to backtrack into the `oneOfMap1`. Effectively it worked like the version I changed it to, where `exploreGoals` is explicitly called _inside_ `oneOfMap1`.

That change clarifies the backtracking behavior and allows using `uncurried-transformers` (which was a pretty minor diff otherwise, as you can see).

It also allows me to make progress on better errors. Unfortunately it invalidates my mental model in the process, so it will require more tweaking.

Now that section looks like this:
```purescript
        versions <- getRelevantVersions pos name constraint
        let
          act = versions # oneOfMap1 \version ->
            addVersion pos name version *> exploreGoals
        catchError act \e1 -> do
          _ <- catchError exploreGoals \e2 -> do
            throwError (e1 <> e2)
          throwError e1
```

Right now it's probably gathering too many errors, _all_ of the errors that occurred on each failed version, so I'll look into trimming that down. In particular, we probably only want one error for each package@version that failed, so `oneOfMap1` should only take one error from each branch.

thomashoneyman:
I think we would at most report one error per dependency, likely whatever the first error to be encountered was. An output such as:

`{ failed :: Map PackageName (Tuple Range Error), succeeded :: Map PackageName Version }`

…seems like it would give us what we need, where any failure represents a failure to produce a valid build plan.

thomashoneyman:
As far as formatting the error output goes, I think something like this (suggested by @natefaubion) looks good:

```
Can't find compatible ranges for:
* foldable-traversable
  * >=5.0.0 <6.0.0
  * >=6.0.0 <7.0.0 (from datetime)
```

thomashoneyman:
I've removed the integration test and I think we should merge this as-is. We can then do a few followups:

1. Improve the error messages at the very least with de-duplication
2. Add a 'strict' section to the integration test that tests Bower solving the exact Bower files from each package version against our solver solving the same files
3. Make a decision around how to handle the registry importer currently creating a registry index with incoherent dependency ranges for some packages

thomashoneyman (approved):

PR #480, opened 2022-08-05
Package Transfers: Fix bug when attempting to commit no changes

thomashoneyman:
I noticed that the transferrer failed to run yesterday because there were no changes, and committing with no changes is an error. This adds a little logging and fixes the issue, as demonstrated in:
https://github.com/purescript/registry-preview/actions/runs/2804252259


thomashoneyman:
The commit in this PR came from:
https://github.com/purescript/registry-preview/runs/7697229040?check_suite_focus=true

However, we don't want this to be committed to branches, only master, so I'm going to revert that part of this commit.

thomashoneyman:
Uncovered a few issues while working on this. They include:

1. We have to include both the public key and the private key as files on the file system when producing an SSH signature (I thought it was only the private key, as that's the only argument, and in testing I had both on disk already). I have updated the code to reflect that change.
2. SSH signatures _must_ use a public key and a private key that have a trailing newline. It's hard to tell with GitHub secrets and in .env files what the state of your newlines is and whether they're proper, so now we store the secrets as base64-encoded strings where it's not possible to mess that up.
3. I no longer store the public key in the code, so we can update the key without touching the code.

Unfortunately, there appears to be a difference between my local openssh version and the version used in GitHub Actions, so while the tests pass on my end they fail in CI.

thomashoneyman:
This works, but the fact this fails with the built-in OpenSSH for GitHub Actions (I haven’t been able to look up what that is) and succeeds with the one provided by nixpkgs is a bit concerning. I’d like to make sure that we can support as many SSH versions as possible for the sake of signing and verifying signatures.

thomashoneyman:
I still think we should get this in because without it transfers don’t work period, but we should circle back and see what to do about the OpenSSH versions.

https://github.com/purescript/registry/pull/480/commits/2a4c956c232d941e02fcbb0143327cd5693e813d

^ that’s the commit generated by pacchettibotti from the registry-preview repo, with the corresponding update to the metadata:

https://github.com/purescript/registry-preview/commit/88cf348ca2f41e42b2eb4b20c9b1d782f6c20803

ci/src/Registry/API.purs
```diff
@@ -1184,30 +1180,86 @@ acceptTrustees username authData@(AuthenticatedData authenticated) maybeOwners =
...
 
-        pacchettiBottiPrivateKey <- liftEffect do
-          mbKey <- Node.Process.lookupEnv "PACCHETTIBOTTI_ED25519"
-          maybe (throw "PACCHETTIBOTTI_ED25519 not defined in the environment.") pure mbKey
+        { publicKey, privateKey } <- readPacchettiBottiKeys
 
-        signature <- liftAff (SSH.signPacchettiBotti { rawPayload: authenticated.rawPayload, pacchettiBottiPrivateKey }) >>= case _ of
-          Left _ -> throwWithComment "Error signing transfer. cc: @purescript/packaging"
+        signature <- liftAff (SSH.signPayload { publicKey, privateKey, rawPayload: authenticated.rawPayload }) >>= case _ of
+          Left err -> log err *> throwWithComment "Error signing transfer. cc: @purescript/packaging"
```
colinwahl: Did you intend to leave this log statement here?

colinwahl (approved):

ci/src/Registry/API.purs
thomashoneyman:
```suggestion
          Left _ -> throwWithComment "Error signing transfer. cc: @purescript/packaging"
```

PR #482, opened 2022-08-07
Make registry scripts runnable from any directory in the project

thomashoneyman:
The registry scripts (`verify-registry-unique`, `registry-importer`, and so on) are expected to be run from different locations in the repository. For example, of the two scripts mentioned, the former must run in the root and the latter must run in the `ci` directory.

This commit updates the scripts so that they are always run in the location they are supposed to be, regardless of where you executed the command from. For example, I can now execute `registry-importer` in the root of the project and it will work, or execute `verify-registry-unique` from within `ci`, and so on.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,11 +46,11 @@ jobs:
       - name: "Check that all Dhall compiles, and examples correctly conform to a Manifest"
         run: nix develop --command 'registry-verify-dhall'
 
+      - name: "Install dependencies"
+        run: nix develop --command 'registry-install'
+
       - name: "Run tests"
-        run: |
-          cd ci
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -46,11 +46,11 @@ jobs:
...
 
+      - name: "Install dependencies"
+        run: nix develop --command 'registry-install'
+
       - name: "Run tests"
-        run: |
-          cd ci
-          nix develop --command 'registry-install'
-          nix develop --command 'registry-test'
+        run: nix develop --command 'registry-install'
```
thomashoneyman:
```suggestion
        run: nix develop --command 'registry-test'
```

colinwahl (approved):

PR #485, opened 2022-08-09
Upgrade codebase to PureScript 0.15

thomashoneyman:
This PR updates our codebase to PureScript 0.15 and a recent package set. This is done for a few reasons:

1. It keeps us up-to-date with the most recent compiler and library versions
2. Some packages that contributors to this project would like to use (#479, for example) are only available in the 0.15 package sets.
3. Some packages have seen significant improvements recently (`parsing`, for example), and I'd like to take advantage of them.

The main changes here come from using ES modules in the FFI and switching from `string-parsers` to `parsing` (now that the latter is performant on strings).


ci/src/Foreign/S3.js
```diff
@@ -1,24 +1,24 @@
-const AWS = require("aws-sdk");
+import aws from "aws-sdk";
```
thomashoneyman, line 1: This is not an ES modules project, so we can't destructure the imports -- hence using the `aws` namespace.

ci/src/Foreign/SPDX.js
```diff
@@ -1,13 +1,18 @@
...
-var deprecatedIds = require("spdx-license-ids/deprecated");
-var Fuse = require("fuse.js");
+import parse from "spdx-expression-parse";
+import Fuse from "fuse.js";
+import { createRequire } from "module";
 
-const identifiers = ids.concat(deprecatedIds);
+// The spdx-license-ids project is just a set of JSON files, which can't be
+// imported directly into ESM without import assertions to guide TypeScript.
+const require = createRequire(import.meta.url);
```
thomashoneyman, line 7: It's funky to import files directly using ES modules imports, whereas this is somewhat common in CJS. The standard approach as far as I'm aware is to create `require` and use it for those kinds of imports. You can see this import being exercised in the tests, and it's how we do this at work as well.

ci/src/Registry/PackageName.purs
```diff
@@ -58,36 +59,35 @@ parse inputStr = flip Parser.runParser inputStr do
...
-    char = ParseC.choice [ Parse.lowerCaseChar, Parse.anyDigit ] <?> charErr
-    dash = void $ Parse.char '-'
-    chunk = ParseC.many1 char
+    acceptedChars = Parsing.Combinators.choice [ Parsing.String.Basic.lower, Parsing.String.Basic.digit ] <|> Parsing.fail charErr
+    chunk1 = Parsing.Combinators.Array.many1 acceptedChars
 
-  -- A "chunk" is an alphanumeric word between dashes
-  firstChunk <- chunk <?> startErr
+  -- A "chunk" is a lowercase alphanumeric word delimited by dashes
+  firstChunk <- chunk1 <|> Parsing.fail startErr
```
thomashoneyman, line 69: The `<?>` operator in `Parsing` prepends `"Expected "` to the error message, which we don't want. Instead, we just provide the failure message as the fallback case using `<|>` directly.

ci/src/Registry/PackageName.purs
```diff
@@ -58,36 +59,35 @@ parse inputStr = flip Parser.runParser inputStr do
...
 
   nextChunks <- do
-    chunks <- ParseC.many do
-      void dash
-      void $ ParseC.optionMaybe (ParseC.lookAhead Parse.anyChar) >>= case _ of
+    chunks <- flip Parsing.Combinators.Array.manyTill_ Parsing.String.eof do
```
thomashoneyman, line 72: We don't want to use `many` because in the `parsing` library (unlike the `string-parsers` library), on failure this will simply return the results so far. We were previously relying on a failure terminating the overall parser. Instead, we run our parser until it reaches the end of the string and a failure along the way should exit the parser.

natefaubion: Is this a bug in `Parsing.Combinators.Array`?

thomashoneyman: It could be; I don’t know which behavior is the desired one. I do know that the string-parsers one behaves differently.

thomashoneyman: @jamesdbrock do you have thoughts on the difference in behaviors between the string-parsers ‘many’ and the use of ‘many’ in parsing? Specifically, all three of these implementations exhibit different behavior (you can see for yourself by cloning this repo and running `spago test`; if you're not in the Nix shell you'll see some other test failures but the ones to look for specifically are the `PackageName` tests):

Straightforward adaptation of the original retaining the use of the `many` combinator (this fails):

```purs
  nextChunks <- Parsing.Combinators.Array.many do
    _ <- Parsing.String.char '-' <|> Parsing.fail charErr
    _ <- Parsing.Combinators.optionMaybe (Parsing.Combinators.lookAhead Parsing.String.anyChar) >>= case _ of
      Just v
        | v == '-' -> Parsing.fail manyDashesErr
        | otherwise -> pure unit
      Nothing -> Parsing.Combinators.lookAhead Parsing.String.eof *> Parsing.fail endErr
    map (NonEmptyArray.cons '-') chunk1 <|> Parsing.fail endErr
```

Using `Array.many` directly (this fails, but differently):

```diff
+  nextChunks <- Array.many do
-  nextChunks <- Parsing.Combinators.Array.many do
```

Using `manyTill_` (this matches the `string-parsers` version of `many` for our use case and works):

```diff
+  nextChunks <- map fst $ flip Parsing.Combinators.Array.manyTill_ Parsing.String.eof do
-  nextChunks <- Parsing.Combinators.Array.many do
```

colinwahl (approved):
Barring the questions on behavior of `many`, this looks good to me :+1:

thomashoneyman:
I'm going to merge because that question affects external libraries; we can continue the discussion here, but it isn't causing an issue with correctness for this PR.

thomashoneyman:
This is working:
https://github.com/purescript/registry-preview/runs/7773985610?check_suite_focus=true

PR #487, opened 2022-08-14
Clean up solver tests

thomashoneyman:
@MonoidMusician this PR makes a couple tweaks to the solver tests (I'm still working on a real-world usage test). Namely:

1. The error / success helpers are a bit more informative in the case a test goes wrong (you get a selective part of the diff instead of the entire blob)
2. Each test case has its own `Spec.it` block, which means they can be enabled or disabled selectively, or if one fails you know what specifically failed.
3. Package names and versions have been stored as variables so they stay in sync everywhere in the tests.
4. We're now testing error messages as well as errors; previously the messages were in comments, but these'll fall out out of sync, and testing the messages revealed some errors coming from the use of `show` (which I've fixed).


thomashoneyman:
In the latest commits I've added an integration test that attempts to solve all core libraries from the 1.0.0 package set. Those versions are hard-coded into the tests. The test can be run with:

```console
spago test -m Test.Integration
```

Right now I'm seeing some interesting errors! For example:

```
✗ tuples@7.0.0:

Committed to prelude in range >=6.0.0 <7.0.0 but the range >=6.0.0 <7.0.0 was also required while solving control@6.0.0
Committed to prelude in range >=6.0.0 <7.0.0 but the range >=6.0.0 <7.0.0 was also required while solving invariant@6.0.0
Committed to prelude in range >=6.0.0 <7.0.0 but the range >=6.0.0 <7.0.0 was also required while solving newtype@5.0.0 while solving control@6.0.0
```

Here are the dependencies for `tuples@7.0.0` according to the registry index:

```jsonc
// tuples@7.0.0
{"control":">=6.0.0 <7.0.0","invariant":">=6.0.0 <7.0.0","prelude":">=6.0.0 <7.0.0"}

// invariant@6.0.0
{"control":">=6.0.0 <7.0.0","prelude":">=6.0.0 <7.0.0"}

// control@6.0.0
{"newtype":">=5.0.0 <6.0.0","prelude":">=6.0.0 <7.0.0"}

// newtype@5.0.0
{"prelude":">=6.0.0 <7.0.0","safe-coerce":">=2.0.0 <3.0.0"}

// safe-coerce@2.0.0
{"unsafe-coerce":">=6.0.0 <7.0.0"}

// unsafe-coerce@6.0.0
{}

// prelude@6.0.0
{}
```

Accordingly, I would expect this to produce a solution.

thomashoneyman:
In the latest I fix an issue with `intersect`. The result is that essentially all packages in the set solve, except for three which all do indeed have incorrect ranges. Here's my favorite of the errors:

<details>
<summary>ps-cst@1.2.0</summary>
✗ ps-cst@1.2.0:

Committed to ansi@6.1.0 but the range >=7.0.0 <8.0.0 was also required while solving dodo-printer@2.2.1
Committed to ansi@6.1.0 but the range >=7.0.0 <8.0.0 was also required while solving dodo-printer@2.2.0
Committed to aff in range >=7.0.0 <8.0.0 while solving dodo-printer@2.2.1 but the range >=6.0.0 <7.0.0 was also required while solving dodo-printer@2.1.0
Committed to aff in range >=7.0.0 <8.0.0 while solving dodo-printer@2.2.1 but the range >=6.0.0 <7.0.0 was also required while solving dodo-printer@2.0.0
Committed to effect in range >=3.0.0 <4.0.0 but the range >=4.0.0 <5.0.0 was also required while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to effect in range >=3.0.0 <4.0.0 but the range >=4.0.0 <5.0.0 was also required while solving aff@7.0.0 while solving dodo-printer@2.2.1
Committed to foldable-traversable in range >=5.0.0 <6.0.0 while solving ansi@6.1.0 but the range >=6.0.0 <7.0.0 was also required while solving arrays@7.1.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to foldable-traversable in range >=5.0.0 <6.0.0 while solving ansi@6.1.0 but the range >=6.0.0 <7.0.0 was also required while solving arrays@7.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving const@6.0.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving control@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to foldable-traversable in range >=5.0.0 <6.0.0 while solving ansi@6.1.0 but the range >=6.0.0 <7.0.0 was also required while solving datetime@6.1.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to foldable-traversable in range >=5.0.0 <6.0.0 while solving ansi@6.1.0 but the range >=6.0.0 <7.0.0 was also required while solving datetime@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving either@6.1.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving either@6.0.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving enums@6.0.0 while solving datetime@6.1.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to bifunctors@6.0.0 but the range >=5.0.0 <6.0.0 was also required while solving foldable-traversable@5.0.1 while solving ansi@6.1.0
Committed to bifunctors@6.0.0 but the range >=5.0.0 <6.0.0 was also required while solving foldable-traversable@5.0.0 while solving ansi@6.1.0
Committed to foldable-traversable@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving gen@4.0.0 while solving enums@6.0.0 while solving datetime@6.1.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving invariant@6.0.0 while solving const@6.0.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to bifunctors@6.0.0 but the range >=5.0.0 <6.0.0 was also required while solving lists@6.1.0 while solving ansi@6.1.0
Committed to bifunctors@6.0.0 but the range >=5.0.0 <6.0.0 was also required while solving lists@6.0.1 while solving ansi@6.1.0
Committed to bifunctors@6.0.0 but the range >=5.0.0 <6.0.0 was also required while solving lists@6.0.0 while solving ansi@6.1.0
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving maybe@6.0.0 while solving either@6.1.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to prelude in range >=5.0.0 <6.0.0 while solving console@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving newtype@5.0.0 while solving bifunctors@6.0.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
Committed to aff@7.0.0 but the range >=6.0.0 <7.0.0 was also required while solving node-fs-aff@7.0.0
Committed to foldable-traversable@5.0.0 but the range >=6.0.0 <7.0.0 was also required while solving nonempty@7.0.0 while solving enums@6.0.0 while solving datetime@6.1.0 while solving aff@7.1.0 while solving dodo-printer@2.2.1
</details>

thomashoneyman:
In 822ce43 I allowed for testing the solver against the entire registry index. I can revert the commit if desired. However, it lets us try the solver on _any_ package and see how it does. Perhaps we can use it to narrow down some test cases.

thomashoneyman:
We now also have https://github.com/thomashoneyman/bower-solver-results so we can compare the results of our solver with the Bower solutions for any given package version.

thomashoneyman:
@MonoidMusician I'm going to leave this alone for a while but the tests now:

1. Only test package versions that also have a known Bower solution
2. Only test packages in the core, web, contrib, and node organizations (though it's easy to add or remove these as needed, or filter further to only particular packages).

When a test fails it is because Bower has a solution and we don't. Only a few packages fail, but here's an example of the output:

```
✗ Solves free:

free@2.0.0
----------
Committed to unsafe-coerce in range >=2.0.0 <3.0.0 but the range >=1.0.0 <2.0.0 was also required while solving exists@2.0.0
----------
{
  "arrays": "3.2.1",
  "bifunctors": "2.0.0",
  "catenable-lists": "2.0.0",
  "const": "2.0.0",
  "contravariant": "2.0.0",
  "control": "2.0.0",
  "distributive": "2.0.0",
  "eff": "2.0.0",
  "either": "2.2.1",
  "exists": "2.0.0",
  "foldable-traversable": "2.2.0",
  "functors": "1.1.0",
  "generics": "3.0.0",
  "identity": "2.1.0",
  "inject": "3.0.0",
  "invariant": "2.0.0",
  "lazy": "2.0.0",
  "lists": "2.1.0",
  "maybe": "2.1.1",
  "monoid": "2.2.0",
  "newtype": "1.3.0",
  "nonempty": "3.0.0",
  "partial": "1.2.1",
  "prelude": "2.5.0",
  "proxy": "1.0.0",
  "st": "2.0.0",
  "strings": "2.1.0",
  "tailrec": "2.0.2",
  "transformers": "2.3.0",
  "tuples": "3.2.0",
  "unfoldable": "2.0.0",
  "unsafe-coerce": "2.0.0"
}
```

I hope this gives us a way to test on some real-world cases. Let me know what I can do to make it more helpful!

MonoidMusician:
Wow, thanks for your work! and your CPU cycles!

It seems like `registry-index` and bower are disagreeing on dependency requirements:
- `registry-index` records `exists@2.0.0` as depending on `"unsafe-coerce":">=1.0.0 <2.0.0"`: https://github.com/purescript/registry-index/blob/5ae355628344deb462e1710b7b0e272fad3e852e/ex/is/exists#L5
- but the `bower.json` from `exists@2.0.0` lists `"purescript-unsafe-coerce": "^2.0.0"`: https://github.com/purescript/purescript-exists/blob/v2.0.0/bower.json#L20

I don't really know how it works, but could it be because it's taking dependencies from the package set first? https://github.com/purescript/registry/blob/cfb7697354c2877e9a78439e70aa2ad53ab879a7/ci/src/Registry/Legacy/Manifest.purs#L94

thomashoneyman:
Interesting! That certainly could be a source of issues — in the case of conflict between the three sources (package set vs bower vs spago) we select package sets, then the higher of bower or spago. But perhaps we should select an inclusive range? Or maybe we should even attempt to solve while generating the new manifest, to find the maximum range that works? We can always regenerate all the registry packages from scratch it we want to make changes.

As far as the bower solver results in my repo, however, those are produced by running the Bower solver against the manifest we produced. So if the registry says that exists depends on unsafe-coerce >=1.0.0 <2.0.0, then the Bower file that got solved lists the same dependency range — not whatever the _actual_ Bowerfile for that version of exists was. I did that to try and make the input to the Bower solver and the input to our solver the same.

For example:
https://github.com/purescript/registry/blob/9370499a404b06a133e90f963257797f65c8023e/ci/src/Registry/Scripts/BowerInstaller.purs#L135

Unfortunately, I don’t know of a good way to apply the same technique to the transitive dependencies. When bower solves a dependency and sees it needs to solve a dependency of that dependency, then it’s back to the actual bower.json files instead of the manifests in the registry.

MonoidMusician (approved):

PR #488, opened 2022-08-15
Add script to produce Bower resolutions for registry index

thomashoneyman:
This script can be used to produce Bower resolutions for the entire registry index. The resulting artifacts are stored here:
https://github.com/thomashoneyman/bower-solver-results

We can then use these results in our solver tests, at least for the time being, to discover packages where Bower produces a build plan but our solver does not (or when Bower produces one with higher versions). When we're confident in the solver we can either retain this script and repository of solutions or we can archive them.

cc: @MonoidMusician

### DIFF
--- a/ci/src/Registry/Scripts/BowerInstaller.purs
+++ b/ci/src/Registry/Scripts/BowerInstaller.purs
@@ -1,0 +1,168 @@
+-- | This script attempts to install all packages in the registry index using
+-- | Bower, recording which packages have a successful run and which packages
+-- | fail to install. This is used to generate the list of Bower packages that
+-- | succeed in the integration tests for the registry solver.
+module Registry.Scripts.BowerInstaller where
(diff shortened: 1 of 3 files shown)


ci/src/Registry/Scripts/BowerInstaller.purs
```diff
@@ -0,0 +1,216 @@
...
+
+      liftEffect (Cache.writeJsonEntry key parsed cache)
+      pure parsed
+
+    Right { value } ->
+      pure value
+
+-- case value of
+--   Nothing -> liftEffect (cache.remove key) *> runBowerInstall name version contents
+--   otherwise -> pure otherwise
```
thomashoneyman: Whoops! No need for this.

```suggestion
```

MonoidMusician (approved):

PR #496, opened 2022-08-24
Errors

MonoidMusician:
Not sure if this actually does anything! But it looks like it should.


MonoidMusician:
With the `recover` parameter, I'm hoping to reduce the number of errors shown for each particular version, since it won't be recovering from those errors. (You'll still have multiple errors since the alternatives all need to fail.)

But the diff on the integration is weird since it shows more errors and changed errors:
```diff
diff --git a/test1.txt b/test2.txt
index 823b5ce..1f31004 100644
--- a/test1.txt
+++ b/test2.txt
@@ -22,6 +22,9 @@
 
   free@2.0.0
 ----------
+Committed to tuples in range >=3.0.0 <4.0.0 while solving catenable-lists@2.0.0 but the range >=2.0.0 <3.0.0 was also required while solving inject@3.0.0
+Committed to tuples in range >=3.0.0 <4.0.0 while solving catenable-lists@2.0.0 but the range >=2.0.0 <3.0.0 was also required while solving const@2.0.0 while solving inject@3.0.0
+Committed to tuples in range >=3.0.0 <4.0.0 while solving catenable-lists@2.0.0 but the range >=2.0.0 <3.0.0 was also required while solving contravariant@2.0.0 while solving inject@3.0.0
 Committed to unsafe-coerce in range >=2.0.0 <3.0.0 but the range >=1.0.0 <2.0.0 was also required while solving exists@2.0.0
 ----------
 {
@@ -154,10 +157,10 @@ Committed to unsafe-coerce in range >=1.0.0 <2.0.0 but the range >=2.0.0 <3.0.0
 
   argonaut@5.0.0
 ----------
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0
 Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.0.0 <2.0.0 was also required while solving argonaut-core@4.0.1
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0
 ----------
 {
   "argonaut-codecs": "5.1.3",
@@ -307,11 +310,11 @@ Committed to record in range >=1.0.0 <2.0.0 but the range >=2.0.0 <3.0.0 was als
 
   argonaut-traversals@5.0.0
 ----------
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.0.0
 Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.0.0 <2.0.0 was also required while solving argonaut-core@4.0.1
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.0.0
 ----------
 {
   "argonaut-codecs": "5.1.3",
@@ -366,6 +369,8 @@ Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codec
   arraybuffer@11.0.0
 ----------
 Committed to control@4.1.0 but the range >=5.0.0 <6.0.0 was also required while solving maybe@5.0.0
+Committed to maybe@5.0.0 but the range >=4.0.1 <5.0.0 was also required while solving uint@5.1.4
+Committed to partial@3.0.0 but the range >=2.0.1 <3.0.0 was also required while solving typelevel@6.0.0
 Committed to maybe in range >=5.0.0 <6.0.0 but the range >=4.0.0 <5.0.0 was also required while solving generics-rep@6.1.4 while solving float32@0.2.0
 Committed to maybe in range >=5.0.0 <6.0.0 but the range >=4.0.0 <5.0.0 was also required while solving generics-rep@6.1.2 while solving float32@0.2.0
 Committed to maybe in range >=5.0.0 <6.0.0 but the range >=4.0.0 <5.0.0 was also required while solving gen@2.1.0 while solving enums@4.0.1 while solving generics-rep@6.1.3 while solving float32@0.2.0
@@ -523,10 +528,10 @@ Committed to unsafe-coerce in range >=1.0.0 <2.0.0 while solving exists@2.0.0 wh
 
   uint@4.1.0
 ----------
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2 while solving argonaut@5.0.0
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1 while solving argonaut@5.0.0
-Committed to foreign-object@2.0.3 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0 while solving argonaut@5.0.0
 Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 while solving argonaut@5.0.0 but the range >=1.0.0 <2.0.0 was also required while solving argonaut-core@4.0.1 while solving argonaut@5.0.0
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 while solving argonaut@5.0.0 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.2 while solving argonaut@5.0.0
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 while solving argonaut@5.0.0 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.1 while solving argonaut@5.0.0
+Committed to foreign-object in range >=2.0.0 <3.0.0 while solving argonaut-codecs@5.1.3 while solving argonaut@5.0.0 but the range >=1.1.0 <2.0.0 was also required while solving argonaut-codecs@5.1.0 while solving argonaut@5.0.0
 ----------
 {
   "argonaut": "5.0.0",

```

MonoidMusician:
Last thoughts before I go to bed:

Besides just grouping errors, perhaps a more high-level approach would be to gather compatibilities and conflicts.

Maybe we don't need compatibilities. The `SolverPosition` (which would give us a sense of compatibilities/requirements in errors) is just to give the user a chance to see where a package was required. But it really should be a breadth-first search, not depth-first. We could probably even reconstruct it after the fact with a BFS. (In particular, if a package is a direct dependency, you never want to show a path to it. The difficulty here is of course the versions …)

Conflicts would be less structured than they are now, more global. It would remove some weird asymmetries/nondeterminism in the error reporting. It would probably just be a map of packages and version ranges that were known to conflict at some point, like “You cannot have X@1.0.0 and Y@1.0.0-1.1.0 because they conflict at Z”. Is that much different than it is now? Maybe not. It suggests an almost geometrical approach to it though (in the sense of combinatorial geometry, not 3D spatial geometry).

MonoidMusician:
1. Good errors will probably take forever to run, do we care?
2. Should we just add an `Ord Range` instance so I don't have to do `compare (printRange r1) (printRange r2)`?

MonoidMusician:
Currently takes forever and OOMs on `purescript-foreign` in the (old) integration test …
It still takes a long time (and fails) if I rip out the errors, but it doesn't OOM then …

f-f:
> Should we just add an `Ord Range`

How should it behave? E.g. what's the result for `">1.0.0 <=2.1.0" >= ">1.1.0 <=2.0.0"`?

> Good errors will probably take forever to run, do we care?

We care in the sense that the solver should generally be fast, but it's fine if tests take a little while.

MonoidMusician:
The behavior of `Ord Range` doesn’t matter, just something to compare on for grouping/nubbing things.

MonoidMusician:
Okay, I got the integration test to work by deduplicating early and deduplicating often, so less objects are stuck in memory. It took about 2.25 hours for the core sets since maybe 1/3 of them fail to have solutions against the old integration test.

f-f:
> The behavior of `Ord Range` doesn’t matter, just something to compare on for grouping/nubbing things

It does, as we'll expose the `Range` type for external use, so we wouldn't want an "incorrect" instance.

Is "wrapping in a newtype" a viable option here?

MonoidMusician:
Eh I'll end up using my own custom type when I rewrite it so nevermind!

f-f:
> It took about 2.25 hours

That's a long time, but I guess solving everything is a lot of work - what does the performance look like to solve one package? (e.g. something big like Halogen?)

MonoidMusician:
Foreign still takes a long time but it produces reasonable errors now (again, on the old integration tests – the new registry is much quicker with fewer errors!):
```
Committed to foldable-traversable@2.0.0 but the range >=2.1.0 <3.0.0 was also required while solving generics@3.3.0
Committed to foldable-traversable@2.0.0 but the range >=2.1.0 <3.0.0 was also required while solving lists@3.4.0
Committed to foldable-traversable@2.0.0 but the range >=2.1.0 <3.0.0 was also required while solving transformers@2.2.0
Committed to foldable-traversable@2.0.0 but the range >=2.1.0 <3.0.0 was also required while solving tuples@3.1.0
Committed to generics@3.2.0 but the range >=3.3.0 <4.0.0 was also required while solving lists@3.4.0
Committed to newtype@1.1.0 but the range >=1.2.0 <2.0.0 was also required while solving transformers@2.2.0
Committed to newtype@1.1.0 but the range >=1.2.0 <2.0.0 was also required while solving tuples@3.1.0
Committed to prelude@2.1.0, 2.2.0 but the range >=2.3.0 <3.0.0 was also required while solving transformers@2.2.0
Committed to prelude@2.1.0, 2.2.0 but the range >=2.3.0 <3.0.0 was also required while solving tuples@3.1.0
Committed to prelude@2.1.0, 2.2.0, 2.3.0 but the range >=2.4.0 <3.0.0 was also required while solving tuples@3.2.0
Committed to prelude@2.3.0 but the range >=2.4.0 <3.0.0 was also required while solving tuples@3.2.0 while solving generics@3.3.0
Committed to prelude@2.3.0 but the range >=2.4.0 <3.0.0 was also required while solving tuples@3.2.0 while solving transformers@2.2.0
Committed to tuples@3.0.0, 3.1.0, 3.2.0 but the range >=2.0.0 <3.0.0 was also required while solving unfoldable@2.0.0
```

MonoidMusician:
Wow, catenable-lists went down from 42 -> 14 errors with the latest change! (`minimizeSolverPositions`)

I think I can fix some of the speed issues but it'll take a few days. And I'm convinced that I want to rewrite it to BFS eventually.

src/Registry/Solver.purs
```diff
@@ -36,17 +41,50 @@ instance Show SolverPosition where
...
+  toBeFound = errored # foldMapWithIndex (Set.map <<< Tuple)
+
+  go
+    :: Map (Tuple PackageName Version) (First SolverPosition)
+    -> Map PackageName (NonEmptyArray (Tuple SolverPosition Range))
+    -> Map (Tuple PackageName Version) (First SolverPosition)
+  -- Found enough, go home
+  go found _ | toBeFound `Set.subset` Map.keys found = found
+  -- Add relevant things from `currentGoals`, go to next layer
+  go alreadyFound currentGoals =
```
thomashoneyman, line 67: Minor stylistic comment -- we use `do` / `let` everywhere else in the codebase; do you mind doing the same here?

thomashoneyman (approved):

PR #503, opened 2022-08-31
Fix duplicate module compilation failure

### LINKED ISSUE #497: Package set updater throws on duplicate module name
See:
https://github.com/purescript/registry-preview/actions/runs/2893901334

This shouldn’t terminate the import; it should just remove the duplicate from the package set batch.

thomashoneyman:
When the compiler fails to build a package because of a duplicate module, the resulting error has no module name set. Our JSON decoding for these errors always expected a module name to be present, so instead of reporting a compilation failure we'd get an "unknown compiler error" and crash.

Here's what the JSON looks like for a duplicate module error:

```json
{
  "warnings": [],
  "errors": [
    {
      "position": {
        "startLine": 1,
        "startColumn": 1,
        "endLine": 1,
        "endColumn": 18
      },
      "message": "  Module Main has been defined multiple times\n",
      "errorCode": "DuplicateModule",
      "errorLink": "https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md",
      "filename": "src/Main.purs",
      "moduleName": null,
      "suggestion": null
}
```

I've tested out this change, and duplicate module errors now properly report as compilation failures.

### DIFF
--- a/src/Foreign/Purs.purs
+++ b/src/Foreign/Purs.purs
@@ -91,7 +91,7 @@ callCompiler compilerArgs = do
     Right { exit: NodeProcess.Normally 0, stdout } -> Right $ String.trim stdout
     Right { stdout } -> Left do
       case Json.parseJson (String.trim stdout) of
-        Left err -> UnknownError stdout
+        Left err -> UnknownError $ String.joinWith "\n" [ stdout, err ]
         Right ({ errors } :: { errors :: Array CompilerError })
           | Array.null errors -> UnknownError "Non-normal exit code, but no errors reported."
           | otherwise -> CompilationError errors
--- a/src/Foreign/Purs.purs
+++ b/src/Foreign/Purs.purs
@@ -3,6 +3,7 @@ module Foreign.Purs where
 import Registry.Prelude
 
 import Data.Array as Array
+import Data.Foldable (foldMap)
 import Data.String as String
 import Effect.Exception as Exception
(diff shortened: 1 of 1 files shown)


src/Foreign/Purs.purs
```diff
@@ -90,7 +91,7 @@ callCompiler compilerArgs = do
     Right { exit: NodeProcess.Normally 0, stdout } -> Right $ String.trim stdout
     Right { stdout } -> Left do
       case Json.parseJson (String.trim stdout) of
-        Left err -> UnknownError $ String.joinWith "\n" [ stdout, err ]
+        Left err -> UnknownError stdout
```
thomashoneyman:
```suggestion
        Left err -> UnknownError $ String.joinWith "\n" [ stdout, err ]
```

colinwahl (approved):

src/Foreign/Purs.purs
```diff
@@ -52,7 +53,7 @@ printCompilerErrors errors = do
   printCompilerError :: CompilerError -> String
   printCompilerError { moduleName, filename, message, errorLink } =
     String.joinWith "\n"
-      [ "  Module: " <> moduleName
+      [ foldMap (\name -> "  Module: " <> name) moduleName
```
colinwahl: Actually, will this result in a leading newline when there is no module name? Maybe we don't want that?

PR #504, opened 2022-08-31
Deterministic tarballs

### LINKED ISSUE #500: Deterministic tarballs
Uploading tarballs to the registry is currently non-reproducible because file modification times (among other things) are included in the tar / gzip process. There is no way to make deterministic tarballs via the `node-tar` library ([the solution in the issue for it does not work](https://github.com/npm/node-tar/issues/176); we tried it and had to revert the commit).

We'd like to be able to regenerate the registry with all the metadata hashes remaining the same, at least for the duration of the alpha. Otherwise, changing something about the manifest format or files we include in the tarballs will cause all hashes to change.

Some notes on reproducible builds:
- https://reproducible-builds.org/docs/archives
- https://superuser.com/questions/705877/compressing-compressed-tar-gz-files-deterministically
- https://stackoverflow.com/questions/32997526/how-to-create-a-tar-file-that-omits-timestamps-for-its-contents/54908072#54908072

I would recommend that we rely on `tar` and `gzip` via the Nix shell so that we can use the recommended approaches from Debian et al.

thomashoneyman:
Fixes #500 following [this advice on reproducible tarballs](https://reproducible-builds.org/docs/archives/).


src/Foreign/Tar.purs
```diff
@@ -14,20 +16,30 @@ getToplevelDir filename = do
...
+type ExtractArgs = { cwd :: String, archive :: String }
 
--- | Extracts the tarball at the given filename into cwd.
--- |
--- | Note: `filename` should be an absolute path. The extracted result will be
--- | a directory within `cwd`.
+-- | Extracts the tarball at the given relative file path into cwd.
 extract :: ExtractArgs -> Effect Unit
-extract { cwd, filename } = runFn2 extractImpl cwd filename
+extract { cwd, archive } = do
```
thomashoneyman, line 23: `extract` and `create` now operate on paths relative to the `cwd`. If that's a problem, I can make them always absolute paths instead (and omit the `cwd` argument, as it would be unnecessary).

colinwahl (approved):
Are any of these `archive` or `folderName` ever open ended inputs from the user?

thomashoneyman:
No, we never allow a user to generate tarballs. Though perhaps it's worth adding a warning around this.

PR #505, opened 2022-08-31
Update integration test to include strict Bower comparison

### LINKED ISSUE #490: Add strict Bower integration test for solver
When implementing an integration test for the new registry solver in #487, @MonoidMusician noticed that a number of packages have a disagreement between the ranges reported in their manifest in the registry index and the ranges reported in their Bowerfiles. That is causing the integration test to report that our solver can't solve some packages that the Bower solver can, though it's actually because the ranges are different.

A better head-to-head integration test would strictly look at the Bower files. To that end, we need to change the `BowerInstaller` script so that it produces the following output for each package version:

1. The parsed dependencies and dependency ranges for every dependency listed in the Bower file, in a format suitable for the solver (ie. using the `PackageName` and `Version` and `Range` types). Anything that doesn't parse should be omitted.
2. The Bower solution for the registry manifest (this is what already exists in the integration test)
3. The Bower solution for the Bowerfile only

This can be stored at https://github.com/thomashoneyman/bower-solver-results, or that can be transferred to the `purescript` org so we can all access it, alternately. Then, these cached results can be reused so that we can run our own solver and see how it compares to Bower directly.

thomashoneyman:
cc: @MonoidMusician -- this fixes #490 by adding a strict comparison to Bower's solution on its Bowerfile against the solver's solution on the parsed Bowerfile. I've also retained the solver running on the manifest file, but we can disable that if you'd like.

### DIFF
--- a/test/Integration.purs
+++ b/test/Integration.purs
@@ -1,26 +1,25 @@
-module Test.Integration where
+module Test.Integration
+  ( main
+  ) where
 
 import Registry.Prelude
 
 import Control.Monad.Except as Except
 import Control.Monad.State as State
 import Data.Array.NonEmpty as NEA
-import Data.Filterable (filterMap)
-import Data.Foldable (foldl)
 import Data.Map as Map
 import Data.String as String
 import Effect.Exception as Exception
 import Foreign.Git as Git
 import Foreign.Tmp as Tmp
-import Node.FS.Aff as FS.Aff
 import Node.Path as Path
-import Parsing as Parsing
 import Registry.Index (RegistryIndex)
 import Registry.Index as Index
 import Registry.Json as Json
 import Registry.PackageName (PackageName)
 import Registry.PackageName as PackageName
 import Registry.Schema (Location(..), Manifest(..))
+import Registry.Scripts.BowerInstaller (BowerSolved)
 import Registry.Solver as Solver
 import Registry.Version (Range, Version)
 import Registry.Version as Version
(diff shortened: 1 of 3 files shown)


thomashoneyman:
As usual, runnable with `spago test -m Test.Integration`.

test/Integration.purs
```diff
@@ -31,130 +30,114 @@ import Test.Spec.Runner (defaultConfig, runSpec')
...
+  pure { solverIndex, solutions: segmentedIndex }
+
+mkTest :: Solver.Dependencies -> Map PackageName (Map Version { bower :: BowerSolved, manifest :: Map PackageName Range }) -> Spec.Spec Unit
+mkTest solverIndex pkgs = void $ forWithIndex pkgs \package versions -> do
+  Spec.describe ("Solves " <> PackageName.print package) do
+    -- We try comparing Bower's solution from the Bowerfile to our
+    -- solver's attempt on the Bowerfile dependencies.
+    Spec.it "Bowerfile" do
+      void $ forWithIndex versions \version { bower } -> do
+        solve package version bower.dependencies bower.bowerfileSolution
```
thomashoneyman, line 111: This test is the solver on the Bowerfile dependencies vs. the Bowerfile solution.

test/Integration.purs
```diff
@@ -31,130 +30,114 @@ import Test.Spec.Runner (defaultConfig, runSpec')
...
+
+    -- We also try comparing Bower's solution using the manifest dependencies
+    -- to our solver's attempt on the same.
+    Spec.it "Manifest file " do
+      void $ forWithIndex versions \version { bower, manifest } -> do
+        solve package version manifest bower.manifestSolution
```
thomashoneyman, line 117: This test is the solver on the manifest dependencies vs. the Bower solution on the manifest.

test/Integration.purs
```diff
@@ -31,130 +30,114 @@ import Test.Spec.Runner (defaultConfig, runSpec')
...
+        let printedErrs = String.joinWith "\n" $ NEA.toArray $ map Solver.printSolverError errs
+        Assert.fail $ String.joinWith "\n----------\n"
+          [ name
+          , printedErrs
+          , Json.printJson solution
+          ]
+      -- If we found a solution then we passed the test.
+      -- TODO: We can also check that our solution produces versions as
+      -- high as those produced by Bower, if we want.
+      Right _ -> pure unit
```
thomashoneyman, line 143: I'm assuming we can do this later. But we've got all the information we need .

f-f (approved):

PR #506, opened 2022-09-01
Rework version ranges

### LINKED ISSUE #491: Reconsider how we produce version ranges in the registry importer
While working on the dependency solver, the integration tests revealed some incoherent ranges being produced by the registry importer. For example:

https://github.com/purescript/registry/pull/487#issuecomment-1217390749

The code that decides what dependencies and version ranges to apply for a newly-generated manifest is located here:

https://github.com/purescript/registry/blob/cfb7697354c2877e9a78439e70aa2ad53ab879a7/ci/src/Registry/Legacy/Manifest.purs#L54-L109

Specifically, we always take the package set dependencies if they exist, or we take whichever of the bower.json or spago.dhall dependencies report higher versions:

https://github.com/purescript/registry/blob/cfb7697354c2877e9a78439e70aa2ad53ab879a7/ci/src/Registry/Legacy/Manifest.purs#L90-L94

Unfortunately, these files all fall out sync with each other in one way or another. Relying on any of these manifests has issues:

1. The package sets are machine-checked to verify that the listed dependencies and version compile together. However, when a new version of a package is released (like `exists@2.0.0`), it may be updated in the package sets without its dependencies being updated, if those dependencies still work. This is how you have `exists@2.0.0` listing a dependency on `unsafe-coerce@^2.0.0` in its bower file, but the package sets continuing to list it with a dependency on `unsafe-coerce@1.0.0` (unsafe-coerce works either way).
2. The bower.json files give us ranges to work with, but there are many projects in the ecosystem that switched to Spago and haven't kept their bower.json files up to date even though they're still in the repository. (Or they forget until they've already tagged the release, and then have to update the bower.json in a patch release so they can publish docs.) This can lead the bower.json files to report out-of-date, old versions of packages that don't actually work.
3. The spago.dhall files suffer from a similar problem. Sometimes they aren't kept up-to-date (ie. a project added them, but then realized they can't publish to Pursuit with them, so ended up sticking with Bower), or they point to package sets that no longer exist like the `prepare-0.15` package set. They often list way more dependencies than are actually necessary (ie. all the test dependencies).

All three manifests are problematic. We need _some_ way to resolve these issues to try and produce manifests that will work in practice when they are used in the wild. There are some ideas:

1. Stop preferring the package sets. Treat the package sets, spago.dhall file, and bower.json files with equal weight when determining package dependency versions.
2. In terms of deciding what dependency _names_ to consider, we use the a) package sets, then b) bower.json file, then c) spago.dhall file. The package sets are machine-checked, which gives them the highest chance to be listing the correct dependencies. The bower.json file separates dependencies and devDependencies, which makes it better than the spago.dhall file, where people just shove their library and test dependencies together.
3. In terms of deciding what dependency _versions_ to consider, we could take the intersection of dependencies from all three sources. Then, for each dependency that intersects, should take the _most lenient_ range possible: find the lowest dependency bound and the highest dependency bound and merge them. For example, if `exists@2.0.0` has `unsafe-coerce@^1.0.0` from the package sets and `unsafe-coerce@^2.0.0` from its Bower file, then we should assign it with `unsafe-coerce@>=1.0.0 <3.0.0`.

As far as testing our changes goes, we can continue to rely on the solver integration test to find packages that don't solve due to incompatible ranges and look closely at those.

cc: @MonoidMusician @colinwahl – love to hear y'all's ideas as well!

thomashoneyman:
Fixes #491. Specifically, we now:

1. Prioritize the package sets, then the bower.json file, and finally the spago.dhall file when determining what dependencies the project is using.
2. Take the union of the version ranges represented in all three files, rather than prioritize any given one.

I've run this to regenerate the registry index, which can be seen in this PR:
https://github.com/purescript/registry-index/pull/4

On review, these changes look good and properly pick up some of the wide (but acceptable) ranges represented in packages like `unsafe-coerce` which work with multiple compiler versions.

### DIFF
--- a/src/Registry/Legacy/Manifest.purs
+++ b/src/Registry/Legacy/Manifest.purs
@@ -6,7 +6,6 @@ import Control.Monad.Except as Except
 import Control.Monad.Reader (ask)
 import Data.Array as Array
 import Data.Either as Either
-import Data.Foldable (sum)
 import Data.FunctorWithIndex (mapWithIndex)
 import Data.Map as Map
 import Data.String as String
@@ -58,7 +57,7 @@ fetchLegacyManifest packageSetsDeps address ref = do
       -- "strings: >=4.1.2 <4.1.3", which while technically correct is almost
       -- certainly not what the user wants. We instead treat these ranges as
       -- caret ranges, so "strings: 4.1.2" becomes "strings: >=4.1.2 <5.0.0".
-      toRange (RawVersion fixed) = do
+      fixedToRange (RawVersion fixed) = do
         let parsedVersion = Version.parseVersion Version.Lenient fixed
         let bump version = Version.printVersion (Version.bumpHighest version)
         let printRange version = Array.fold [ ">=", fixed, " <", bump version ]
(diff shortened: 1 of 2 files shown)


MonoidMusician (approved):
To elaborate a bit on what we've discussed behind the scenes: these “lenient” version ranges may be inaccurate in some cases, but that's okay for a couple reasons:
1. These ranges are only for legacy packages, and we aren't expecting to be the exact source of truth about what does or does not compile wrt those.
2. Outright incompatibilities of loose upper bounds seem unlikely, given where this data is coming from.
3. This change mainly loosens lower bounds anyways, and having these loose lower bounds is probably okay, since package managers will **prefer later versions** when possible.

Most importantly, we want to ensure that if they built at one time, they will continue to be usable as legacy package sets with future package managers. Before this change some packages had dependency ranges from legacy import that were unable to be solved as a whole, which would have posed a problem for new packages coming into the package set if they could not depend on that version of that package.

PR #507, opened 2022-09-01
Implement legacy package set mirroring

thomashoneyman:
This implements legacy package set mirroring so that we can officially deprecate the package-sets publishing process in the `purescript/package-sets` repository. The implementation largely replicates these two scripts:

* https://github.com/purescript/package-sets/blob/psc-0.15.4-20220829/latest-compatible-sets.json
* https://github.com/purescript/package-sets/blob/psc-0.15.4-20220829/release.sh

We will retain the [`release.yml` file that builds release assets](https://github.com/purescript/package-sets/blob/9139d7ee3992040ebf66e4d24d437574b6e9b58c/.github/workflows/release.yml), so that hasn't been included in this PR.

When a new package set is produced via the daily package sets _or_ a manual release, then we convert it to the legacy format and:

* If there has already been a package sets release that day, nothing is mirrored to the legacy sets (in keeping with the existing implementation).
* If there has not been a package sets release, then we write the `latest-compatible-sets.json` file with the new tag, write the `src/packages.dhall` file in Dhall, and write the `packages.json` file in JSON. The three files are committed. Then, we tag the commit with `psc-COMPILER-DATE`. We also tag it with `psc-COMPILER` if this is the first compiler version in the series. The commit and tags are all pushed.

At that point the `release.yml` file picks it up and makes the GitHub release with the proper assets. Therefore, all files in the package sets repo can be removed except:

* `.github/workflows/release.yml`
* `packages.json`
* `latest-compatible-sets.json`
* `src/packages.dhall`
* `default.nix`

(as well as the README and LICENSE files, which aren't directly relevant to the code here).

### DIFF
--- a/src/Foreign/Git.purs
+++ b/src/Foreign/Git.purs
@@ -2,8 +2,17 @@ module Foreign.Git where
 
 import Registry.Prelude
 
+import Affjax as Http
+import Control.Monad.Except as Except
+import Data.JSDate as JSDate
+import Data.PreciseDateTime as PDT
+import Data.RFC3339String (RFC3339String)
 import Data.String as String
+import Effect.Aff as Aff
+import Effect.Exception as Exception
+import Foreign.GitHub as GitHub
 import Node.ChildProcess as NodeProcess
+import Node.Process as Env
 import Sunde as Process
 
 runGit_ :: Array String -> Maybe FilePath -> ExceptT String Aff Unit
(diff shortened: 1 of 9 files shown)


src/Foreign/Git.purs
```diff
@@ -30,3 +39,32 @@ runGitSilent args cwd = ExceptT do
       let stdout = String.trim result.stdout
       pure $ Right stdout
     _ -> pure $ Left $ "Failed to run git command via runGitSilent."
+
+-- | Clone a package from a Git location to the provided directory.
```
thomashoneyman, line 43: These were moved from the API file to avoid circular dependencies.

thomashoneyman:
@f-f @colinwahl If merged as-is, this will begin mirroring to the package sets repository right away. Presumably we don't want that: at the very least, we need to do some preparatory steps:

1. We need to write a short user guide for the `package-sets` repo that says to go over to the registry to register your package, and retains the existing "use this with Spago, use this with psc-package" documentation
2. We need to update the `registry-preview` repository to delete all existing package sets and begin with the same latest package set that is listed in the package-sets repo (they've diverged over time)
3. We _may_ wish to go ahead and reorganize the repositories so we're directing folks to the `registry` repo, not the `registry-preview` repo.

Moving repos around will be a bit dodgy, so if this PR looks good in general then perhaps you and I can set up a call @f-f where we do it together and fix any issues that arise?

test/Registry/PackageSet.purs
```diff
@@ -31,8 +32,22 @@ spec = do
   Spec.it "Encodes package set" do
     Json.printJson packageSet `Assert.shouldEqual` packageSetJson
 
+  Spec.it "Parses legacy package set tags" do
+    let valid = "psc-0.12.18-20220102"
```
f-f, line 36: Not sure if you went far back enough to find one, but when publishing more than once per day we had package set names like `psc-0.12.18-20220102-1`. Might be worth adding that to the parser?

f-f:
@thomashoneyman yes this sounds great!
I think we should first move the repos, then in order to try this out without potentially disrupting things we could fork the package-sets repo, and point this to the fork at first, then move it to the "production" repo after everything else (e.g. documentation) is in place.

test/Registry/PackageSet.purs
thomashoneyman: I skipped those simply because we won't ever encounter one; we just parse and print legacy tags for the `latest-compatible-sets`, and those never have the suffix. Does that seem OK to you?

thomashoneyman:
@f-f Good idea. I've created the fork here:
https://github.com/purescript/package-sets-preview

test/Registry/PackageSet.purs
f-f: Yeah sounds great!

thomashoneyman:
@f-f this is now ready to go; we _read_ from the old package-sets repo, but we _write_ to the package-sets-preview repo. When it comes time to mirror back to the actual package-sets repo then we can just replace the TODO comment with the constant.

f-f (approved):

PR #510, opened 2022-09-02
Make repo urls configurable

thomashoneyman:
This PR exposes the various repositories and file paths we use for the registry via a Constants module (one that we can add other things to, such as the S3 bucket location, so that package managers can reuse them). When it comes time to do the repository migration we'll just have to update the Constants module.

### DIFF
--- a/src/Registry/Constants.purs
+++ b/src/Registry/Constants.purs
@@ -1,0 +1,31 @@
+module Registry.Constants
+  ( legacyPackageSetsRepo
+  , metadataPath
+  , packageSetsPath
+  , registryIndexRepo
+  , registryPackagesUrl
+  , registryRepo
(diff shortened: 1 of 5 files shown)


src/Registry/Constants.purs
```diff
@@ -0,0 +1,31 @@
...
+  , packageSetsPath
+  , registryIndexRepo
+  , registryPackagesUrl
+  , registryRepo
+  ) where
+
+import Affjax as Http
+import Node.Path (FilePath)
+
+type Repository = { owner :: String, repo :: String }
```
thomashoneyman, line 13: Can't have a circular dependency with the `Foreign.GitHub` module!

f-f (approved):

PR #512, opened 2022-09-06
Make build plan resolutions optional

thomashoneyman:
This PR makes build plan resolutions optional. In practice, that means that when registering a package you can either:

1. Provide a compiler version and resolved dependencies the registry should use to verify your package and push documentation
2. Provide only the compiler version, and rely on the registry solver to figure out the resolved dependencies

Optional build plans mean that it is feasible to register a package by hand. Since we'd like to move the `new-packages.json` and `bower-packages.json` files into the registry, and disable PRs there, that means people will now need to start registering their packages by hand.

A future improvement we can make is to take the build plan from Spago directly, if it's a Spago package being published. That may make things a little more reliable in the case that folks' dependency ranges are off.

### DIFF
--- a/src/Foreign/Dhall.purs
+++ b/src/Foreign/Dhall.purs
@@ -6,18 +6,6 @@ import Node.ChildProcess as NodeProcess
 import Registry.Json as Json
 import Sunde as Process
 
--- | Attempt to convert a JSON file representing a PureScript manifest into
--- | the corresponding `Manifest.dhall` type using the `json-to-dhall` CLI.
-jsonToDhallManifest :: String -> Aff (Either String String)
-jsonToDhallManifest jsonStr = do
-  let cmd = "json-to-dhall"
-  let stdin = Just jsonStr
-  let args = [ "--records-loose", "--unions-strict", "./v1/Manifest.dhall" ]
-  result <- Process.spawn { cmd, stdin, args } NodeProcess.defaultSpawnOptions
-  pure $ case result.exit of
-    NodeProcess.Normally 0 -> Right jsonStr
-    _ -> Left result.stderr
-
 -- | Convert a string representing a Dhall expression into JSON using the
 -- | `dhall-to-json` CLI.
 dhallToJson :: { dhall :: String, cwd :: Maybe FilePath } -> Aff (Either String Json.Json)
--- a/src/Foreign/Dhall.purs
+++ b/src/Foreign/Dhall.purs
@@ -6,6 +6,16 @@ import Node.ChildProcess as NodeProcess
 import Registry.Json as Json
 import Sunde as Process
 
+jsonToDhallManifest :: String -> Aff (Either String String)
+jsonToDhallManifest jsonStr = do
+  let cmd = "json-to-dhall"
+  let stdin = Just jsonStr
+  let args = [ "--records-loose", "--unions-strict", "./v1/Manifest.dhall" ]
+  result <- Process.spawn { cmd, stdin, args } NodeProcess.defaultSpawnOptions
+  pure $ case result.exit of
+    NodeProcess.Normally 0 -> Right jsonStr
+    _ -> Left result.stderr
+
 -- | Convert a string representing a Dhall expression into JSON using the
 -- | `dhall-to-json` CLI.
 dhallToJson :: { dhall :: String, cwd :: Maybe FilePath } -> Aff (Either String Json.Json)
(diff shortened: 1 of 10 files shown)


src/Foreign/Purs.purs
```diff
@@ -65,14 +65,24 @@ printCompilerErrors errors = do
...
   { version :: String
   , cwd :: Maybe FilePath
-  , args :: Array String
+  , command :: PursCommand
   }
 
+data PursCommand
+  = Version
+  | Compile { globs :: Array FilePath }
+  | Publish { resolutions :: FilePath }
```
thomashoneyman, line 74: I did this because it turns out that using `--json-errors` with the `publish` subcommand will error out.

src/Foreign/Dhall.purs
```diff
@@ -6,18 +6,6 @@ import Node.ChildProcess as NodeProcess
 import Registry.Json as Json
 import Sunde as Process
 
--- | Attempt to convert a JSON file representing a PureScript manifest into
--- | the corresponding `Manifest.dhall` type using the `json-to-dhall` CLI.
-jsonToDhallManifest :: String -> Aff (Either String String)
-jsonToDhallManifest jsonStr = do
```
thomashoneyman: We don't need this because we have explicit JSON parsing for manifests, which we have to run anyway. Using this just incurs a needless dependency on the Dhall executable in the API pipeline.

src/Registry/API.purs
```diff
@@ -111,7 +113,14 @@ main = launchAff_ $ do
         fetchRegistry
         fetchRegistryIndex
         fillMetadataRef
-        runOperation operation
+        runOperation API operation
+
+-- | Operations are exercised via the API and the importer. If the importer is
+-- | used, then we have relaxed restrictions on the build (ie. the package does
+-- | not have to compile).
+data Source = API | Importer
```
thomashoneyman, line 124: We need a way to distinguish between a package being uploaded via the API (or us pretending it was) and a package that comes via the importer. For example, we don't require that packages coming in via the importer compile and we don't push their docs. 

I've set things up so that legacy packages sent via the API are compiled, but if the compilation fails we just log it and still publish the package.

src/Registry/API.purs
```diff
@@ -551,15 +570,13 @@ addOrUpdate { updateRef, buildPlan, packageName } inputMetadata = do
       ]
     Right _ -> pure unit
 
-  -- We don't upload legacy packages to Pursuit because we don't have a build
-  -- plan for them. If we want to upload them, we have to produce a build plan.
-  unless isLegacyImport do
+  unless (source == Importer || isLeft compilationResult) do
```
thomashoneyman, line 598: With this change, legacy packages published via the API will have their docs published if they can be compiled. :tada:

thomashoneyman:
As far as testing goes, this can be _somewhat_ exercised by running `spago run -m PublishPursuit`.

src/Foreign/Dhall.purs
f-f: This code is not much about "JSON parsing", rather than about "typecheck this JSON file against this Dhall schema".
I think it's useful to have this running all the time if we think that the Dhall type should be the source of truth (rather than the PureScript type/parsing logic), otherwise we'd risk divergence from the schema in the spec.
The runtime dependency is very cheap because this is code that is only run on the CI pipeline and not e.g. in the library exposed to users.

src/Registry/API.purs
f-f: > I've set things up so that legacy packages sent via the API are compiled, but if the compilation fails we just log it and still publish the package.

Why so?

f-f: Ah, this is because we always the package to be published but it's alright if it doesn't get its docs published right?

src/Foreign/Dhall.purs
thomashoneyman: I’ll revert this. I’d like to talk about our use of Dhall files as the “spec” another time, though!

src/Registry/API.purs
thomashoneyman: Right, for legacy packages we have never enforced that it compiles (though we _could_, and we could regenerate the registry only taking packages that compile). So we continue that policy here. Legacy package = no forced compilation check. But if it does compile then we want to publish the docs.

src/Foreign/Dhall.purs
f-f: Sure thing!

src/Registry/API.purs
f-f: Right, makes sense. I'd say it's worth expanding on the comment above to make this clearer

f-f (approved):

PR #513, opened 2022-09-06
Change S3 location naming scheme

thomashoneyman:
While working on package manager integration I noticed that the current naming scheme can introduce problems if you download tarballs in parallel (or you don't immediately process each tarball).

The current S3 scheme is to have a path `/package-name/version.tar.gz`. However, if you download the tarballs for `effect@4.0.0` and `console@4.0.0` using these paths and accept the default download location, you'll receive two tarballs named `4.0.0.tar.gz`, and thus you'll overwrite the first download.

This PR proposes an alternate path: `/package-name/package-name-version.tar.gz`. The top-level package name still allows listing objects under a specific package name, and the nested package name allows distinguishing among downloaded tarballs. That same download to default paths under this scheme would produce `effect-4.0.0.tar.gz` and `console-4.0.0.tar.gz`. This also better matches the unpacked names, as extracting these tarballs will give you two directories `effect-4.0.0` and `console-4.0.0`. So it's the same naming throughout.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -8,7 +8,7 @@ A longer summary of the registry, why it came to exist, its high-level design. A
 
 **1.1 Important Terminology**
 
-Quick definitions of words that will be used frequently throughout the document. Readers can breeze through this section and refer back to it if they forget to what a common term refers. Some of these terms are in common use (like "hash"), but they're included because we mean something specific when we use it (ie. "a base64-encoded subresource integrity hash using the SHA256 algorithm.") 
+Quick definitions of words that will be used frequently throughout the document. Readers can breeze through this section and refer back to it if they forget to what a common term refers. Some of these terms are in common use (like "hash"), but they're included because we mean something specific when we use it (ie. "a base64-encoded subresource integrity hash using the SHA256 algorithm.")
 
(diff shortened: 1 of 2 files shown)


thomashoneyman:
Of course, this will require reuploading the registry (though not regenerating the registry index or metadata).

f-f:
We talked about this the other day and I was positive about it, but now that I think of it I wonder - is it necessary to make names this redundant?

As in, in any fetching method that I can think of there's a trivial way to override the destination file name to avoid name clashes.
In some cases (e.g. the new Spago) we don't even write files to disk, as we just keep the `ArrayBuffer` in memory and write it directly at the destination location.

thomashoneyman:
That’s true, it’s easy to override this. It’s not a big deal. On the other hand, I don’t really see any cost to changing the tarball URLs, and at least then you don’t _have_ to override the default behavior.

f-f:
Aside from having to reupload all the tarballs, we'd have longer and more redundant URLs. We have already put some effort in having a nice Registry URL before (you might recall the subdomain/DNS/certificates struggle), and I think it makes sense to continue on that track?

PR #514, opened 2022-09-06
More lenient parsing in GitHub issues

thomashoneyman:
When we decode operations submitted to the API via GitHub Actions we attempt to parse their contents as JSON. However, this means the entire issue body must be valid JSON, ie. something like this will work:

{ "newPackageName": "prelude" }

...but something like this will not:

```json
{ "newPackageName": "prelude" }
```

This PR adds a minimal cleanup pass to the body we receive as input, trimming the input to a matching `{ }` pair. Then we defer to our JSON parser for the rest of the work.

### DIFF
--- a/src/Registry/API.purs
+++ b/src/Registry/API.purs
@@ -136,11 +136,22 @@ readOperation eventPath = do
     Right event ->
       pure event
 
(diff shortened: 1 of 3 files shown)


f-f (approved):
Lovely!

PR #515, opened 2022-09-06
Add sync logic for legacy registry files

thomashoneyman:
Currently, users register their packages by submitting a PR to the `new-packages.json` file in this repository. However, they can also register their packages via the registry API using GitHub Issues. If they do so, then the legacy registry files in this repository won't be updated -- they aren't kept in sync.

This PR remedies this issue by synchronizing with the new-packages and bower-packages files in this repository when an addition or transfer happens.

This is also a step towards migrating the legacy registry files into the registry proper, and disabling pull requests, so that all registrations happen via the registry API (which provides feedback in case the registration fails, unlike the PR to the legacy files).

### DIFF
--- a/src/Registry/API.purs
+++ b/src/Registry/API.purs
@@ -1282,6 +1282,12 @@ syncLegacyRegistry package location = do
   let
     rawPackageName = "purescript-" <> PackageName.print package
 
+    -- Here we determine which, if any, legacy registry file should be updated with this package.
+    -- If the package is new (ie. not listed in either registry file) then we insert it into the
+    -- new-packages.json file. If not (ie. we found it in one of the registry files), and the location
+    -- of the package in the registry file is different from its one in the registry metadata, then we
+    -- update the package in that registry file. If the package exists at the proper location already
+    -- then we do nothing.
     targetFile = case Map.lookup rawPackageName newPackages, Map.lookup rawPackageName bowerPackages of
       Nothing, Nothing -> Just NewPackages
       Just url, _
--- a/src/Registry/API.purs
+++ b/src/Registry/API.purs
@@ -415,8 +415,13 @@ runOperation operation = case operation of
                   , err
                   , "cc: @purescript/packaging"
                   ]
-                Right _ ->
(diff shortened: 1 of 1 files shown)


src/Registry/API.purs
```diff
@@ -1246,3 +1255,58 @@ pacchettiBottiKeyType = "ssh-ed25519"
...
+  let newPackagesPath = legacyRegistryFilePath NewPackages
+  let bowerPackagesPath = legacyRegistryFilePath BowerPackages
+
+  newPackages <- Except.ExceptT $ Json.readJsonFile newPackagesPath
+  bowerPackages <- Except.ExceptT $ Json.readJsonFile bowerPackagesPath
+
+  let
+    rawPackageName = "purescript-" <> PackageName.print package
+
+    targetFile = case Map.lookup rawPackageName newPackages, Map.lookup rawPackageName bowerPackages of
```
f-f, line 1302: Maybe I'm missing something, but in the case of a new package I think this would return `Nothing` (because the new package is not in these files yet), so we wouldn't write/commit any file?

thomashoneyman: That’s right — if the package is in neither file, we return Just NewPackages to indicate it should be inserted there.

f-f: Oh, I see now. I'd say this piece of code it's confusing enough to be worth a comment explaining the logic

thomashoneyman, line 1302:
```suggestion
    -- Here we determine which, if any, legacy registry file should be updated with this package.
    -- If the package is new (ie. not listed in either registry file) then we insert it into the
    -- new-packages.json file. If not (ie. we found it in one of the registry files), and the location
    -- of the package in the registry file is different from its one in the registry metadata, then we
    -- update the package in that registry file. If the package exists at the proper location already
    -- then we do nothing.
    targetFile = case Map.lookup rawPackageName newPackages, Map.lookup rawPackageName bowerPackages of
```

f-f (approved):

PR #516, opened 2022-09-07
Verify manifest against request and metadata

thomashoneyman:
This PR adds two verification checks we've long planned but never implemented:

1. Make sure the manifest matches the package name submitted via the API (we try not to trust the API-provided information)
2. Make sure the manifest matches the registry metadata for everything _except_ the owners field. Transfers should happen via a Transfer operation, not by changing the manifest. This ensures transfers happen correctly as discussed in https://github.com/purescript/registry-dev/issues/331.

### DIFF
--- a/app/src/Registry/API.purs
+++ b/app/src/Registry/API.purs
@@ -536,10 +536,8 @@ publish source { name, ref, compiler, resolutions } inputMetadata = do
         let manifest = Legacy.Manifest.toManifest name version inputMetadata.location legacyManifest
         liftAff $ Json.writeJsonFile manifestPath manifest
 
-  -- TODO: Verify the manifest against metadata.
-  -- We trust the manifest for updating the owners field in the metadata, and
-  -- trust the metadata for everything else. No field should be different.
-  manifest@(Manifest manifestRecord) <- liftAff (try $ FS.readTextFile UTF8 manifestPath) >>= case _ of
+  -- Try to read the manifest, typechecking it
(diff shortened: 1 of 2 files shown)


src/Registry/API.purs
```diff
@@ -499,25 +498,42 @@ addOrUpdate { updateRef, buildPlan, packageName } inputMetadata = do
...
-  let
-    -- As soon as we have the manifest we need to update any fields that can
-    -- change from version to version.
-    metadata =
-      inputMetadata { owners = manifestRecord.owners }
+  -- We trust the manifest for any changes to the 'owners' field, but for all
+  -- other fields we trust the registry metadata.
+  let metadata = inputMetadata { owners = manifestFields.owners }
+
+  unless (isLegacyImport || manifestFields.name == packageName) do
```
f-f: I think this would throw for any legacy import?

thomashoneyman: It's saying _unless_ this is a legacy import, or unless the fields match, throw an error. But I could switch it to `when` instead:

```purs
when (not isLegacyImport && manifestFields.name /= packageName) do
```

f-f: Yeah, I can read this version better :smile:

f-f:
Did we just forget about this PR?

thomashoneyman:
Yes, it just never got approved. I think it's still the correct thing to do.

f-f:
Cool, let's get this through - does it make sense to rebase it now or best to wait until all the other things are settled?

thomashoneyman:
@f-f Want to give this another look?

f-f (approved):
Great, thanks!

PR #522, opened 2022-09-24
Add self-contained check to package set generation

thomashoneyman:
The 0.0.1 package set we converted from the legacy package sets and released has an issue: all packages work using the dependencies listed in the legacy sets, but since the new package sets use the registry manifest files, two packages are not actually self-contained (ie. they don't have all dependencies in the set).

We already have self-containment checks for any packages being added or removed to the sets, but we assumed that the prior package set was blessed and all packages proven to work. Alas, this was not the case.

This PR adds a self-containment check on the previous package set before we go to produce a new one. It also adds the check on the result of the new package set we produce, just to be extra sure. When I run the package set updater, I now see this error:

```
[COMMENT] Package set 0.1.0 is invalid! Some packages have dependencies not in the set:

  - quotient (proxy)
  - refined (generics-rep)
```

This error means that `quotient` has a list of dependencies, and they are all in the package set _except_ for the `proxy` dependency. To be usable in the set this dependency must be dropped.

### DIFF
--- a/src/Registry/Legacy/Manifest.purs
+++ b/src/Registry/Legacy/Manifest.purs
@@ -6,6 +6,8 @@ import Control.Monad.Except as Except
 import Control.Monad.Reader (ask)
 import Data.Array as Array
 import Data.Either as Either
+import Data.Filterable (filterMap)
+import Data.Formatter.DateTime as Formatter.DateTime
 import Data.FunctorWithIndex (mapWithIndex)
 import Data.Map as Map
 import Data.String as String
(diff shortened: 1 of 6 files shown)


thomashoneyman:
Once released, we can remove these two packages from the package sets, replace the 0.0.1 package set, and then run the importer tool so that the packages from the last few days are uploaded and a new package set is released.

src/Registry/Legacy/Manifest.purs
```diff
@@ -303,7 +306,16 @@ fetchLegacyPackageSets = do
   result <- liftAff $ Except.runExceptT $ GitHub.listTags octokit cache Constants.legacyPackageSetsRepo
   tags <- case result of
     Left err -> throwWithComment (GitHub.printGitHubError err)
-    Right tags -> pure $ map _.name tags
+    Right tags -> pure do
```
thomashoneyman, line 309: I've added this filter because all package sets after 2022-09-01 are registry-produced, and so there's no use in looking at them to determine dependencies (we produced those based on the package manifests in the first place). So we can save ourselves some processing power here.

src/Registry/PackageGraph.purs
```diff
@@ -80,41 +110,51 @@ checkRegistryIndex original = do
...
+  { satisfied :: Map PackageName (Array Version)
+  , constraints :: Array PackageWithDependencies
+  }
+
+type ConstraintResult =
+  { satisfied :: Map PackageName (Array Version)
+  , unsatisfied :: Array PackageWithDependencies
+  }
+
+solveConstraints :: ConstraintArgs -> ConstraintResult
```
thomashoneyman, line 124: I haven't changed the logic here, just moved it out to the top level so it can be reused in both the registry index and packages checks.

thomashoneyman:
I'm going to merge this just so that the process is unblocked, but please leave comments if you have them and I can address them in a followup PR.

PR #529, opened 2022-10-17
Switch to spago-next, split into monorepo with many packages

f-f:


thomashoneyman (approved):
Looks great!

What do you think of having an `app` top-level folder for the API and a `scripts` top-level folder for the legacy importer, package transferrer, etc. as those aren't really meant to be independent "apps'? We may get rid of them / move the code out into GitHub Actions fairly soon, too.

Other than that :+1: from me.

f-f:
Sure, the current setup is pretty temporary anyways, as e.g. the `Registry.API` module can't yet be in the `api` package. I shuffled things around as you suggested and will merge once CI is happy.

What code (and why) would we need to move into GitHub Actions? Should this be tracked in an issue?

thomashoneyman:
> What code (and why) would we need to move into GitHub Actions? Should this be tracked in an issue?

We won't run the legacy importer or the package transferrer once the alpha is complete and we're in beta (after all, those are only usable for packages that release semver tags on GitHub, and the registry is meant to support a wider range of packages including monorepos and non-GitHub projects).

When we're in beta we can remove the package transferrer altogether and move the legacy importer into a GitHub Action for core / contrib / web / node to use until they all switch over to the new Spago.

PR #532, opened 2022-10-24
Ensure scripts use the root for cache, scratch, env files

thomashoneyman:
@f-f The various CI jobs are currently failing because the scripts and `registry-api` utility expect to be executed from the root of the repository (for example, `DotEnv.loadFile` expects there to be a `.env` file in the current directory; we also expect the current directory to contain the `scratch` dir, `.cache` directory, and so on). However, in the new monorepo setup Spago is executing in the context of the individual packages so the paths are no longer correct.

I've fixed this by ensuring that we refer to the `scratch`, `.cache` directories and the `.env` file in their proper location (the root). However, question for you: to get to the root I'm having to do `../../`, but I would expect Spago to execute with the `cwd` being the directory containing the `spago.yaml` file. Is this expected? If not, it's an easy change on our side (just change `rootDir`).


thomashoneyman:
If you check before this change you'll see that attempting to read the Dhall directory fails (incorrect path); attempting to read the .env file fails silently (you just get a "not found" for each env variable in it rather than an outright crash); you'll see that the `scratch` dir is incorrectly created in the various script directories.

f-f (commented):
This makes sense overall, as indeed we're going from running everything in the root to running every package on its own - that means that Spago will set the package folder (where the `spago.yaml` for the package is) as the `cwd` of the running process.

Now, hoping that this all makes sense, the gotcha here is that while the things in `scripts` will need to point to the root as `../../`, the API script will need only `../`.

We should ideally make this uniform so that we don't have to adjust paths all the time - I see a few solutions to this:
1. Move the `api` folder inside `scripts` 
2. or add a `cwd` flag/option to `spago run` so that we can go back to run in the root
3. or unify all the script-y things in a single package and go back to running them with `-m`
4. or patch Spago to provide the root of the repo in the `Spago.Generated.BuildInfo` module (there is some info in there already, and we'll want to add more anyways, like the version of the compiler and the timestamp for the build)

thomashoneyman:
I think for the time being I'd prefer to unify the scripts in the `scripts` dir to be one package and run it with `-m`. Open to other things for the long term. What do you think?

f-f:
Note that this implies that the `API` module would end up there too (so option 1 is very similar to option 3), I was under the impression that you wanted to have that separate from the legacy scripts?

f-f:
Ah well right, the scripts don't have to be together with the API module, it's enough that they are at the same level of nesting

f-f:
So yeah, this sounds good to me!

thomashoneyman:
> Note that this implies that the `API` module would end up there too (so option 1 is very similar to option 3), I was under the impression that you wanted to have that separate from the legacy scripts?

If the `scripts` directory becomes:

- `scripts/LegacyImport.purs`
- `scripts/PackageTransferrer.purs`
- `scripts/spago.yaml`

...then that solves the problem because the nesting becomes even with the API.

thomashoneyman:
Oh, I see you already commented that. Cool -- I'll go ahead with the change.

f-f (approved):
Great!

Issue #531, opened 2022-10-23
Remove legacy packages that do not solve


MonoidMusician:
And prevent reupload.

I think our integration test isn't doing the exact thing we need (for two reasons: I think part of it is still mixed with Bower data, and it only checks for the most recent version), but it gives an idea of which packages are problematic:

<details>

```
bifunctors@0.0.3
bifunctors@0.0.3
free@2.0.0
free@3.3.0
lazy@0.1.1
lazy@0.1.1
quickcheck@0.3.1
quickcheck@0.3.1
transformers@0.3.0
transformers@0.3.0
ace@2.0.0
ace@2.0.0
aff@2.0.2
affjax@3.0.2
affjax@3.0.2
argonaut@0.9.0
argonaut@0.9.0
argonaut-traversals@2.0.1
arraybuffer@11.0.0
arraybuffer@11.0.0
coroutines@3.1.0
fork@1.0.0
formatters@0.1.0
formatters@0.1.0
matryoshka@0.2.0
matryoshka@0.2.0
parsing@0.1.5
parsing@0.1.5
profunctor-lenses@2.6.0
routing@4.0.0
routing@4.0.0
string-parsers@0.2.0
string-parsers@0.2.0
node-child-process@2.0.0
node-child-process@3.0.1
node-fs@2.0.0
node-fs@2.0.0
node-fs-aff@2.0.0
node-fs-aff@2.0.0
```

</details>

Related:
- https://github.com/purescript/registry-dev/issues/300
- https://github.com/purescript/registry-dev/pull/465

MonoidMusician:
Here's the full list of package versions which fail to solve:

https://gist.github.com/MonoidMusician/0f0c4ed70edf7c000a1da012bdca5134

Definitely too many to fix 😄

I'm fairly sure this list is accurate, pending more verification of the latest iteration of the solver. That said, I don't think I made substantial changes to the core logic of the solver so I'm fairly confident in it. In fact, all but 2 of these failures were found with merely the transitive dependencies check in the solver, not the full logic for picking versions.

Interestingly, all of the integration test packages failures I listed above are included in this new list, which I was not expecting since (afaik) that was running on mixed registry/bower data, although it was filtering out packages that just failed to solve on bower.

f-f:
How could these be published under Bower? If they don't solve they should not go through their publishing process, right?

MonoidMusician:
[...]

PR #533, opened 2022-10-25
Update spec with data types

thomashoneyman:
This PR updates the `SPEC.md` file with richer descriptions of the various data types used by the registry. Along the way I have added links to the "source code" for these data types, but in practice those files don't yet exist (they're in the `trh/new-codecs` branch). Please forgive the broken links, and I'll move the files to their proper location in a new PR to minimize the review load.


SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+**[Source](./lib/src/Registry/Metadata.purs)**
+**[Spec](./specs/v1/Metadata.dhall)**
+
+All packages in the registry have an associated metadata file, which is located in the `metadata` directory of the `registry` repository under the package name. For example, the metadata for the `aff` package is located at: https://github.com/purescript/registry/blob/main/metadata/aff.json. Metadata files are the source of truth on all published and unpublished versions for a particular package, as well as where the package is located. Metadata files are produced by the registry, not by package authors, though they take some information from package manifests.
+
+Each published version of a package records three fields:
+
+- `hash`: a [`SRIHash`](#srihash) of the tarball contents of the given version
+- `bytes`: the size of the tarball in bytes
+- `publishedTime`: the time the package was published as an `ISO8601` string
```
thomashoneyman, line 226: While working on the `new-codecs` branch, specifically with datetimes, I realized that the RFC3339String implementation used by the `precise-datetime` library is _not_ actually what the ECMAScript standard uses -- it actually uses a simplified ISO8601String:

https://tc39.es/ecma262/#sec-date-time-string-format

They're basically the same, except that the simple ISO8601 format always has 3 milliseconds places and the RFC3339 format used by `precise-datetime` can have between 1 and 3 places (`.000` vs `.0`, or `.520` vs. `.52`). I want to use the ECMAScript format because under the hood we're using JavaScript `DateTime` values and relying on the `Date` object, which required by the spec to be able to parse the simple ISO8601 format. This also lets us drop a JS dependency on `decimal.js` in the registry library, and the full dependency on `precise-datetime`, which is not a core or contrib library.

So across the board I think it's good to switch to the simple ISO8601 format. The only downside is that we will have to regenerate the registry metadata, but that's a quick tweak.

I'm already using the ISO8601 format throughout the `trh/new-codecs` branch.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+Each unpublished version of a package records three fields:
+
+- `reason`: a plain text string up to 300 characters long explaining why the version was unpublished
```
thomashoneyman, line 230: For plain text descriptions submitted by users, such as descriptions or reasons in the metadata, I've chosen an arbitrary character limit of 300 characters. Not sure how everyone feels about this. I just don't want it to be unbounded, and it shouldn't be used as a bona-fide README, so some reasonably short character limit makes sense.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
-**3.7 Package Sets**
-Schemas for the package sets
+Each package file is a JSON Lines file where each line is a package manifest, stored in sorted order ascending by version. This format follows the one documented in the Cargo book: https://doc.rust-lang.org/cargo/reference/registries.html#index-format.
 
 ## 4. Registry Infrastructure
 
 A section specifying how the registry should relate to other infrastructure and the responsibility of that infrastructure as it relates to the registry.
 
-**4.1 Registry Index**
+**4.1 Package Index**
```
thomashoneyman: I've opted for "Package Index" instead of "Registry Index" and "Package Storage" instead of "Storage Backend" for consistency's sake. We then get things like the "Package Storage", "Package Index", "Package Sets", "Package Metadata", "Package API", and so on, so terms like "Storage", "Index", etc. can be assumed to mean "Package ..."

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
 1. This section is only about _behavior_. The data types used in the API have already been described in prior sections.
 2. This section is only about _registry_ concerns. Actions taken after an operation (such as pushing to package sets, pursuit, the registry index) are in the next section.
 
-**5.1 Register a Package**
-Walkthrough of package registration. Includes the various checks performed to ensure a package is OK.
+### Package Operations
```
thomashoneyman, line 330: I've separated the concept of a "package operation" from a "package set operation", because they're quite different in how we authenticate them, how they affect the registry, how the request is submitted (issues vs. the http api), and so on.

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
+2. If the highest SemVer upgrade in the set was a minor version, or any new packages were added to the package set, then the package set increments a minor version.
+3. If the highest SemVer upgrade in the set was a patch version, then the package set increments a patch version.
+
+For example, if the previous release was `2.1.1`, and the package set update adds a new package, then the new version would be `2.2.0`.
+
+## 6. Post-Publishing Operations
+
+Summarizes what operations the registry will take after a package is published, transferred, or unpublished. This section _could_ be folded back into the previous section, but I worry that doing so would make that section too long and unfocused.
+
+#### 6.1 Update PAckage Index
```
thomashoneyman:
```suggestion
#### 6.1 Update Package Index
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+- Must be no more than 50 characters long
+- Must contain only letters and digits, optionally separated by hyphens
+- Must begin with a letter or digit
+- Cannot contain consecutive hyphens
+- Cannot begin with `purescript-`
+
+The final point deserves elaboration. Historically, PureScript packages have been registered in the Bower registry using a `purescript-` prefix so as to avoid naming conflicts with JavaScript packages. However, the prefix is not actually part of the package name, so package managers like Spago refer to the package without the prefix. For example, the Bower-registered name `purescript-prelude` refers to the `prelude` package, which is located in the GitHub repository `purescript/purescript-prelude`. This prefix is not used in the PureScript registry. The `purescript-prelude` Bower package is simply `prelude` in the PureScript registry.
+
+To prevent users from accidentally registering their packages with an unnecessary prefix, the registry will not accept packages that begin with `purescript-` unless the package author confirms this is what they want with the Registry Trustees.
```
JordanMartinez: Just a general question. Users have traditionally created repos that are prefixed with `purescript-`. Should they continue to do so?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+**GitHub Repository**
+
+The registry has special support for fetching Git repositories hosted on GitHub. Instead of cloning the repository it will fetch the tarball of the package contents. To specify a GitHub repository you must provide the name of the owner and of the repository, along with an optional subdirectory containing the package source. A JSON example:
+
+```json
+{
+  "githubOwner": "purescript",
+  "githubRepo": "prelude"
+}
```
JordanMartinez, line 138: Should this include the `libDir` as an example?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+#### Owner
+
+**[Source](./lib/src/Registry/Owner.purs)**
+**[Spec](./specs/v1/Owner.dhall)**
+
+The registry relies on SSH key pairs to verify package ownership for the purposes of sensitive API operations like unpublishing versions or transferring packages. An `Owner` is made up of the three components of an SSH public key in text format ([RFC4253](https://www.rfc-editor.org/rfc/rfc4253#section-6.6)). That format looks like this:
+
+`[keytype] [ssh-public-key] [email]`
+
+Note that the email address does not have to be valid, but it must be the address associated with the private key. A JSON example:
```
JordanMartinez: Perhaps something other than `valid` should be used here for describing emails? One could read this as "I can use `foo.com` as an email" when I think you actually mean the email address may not be able to receive/send emails.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+```
+
+#### License
+
+**[Source](./lib/src/Registry/License.purs)**
+**[Spec](./specs/v1/License.dhall)**
+
+All packages in the registry must have a license that grants permission for redistribution of the source code. Concretely, the registry requires that all packages use an SPDX license and specify an [SPDX license identifier](https://spdx.dev/ids/). `AND` and `OR` conjunctions are allowed. A `License` is represented as a string, which must be a valid SPDX identifier. For example:
+
+`"MIT OR APACHE-2.0"`
```
JordanMartinez, line 169: What about the `WITH` conjunction? Last example in the 'How' section: https://spdx.dev/ids/#how

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+**[Source](./lib/src/Registry/Manifest.purs)**
+**[Spec](./specs/v1/Manifest.dhall)**
+
+All packages in the registry contain a `purs.json` manifest file in their root directory. The manifest file specifies information necessary for the registry to package the source code, and it also serves as a lingua franca for package managers: all package managers are expected to support this manifest format. The manifest file is an object with the following fields:
+
+- `name`: a valid [`PackageName`](#packagename)
+- `version`: a valid [`Version`](#version)
+- `license`: a valid [`License`](#license)
+- `location`: a valid [`Location`](#location)
+- `owners`: a non-empty array of [`Owner`](#owner) (optional)
```
JordanMartinez: Could `(optional)` be updated to `(optional field)`. I found it surprising/confusing initially to see `non-empty array` and `optional`, as though the array itself was optional.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+- `description`: a description of your library as a plain text string, not markdown, up to 300 characters (optional)
```
JordanMartinez: Just a thought, but since `description` is more for human-readable content, should there be a corresponding `tags` or `categories` field that serves to indicate what kind of a library it is (e.g. database, front-end, strings, utils, etc.)? While Pursuit doesn't support this now, it might be something we'll want for the future.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+- `files`: a non-empty array of globs, where globs are used to match files outside the `src` directory you want included in your package tarball
```
JordanMartinez: Could you clarify what style of source globs are allowed here? I'm assuming it's probably:
- `dir/**/*.purs`
- `dir/**/*`
- `dir/*`

And not something like `dir/**/*.(js|purs)`

JordanMartinez: Ah... nevermind. You clarify this below. Perhaps that comment should be put in an indented bullet list within this one?

SPEC.md
JordanMartinez: I agree with using the full 3 milliseconds place.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+}
+```
+
+### 3.5 Package Set
+
+**[Source](./lib/src/Registry/PackageSet.purs)**
+
+Package sets are stored in the `registry` repository under the `package-sets` directory. For example, the first package set was published at https://github.com/purescript/registry/blob/main/package-sets/0.0.1.json. Each package set is stored as a JSON file with the following fields:
+
+1. `version`, which is a string containing the package set [`Version`](#version) for this set.
```
JordanMartinez: What does Major, Minor, and Patch mean in package set versions?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
-**3.6 Registry API**
-Schemas for the registry API, including `Addition`, `Update`, `Transfer`, and `Unpublish`. Also includes `Authenticated` constructor.
+- Packages with 1-character names are stored in the directory named `1`.
+- Packages with 2-character names are stored in the directory named `2`.
+- Packages with 3-character names are stored in the directory `3/{first-character}` where `{first-character}` is the first character of the package name. For example, the `aff` package is stored in the `3/a` directory.
+- All other packages are stored in directories named `{first-two}/{second-two}` where the top directory is the first two characters of the package name, and the subdirectory is the third and fourth characters of the package name. For example, the `prelude` package is stored in the `pr/el` directory.
 
-**3.7 Package Sets**
-Schemas for the package sets
+Each package file is a JSON Lines file where each line is a package manifest, stored in sorted order ascending by version. This format follows the one documented in the Cargo book: https://doc.rust-lang.org/cargo/reference/registries.html#index-format.
```
JordanMartinez: I think this should near or immediately after the `All manifests for a given package...` sentence above. It helps clarify where the above rules come from.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
 
 ## 4. Registry Infrastructure
 
 A section specifying how the registry should relate to other infrastructure and the responsibility of that infrastructure as it relates to the registry.
 
-**4.1 Registry Index**
+**4.1 Package Index**
 Specifies how the registry index works (data format, purpose of it existing, where it is located)
```
JordanMartinez:
```diff
-Specifies how the registry index works (data format, purpose of it existing, where it is located)
+Specifies how the package index works (data format, purpose of it existing, where it is located)
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
 
-**4.2 Storage Backend**
+**4.2 Package Storage**
 Specifies how the storage backend works (data format, naming conventions, location)
```
JordanMartinez, line 310:
```diff
-Specifies how the storage backend works (data format, naming conventions, location)
+Specifies how the package storage works (data format, naming conventions, location)
```

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
 
-**5.2 Update a Package**
-Walkthrough of updating a package version.
+The registry supports three package operations: publishing, unpublishing, and transferring. These operations are exposed via an HTTP API at [registry.purescript.org/api](https://registry.purescript.org). You are expected to provide a JSON body matching the operation data type in a POST request.
 
-**5.3 Update the Package Set**
-Walkthrough of updating the package sets.
+- `/api/publish`: The `Publish` operation
+- `/api/unpublish`: The `Unpublish` operation
+- `/api/transfer`: The `Transfer` operation
```
JordanMartinez: Should these have a version associated with them? For example
```
/api/v1/publish
```

SPEC.md
JordanMartinez: Ah... you clarify this in the 'Releasing a Package Set' section. Perhaps link to that here?

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
+
+For example, if the previous release was `2.1.1`, and the package set update adds a new package, then the new version would be `2.2.0`.
+
+## 6. Post-Publishing Operations
+
+Summarizes what operations the registry will take after a package is published, transferred, or unpublished. This section _could_ be folded back into the previous section, but I worry that doing so would make that section too long and unfocused.
+
+#### 6.1 Update Package Index
+
+A spec of how the registry attempts to update the registry index.
```
JordanMartinez:
```diff
-A spec of how the registry attempts to update the registry index.
+A spec of how the registry attempts to update the package index.
```

SPEC.md
```diff
@@ -180,34 +640,53 @@ Third, we attempt to add the rest of the batch of package versions to the packag
 4. If a package fails to compile with the rest of the package set, then it is filtered from the batch.
 5. Once there are no more packages to consider in the batch, the new package set is ready.
 
-Fourth, we release the new package set (if we could produce one). Automatic package sets follow the below versioning policy:
+Fourth, we release the new package set (if we could produce one). Automatic package sets follow the versioning policy described in [Section 5.5 (Update the Package Set)](#55-update-the-package-set).
```
JordanMartinez: This link points to the overall section, but not quite the version policy specifically. This could be updated to point to that link.

JordanMartinez (commented):
Overall looks good to me! Sorry for taking so long to read through this.

SPEC.md
thomashoneyman: As far as the registry is concerned it doesn’t matter. As far as encouraging a community convention, I think it’s a nice one and I’d encourage it to continue!

SPEC.md
thomashoneyman: It _sort_ of is including `subdir` -- it's optional, so I've tried to show it both in use or omitted.

SPEC.md
thomashoneyman: Perhaps "usable"?

SPEC.md
thomashoneyman: `WITH` is supported, but it isn't a conjunction. It's used to modify an identifier by applying an exception, if I remember correctly, like `MIT WITH files-exception`. That is still considered a "license identifier".

SPEC.md
thomashoneyman: Yep! I can prepend rather than append it. This would become `owners: (optional) a non-empty...`. Alternately, I could take a note from TypeScript and use `?`:

```
- `owners?`: a non-empty ...
```

but I'm not sure if that would be obvious.

SPEC.md
thomashoneyman: I agree with you. The reason it isn't in there right now is we haven't really discussed how it would be used, and so:

> While Pursuit doesn't support this now, it might be something we'll want for the future.

This sort of thing -- "something we'll want for the future" -- is generally omitted, because we can add it in a forward-compatible way. The reason we use `description` is that packages already use it in their `bower.json` files, even though IIRC Pursuit doesn't use it.

So I agree something like that would be nice, but until we have a good discussion on how we'd like to use it I'm holding off adding it to the type.

SPEC.md
thomashoneyman: Yes, that's a good idea. I can move the glob comment and the dependencies comment up.

SPEC.md
thomashoneyman: Are you referring to the link?

SPEC.md
thomashoneyman: On second thought, we ended up deciding to use `Metadata Index` instead of `Package Metadata` and `Manifest Index` instead of `Package Index` to be more clear what exactly is in the map.

SPEC.md
thomashoneyman: Yes, it's a good idea to version the API. Thanks!

SPEC.md
JordanMartinez: Sounds good to me.

SPEC.md
JordanMartinez: Ok, thanks for clarifying.

SPEC.md
JordanMartinez: What about
```
- `owners` (optional): a non empty...
```

- `owners` (optional): a non empty...

SPEC.md
JordanMartinez: Ok, sounds good.

SPEC.md
JordanMartinez: Sorry for not being clearer. Yes. Specifically,
> This format follows the one documented in the Cargo book: https://doc.rust-lang.org/cargo/reference/registries.html#index-format.

specs/v1/Owner.dhall
```diff
@@ -0,0 +1 @@
+let Owner = { email : Text, keytpe : Text, public : Text } in Owner
```
f-f:
```suggestion
let Owner = { email : Text, keytype : Text, public : Text } in Owner
```

flake.nix
```diff
@@ -142,16 +142,16 @@
             cd $(git rev-parse --show-toplevel)
             set -euo pipefail
 
-            for FILE in $(find v1 -iname "*.dhall")
+            for FILE in $(find ./specs/v1 -iname "*.dhall")
```
f-f: How do you feel about renaming `specs` to `types`? That word feels off to me for some reason.

SPEC.md
f-f: I would also like to continue this convention - it is useful to know at a glance which language a package concerns. I wonder if we should mention this convention in the spec to avoid confusion or even just to encourage it?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+Packages are associated with one or more versions, representing their source code at a point in time. The registry uses a restricted version of the SemVer spec which only allows versions with major, minor, and patch places (that means no build metadata and no prerelease identifiers). A version is represented as a `string` with the following restrictions:
+
+- The string must have the form `"X.Y.Z"`, representing major, minor, and patch places.
+- The major, minor, and patch places must each be natural numbers (ie. whole numbers equal to or greater than 0).
+
+We never use a `v` prefix on versions in the registry.
+
+If a package uses all three places (ie. it begins with a non-zero number, such as `"1.0.0"`), then:
+
+- `MAJOR` means values have been changed or removed, and represents a breaking change to the package.
```
f-f: I feel this "values" word is not too clear in this context - it's probably worth elaborating on the concept of "public interface" of a package (this is a useful thing to do on its own in any case), and frame this as "changes or removals to the names exposed by the public interface" (or something a little less wordy than this if we manage)

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+- `MINOR` means values have been added, but existing values are unchanged.
+- `PATCH` means the API is unchanged and there is no risk of breaking code.
```
f-f: Here it comes to mind the case where a bugfix leads to a change to a (previously broken) interface. Is that a new MAJOR? If yes, it's worth mentioning here this corner case, since it might not be immediately intuitive.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+**[Spec](./specs/v1/Location.dhall)**
+
+The registry operates on a pull-based model, in which package authors ask the registry to fetch their code and produce a package instead of uploading their package themselves. Accordingly, every package is associated with the location the registry will use to fetch it. A location is represented in JSON as an `object`.
+
+Packages can be located anywhere online so long as that location is publicy-accessible and the registry understands how to fetch code from it.
+
+By convention, locations are represented in JSON using the following rules:
+
+- Common fields (`url`, `owner`, `repo`) are prefixed with the provider name, ie. `githubOwner` or `gitlabRepo` or `gitUrl`.
+- Providers that support a typical filesystem structure must support an optional `subdir` key so that the registry can provide monorepo support for that location type.
```
f-f, line 105: I love the way this is phrased!

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+The currently-supported location types are below.
```
f-f:
```suggestion
The currently-supported location types are listed below.
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+**Git Repository**
+
+An arbitrary Git repository is represented by a JSON `object` with a required `gitUrl` key and optional `subdir` key. The Git URL must be of the form: `http[s]://host.xz/path/to/repo.git`. The registry uses the `git` CLI tool to fetch your repository. A JSON example:
```
f-f: This specification seems to imply that we are parsing/validating this. Should we be parsing it so that we are sure we adhere to the spec?

SPEC.md
f-f: Not sure if I like "usable", maybe "able to send and receive emails" just to be clear?

SPEC.md
f-f: Should we spec this clarification?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+All packages in the registry must have a license that grants permission for redistribution of the source code. Concretely, the registry requires that all packages use an SPDX license and specify an [SPDX license identifier](https://spdx.dev/ids/). `AND` and `OR` conjunctions are allowed. A `License` is represented as a string, which must be a valid SPDX identifier. For example:
+
+`"MIT OR APACHE-2.0"`
+
+#### Sha256
+
+**[Source](./lib/src/Registry/Sha256.purs)**
+**[Spec](./specs/v1/Sha256.dhall)**
+
+The registry produces a tarball when publishing a package. The hash of this tarball is recorded so that package managers can verify the integrity of packages they download from the registry. The hash is stored in the [subresource integrity (SRI) format](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). , A SRI hash is represented as a `string`. For example:
```
f-f: We haven't defined "tarball" anywhere, as we assume it's commonly used jargon, but maybe we should not take it as an assumption?
Perhaps worth changing the wording a bit, or even just linking to e.g. [the Debian wiki's entry about it](https://wiki.debian.org/TarBall)

f-f:
```suggestion
The registry produces a tarball when publishing a package. The hash of this tarball is recorded so that package managers can verify the integrity of packages they download from the registry. The hash is stored in the [subresource integrity (SRI) format](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).
A SRI hash is represented as a `string`. For example:
```

SPEC.md
f-f: Yes! Because of our forward-compatibility guarantee (i.e. package managers based on the first iteration of the spec need to be able to read manifests from newer iterations, so we can't change/remove fields, only add them) our best strategy for pretty much anything is "make sure we leave the door open for feature X, while we discuss properly how to go at it, because we can't change it afterwards"

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+    "githubOwner": "purescript",
+    "githubRepo": "purescript-control"
+  },
+  "dependencies": { "newtype": ">=3.0.0 <4.0.0", "prelude": ">=4.0.0 <5.0.0" }
+}
+```
+
+A note about dependencies: all dependencies you provide must exist in the registry, and the dependency ranges must be solvable (ie. it must be possible to produce a single version of each dependency that satisfies the provided version bounds, including any transitive dependencies).
+
+A note about globs: globs must contain only `*`, `**`, `/`, `.`, `..`, and characters for Linux file paths. It is not possible to negate a glob (ie. the `!` character), and globs cannot represent a path out of the package source directory.
```
f-f: `..` in a glob is a potential security attack vector, is it necessary to include it or can we ban it?

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+  "license": "BSD-3-Clause",
+  "location": {
+    "githubOwner": "purescript",
+    "githubRepo": "purescript-control"
+  },
+  "dependencies": { "newtype": ">=3.0.0 <4.0.0", "prelude": ">=4.0.0 <5.0.0" }
+}
+```
+
+A note about dependencies: all dependencies you provide must exist in the registry, and the dependency ranges must be solvable (ie. it must be possible to produce a single version of each dependency that satisfies the provided version bounds, including any transitive dependencies).
```
f-f: This sentence effectively bans unsolvable packages, cf #531

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+A note about globs: globs must contain only `*`, `**`, `/`, `.`, `..`, and characters for Linux file paths. It is not possible to negate a glob (ie. the `!` character), and globs cannot represent a path out of the package source directory.
+
+### 3.4 Metadata
+
+**[Source](./lib/src/Registry/Metadata.purs)**
+**[Spec](./specs/v1/Metadata.dhall)**
+
+All packages in the registry have an associated metadata file, which is located in the `metadata` directory of the `registry` repository under the package name. For example, the metadata for the `aff` package is located at: https://github.com/purescript/registry/blob/main/metadata/aff.json. Metadata files are the source of truth on all published and unpublished versions for a particular package, as well as where the package is located. Metadata files are produced by the registry, not by package authors, though they take some information from package manifests.
```
f-f:
```suggestion
All packages in the registry have an associated metadata file, which is located in the `metadata` directory of the `registry` repository under the package name. For example, the metadata for the `aff` package is located at: https://github.com/purescript/registry/blob/main/metadata/aff.json. Metadata files are the source of truth on all published and unpublished versions for a particular package, on what their content is, as well as where the package is located. Metadata files are produced by the registry, not by package authors, though they take some information from package manifests.
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+Each published version of a package records three fields:
+
+- `hash`: a [`Sha256`](#Sha256) of the tarball contents of the given version
```
f-f:
```suggestion
- `hash`: a [`Sha256`](#Sha256) of the compressed archive fetched by the registry for this version
```

SPEC.md
f-f: I agreed as well that ISO8601 is a better format overall.

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+- `bytes`: the size of the tarball in bytes
+- `publishedTime`: the time the package was published as an `ISO8601` string
+
+Each unpublished version of a package records three fields:
+
+- `reason`: a plain text string up to 300 characters long explaining why the version was unpublished
+- `publishedTime`: the time the package was published as an `ISO8601` string
```
f-f:
```suggestion
- `publishedTime`: the time the version was published, as an `ISO8601` string
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+- `unpublishedTime`: the time the package was unpublished as an `ISO8601` string
```
f-f:
```suggestion
- `unpublishedTime`: the time the version was unpublished, as an `ISO8601` string
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+The metadata for a given package contains up to four fields:
+
+- `location`: a valid [`Location`](#location) representing where the registry is currently fetching this package's source code
```
f-f: Not sure if it's idiomatic, but this sounds better to me:
```suggestion
- `location`: a valid [`Location`](#location) representing where the registry is currently fetching this package's source code from
```

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
+
+### 3.6 Package Index
 
-**3.3 Atomic Data**
-Schemas for `PackageName`, `Version`, `Range`, `Location`, `Hash`, `Owner`, `License`, and other simple data used in the registry. Each is a subsection which includes its rules, the representations in Dhall and JSON, and a link to the PureScript implementation in `Schema.purs`.
+**[Source](./lib/src/Registry/PackageIndex.purs)**
 
-**3.4 Manifest**
-Schema for the manifest format.
+The registry maintains an index of all package manifests in the [`registry-index`](https://github.com/purescript/registry-index) repository. This index makes it convenient for package managers to look up the manifest for a particular package version at any time. The package index maintains the invariant that all dependencies of packages in the index are themselves in the index.
```
f-f:
```suggestion
The registry maintains an index of all package manifests in the [`registry-index`](https://github.com/purescript/registry-index) repository. This index makes it convenient for package managers to look up the manifest for a particular package version at any time. The package index - just like the Registry itself - maintains the invariant that all dependencies of packages in the index are themselves in the index.
```

SPEC.md
f-f: I agree. Also I'd like to not see the explicit URL, but it should be a link around `the one documented in the Cargo book`

SPEC.md
```diff
@@ -22,35 +22,285 @@ A simple walkthrough of how a package is published, possibly including diagrams.
...
 
-**3.5 Metadata**
-Schema for the metadata format.
+All manifests for a given package in the registry are stored in the registry index according to the following rules:
```
f-f:
```suggestion
All manifests for a given package in the registry are cached in the registry index according to the following rules:
```

We are not explicitly speccing anywhere that the Index is a cache and the Registry is perfectly able to work without it.
The suggestion above is the bare minimum we can do, but probably worth having a paragraph about how it's possible to rebuild the index from just (1) the metadata files and (2) all the tarballs

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
+**Verification Steps**
 
-This section is meant to be reasonably self-contained, so someone writing a new package manager can understand what they need to do and what resources they have available in one place.
+For a `Publish` operation to succeed, the provided JSON payload must decode successfully and the following verification checks must be satisfied. Package verification is separated into stages depending on what information the registry has available.
 
-## 9. Policies
+First, the registry uses the [`PackageName`](#packagename) listed in the JSON payload to read any existing [`Metadata`](#34-metadata) for the package. To be publishable, one of these two conditions must be true:
 
-Policies followed by the registry — though maybe these are better-suited to be stored as separate specs apart from this document.
+1. The package name is not registered (ie. no metadata exists) AND a location is provided in the payload AND the location is not already in use. In this case, the registry will create new metadata to register this package at the given location.
```
f-f:
```suggestion
1. The package name is not registered (ie. no metadata exists) AND a location is provided in the payload. In this case, the registry will create new metadata to register this package at the given location.
```

I propose we abandon this restriction, as it's not really necessary.

As an example, if we had an `HttpLocation { url :: String }` alongside the existing `Git` and `GitHub` ones, you could have your own server serving packages for the registry to fetch and use the same `url` for different packages, and still have no risk of confusion if you use different `ref`s for different packages, e.g. `ref: "package1/v1.0.1"` and `ref: "package2/v2.2.0"`

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
+
+Unpublishing a package version means that the following things happen:
+
+1. The package version is moved from the `published` to the `unpublished` section of the package metadata
+2. The package version is removed from package storage and the package index, which means that package managers can no longer install it and it will no longer be included in build plans.
+
+Notably, the package version will remain listed in the package sets, rendering those package sets invalid. To minimize the potential disruption caused by unpublishing a package, the following conditions must be met for a package to be unpublished:
+
+1. The package version must have been published (ie. you cannot unpublish a version that does not exist).
+2. The version must have been published within the last 48 hours.
```
f-f: I'm afraid we'll need an additional condition for DMCA takedowns.

SPEC.md
```diff
@@ -68,101 +318,311 @@ This is a good time for diagrams! (cc: @AndrewCondon, @JordanMartinez).Two notes
...
+Anyone can suggest a package set update to the registry. However, community members can only add or upgrade packages in the package set. To remove a package or downgrade a package version, the update must be submitted by a member of the packaging team or a Registry Trustee.
+
+A package set update must be submitted to the registry via GitHub issues on the registry repository. The body of the issue should contain the JSON object specified below, optionally within a code fence.
+
+A package set update is an object with two keys: `compiler`, an optional field that allows updating the compiler version used to compile the package sets, and `packages`, an object where keys are package names and values are either a version number or `null`. A version number indicates the package should be added to the set or updated to the given version, and `null` indicates the package should be dropped from the package set.
+
+```jsonc
+{ // Sets the package set compiler version to 0.15.2
+  "compiler": "0.15.2",
```
f-f, line 571: What happens if I submit an update with a compiler version that is lower than the current one in the sets?

f-f (commented):
Thanks for typing all of this down! I have probably missed something, but here's my first batch of notes 😄

SPEC.md
f-f: A more compelling problem is the situation where a package owner wants to rename a package.
Since we can't unregister packages, that is effectively an addition.

If the owner wants to keep the same location, the workflow is more involved than it should:
1. add SSH keys to the current package
2. publish a new version so that the registry picks up the keys
3. issue a transfer to a different location
4. register the new package

If we got away with this restriction they'd have to go only trough step 4.

SPEC.md
thomashoneyman: I've added a little note to this section.

SPEC.md
thomashoneyman: This was my proposal, and then we'd update the parsing to match the spec. In short, we want to support any Git URL that we can use `git clone` on, which means disallowing some forms that we can't use (like SSH).

But I've realized that my restriction is too strong on the extension -- for example, SourceHut doesn't use the .git extension:

`git clone https://git.sr.ht/~rj/digitalocean-image`

So perhaps `http[s]://host.xz/path/to/repo[.git]` is correct.

SPEC.md
thomashoneyman: Yep! I think we should ban them, but if we find out we _can't_ do that, because there are too many, then we can relax this.

SPEC.md
thomashoneyman: I suppose we could omit it! We always protect against directory traversals with this:

https://github.com/purescript/registry-dev/blob/030a58fe7938e540278c51d551e3ff853d2d3403/lib/src/Registry/Internal/Path.purs#L19

...but I guess we could disallow the use of `..` altogether. Note that right now we just check for directory traversals and then pass off to `fast-glob`, so we don't have explicit parsing for these globs. We'd have to add that later to match the spec.

SPEC.md
thomashoneyman: It should be rejected. I don't think we should be able to downgrade package sets (ie. publish a `0.15.4` set and then start publishing `0.15.2` packages). Once we're on a new series we're on the new series.

thomashoneyman:
I've addressed all feedback — I think everything is addressed, though these may still need discussion @f-f, @colinwahl:

https://github.com/purescript/registry-dev/pull/533#discussion_r1021971165
https://github.com/purescript/registry-dev/pull/533#discussion_r1021962737
https://github.com/purescript/registry-dev/pull/533#discussion_r1021991346
https://github.com/purescript/registry-dev/pull/533#discussion_r1021975264

Jordan is on PTO and won't be able to continue the review.

SPEC.md
f-f: Sounds good! I was just remarking it so that we are all aware where the relevant clause is.

SPEC.md
f-f: Let's have some parsing in here, but we as you noted we shouldn't be too strict on the form of the URL. E.g. I think we can drop the `.git` suffix. (also we are going to implement special cases for SourceHut and GitHub, so we shouldn't worry too much about the details of their URLs)

SPEC.md
f-f: Are we noting anywhere in the spec that we do not accept updates lowering compiler versions? If not, we should note it here.

SPEC.md
f-f: I am not sure how much effort it would be to parse globbed paths, but if it's not much effort then it might make sense to do it? I don't have a strong opinion on this, just want to minimise the potential attack surface.

SPEC.md
thomashoneyman: Yes, above this — here's the link:
https://github.com/purescript/registry-dev/pull/533/files#diff-7426c9e3a694ca6015df5f98637912975f2edea23270203ee89a8bdeed246ee0R567

If that doesn't work, then here's the excerpt:

> A package set update is an object with two keys: `compiler`, an optional field that, if set, will update the compiler version used to compile the package sets (it cannot be downgraded), and `packages`, an object where keys are package names and values are either a version number or `null`. A version number indicates the package should be added to the set or updated to the given version, and `null` indicates the package should be dropped from the package set.

thomashoneyman: Some possible contention in the section directly above that:

>  Anyone can suggest a package set update to the registry. However, community members can only add or upgrade packages in the package set. To remove a package or downgrade a package version, the update must be submitted by a member of the packaging team or a Registry Trustee.

Should we allow packages to be downgraded? I left it in there in case a package version gets unpublished and there is no more-recent version available (so we'd be forced to downgrade it or drop it from the sets altogether).

thomashoneyman: If these two sections check out then this is good to go.

SPEC.md
thomashoneyman: I've updated this to note the relaxed parsing.

SPEC.md
thomashoneyman: It's not hard, with the restricted characters. I can do it in a followup.

thomashoneyman:
I believe everything checks out — potentially this still needs discussion:
https://github.com/purescript/registry-dev/pull/533#discussion_r1025834358

As followup tasks, we should:

1. Implement the Git URL parsing and the glob parsing from the spec
2. Enforce that compiler versions can be upgraded (not downgraded) in the package sets, and check that the upgrading logic works (we haven't used it yet).

SPEC.md
thomashoneyman: Well, maybe I spoke too soon. The easiest thing to do is follow the [POSIX portable filename character set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_276), then add the characters `/` and `*` for globbing.

But this is potentially too restrictive. Alternately, we could ban a selection of characters [such as these](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words), as well as common glob characters like `(` or `?`, and ones we definitely don't want, like `..`, and accept any other Unicode.

SPEC.md
thomashoneyman: I think this covers about everything the registry could fetch, ie. the `git` or `http(s)` protocols, and not the `ssh` or filesystem protocols

```regex
(git|http(s)?)(:(//)?)([\w\.@\:/\-~]+)(\.git)?(/)?
```

You can try on regexr with these tests:

```
# These should be accepted
https://github.com/user/project
https://github.com/user/project.git
http://github.com/user/project.git
https://192.168.101.127/user/project.git
http://192.168.101.127/user/project.git
git://host.xz/path/to/repo.git/
git://host.xz/~user/path/to/repo.git/
http://host.xz/path/to/repo.git/
https://host.xz/path/to/repo.git/

# These ones should not be accepted
git@github.com:user/project.git
git@192.168.101.127:user/project.git
ssh://user@host.xz:port/path/to/repo.git/
ssh://user@host.xz/path/to/repo.git/
ssh://host.xz:port/path/to/repo.git/
ssh://host.xz/path/to/repo.git/
ssh://user@host.xz/path/to/repo.git/
ssh://host.xz/path/to/repo.git/
ssh://user@host.xz/~user/path/to/repo.git/
ssh://host.xz/~user/path/to/repo.git/
ssh://user@host.xz/~/path/to/repo.git
ssh://host.xz/~/path/to/repo.git
/path/to/repo.git/
path/to/repo.git/
~/path/to/repo.git
file:///path/to/repo.git/
file://~/path/to/repo.git/
user@host.xz:/path/to/repo.git/
host.xz:/path/to/repo.git/
user@host.xz:~user/path/to/repo.git/
host.xz:~user/path/to/repo.git/
user@host.xz:path/to/repo.git
host.xz:path/to/repo.git
rsync://host.xz/path/to/repo.git/
```

SPEC.md
f-f: This is all good! Thanks for the clarifications.

f-f:
Sounds great! I will read through the spec once more before we merge.

f-f (approved):

PR #534, opened 2022-10-25
Move bower installer into test scripts

thomashoneyman:
In the monorepo migration it looks like the `BowerInstaller` script got put into the `lib` directory, but this is really just a test script used to produce the integration test. I've moved it into the test directory with this in mind.


f-f (approved):
Wonderful!

PR #536, opened 2022-10-25
Migrate app and lib directories

thomashoneyman:
This PR restructures the monorepo with the following changes:

1. The contents of `lib` have been moved into the `app` directory, indicating that almost all of these files have to do with the registry application and are unsuitable for use as a "library". As per the SPEC.md and `trh/new-codecs` branch, the plan is to then migrate library-specific code into `lib` in a few followup pull requests.
2. The `foreign` directory has been folded in to the `app` directory. I would like to be very careful about what we put in the `lib` directory, as that's our explicit public library, but the `foreign` code pretty much all exists to support the `app` so we may as well unify it.

Going forward, the idea is;

* `lib` contains the registry library, which consists of the data types referred to by the spec: package name, version, manifest, metadata, operation, constants, and so on
* `app` contains the registry application, ie. the HTTP server
* `scripts` contains scripts we intend to be run in CI or on a developer machine, ie. the legacy importer, the package transferrer, and so on.

If anyone feels strongly that `foreign` should remain its own directory then I can split it back out. I just don't want to have to mirror all the JS dependencies in two places.


api/spago.yaml
```diff
@@ -1,8 +0,0 @@
-package:
-  name: registry-api
-  publish:
-    license: BSD-3-Clause
-  run:
-    main: Registry.API
```
thomashoneyman, line 6: I had to remove this from the `app/spago.yaml` file because `spago test` would always try to execute this module @f-f. That of course failed because of the missing `GITHUB_EVENT_PATH` environment variable.

f-f: Uh, yeah. This was a bug I fixed yesterday together with all the cwd stuff. Would you like me to publish a new version and reinstate this?

f-f (approved):

api/spago.yaml
thomashoneyman: That would be great!

PR #537, opened 2022-10-25
Implement Sha256 from spec

thomashoneyman:
This implements the `SRIHash` data type referred to in the SPEC.md file.

(Ideally we merge #536 before this one, as this uses the new directory structure.)


app/src/Registry/API.purs
```diff
@@ -582,8 +582,8 @@ publish source { name, ref, compiler, resolutions } inputMetadata = do
     else
       log $ "WARNING: Package tarball is " <> show bytes <> ".\ncc: @purescript/packaging"
   log "Hashing the tarball..."
-  hash <- liftAff $ Hash.sha256File tarballPath
-  log $ "Hash: " <> show hash
+  hash <- liftAff $ SRIHash.hashFile tarballPath
+  log $ "Hash: " <> SRIHash.print hash
```
thomashoneyman: I have intentionally removed `show` from `SRIHash`. We shouldn't have even been using `show` outside of our tests, and with [`spec@7.1.0`](https://github.com/purescript-spec/purescript-spec/releases/tag/v7.1.0) we don't have to rely on `Show` in our tests either. So I am going to remove its use for each data type we move over to the registry library.

lib/src/Registry/SRIHash.purs
```diff
@@ -0,0 +1,89 @@
+-- | Implementation of the `SRIHash` data type from the registry spec. The
```
thomashoneyman: I am open to discussing whether `SRIHash` is the correct name for our hashes. I think it's better than `Sha256`, because we do include the subresource integrity prefix, and `SRISha256` is unwieldy...

lib/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "registry-lib",
+  "private": true,
+  "license": "BSD-3-Clause",
+  "dependencies": {}
```
thomashoneyman, line 5: I'd love to keep this list as small as possible!

app/src/Registry/API.purs
f-f: As I mentioned in #538, this is fine as long as another way to stringify these is provided

lib/src/Registry/SRIHash.purs
f-f: I would prefer `Sha256` to `SRIHash` - the name change would make sense if we were dealing with more hashing functions than just sha256, and I think the fact that it contains a sha256 is more important than the fact that it contains a SRI, we can document that.

lib/package.json
f-f: Why though? As long as there is even one we need to have an `npm install`, so having 5 vs 10 is not really about improving convenience

lib/src/Registry/SRIHash.js
```diff
@@ -1,3 +1,4 @@
+// Note: This is a Node runtime dependency, not a library dependency.
```
f-f: I am not following - why are we not specifying this in the `package.json`?

lib/src/Registry/SRIHash.purs
thomashoneyman: Works for me! I was thinking we might change the algorithm in the future but perhaps we won’t.

lib/package.json
thomashoneyman: Each additional NPM dependency is something to keep in sync between the registry and a package manager, and which has to be shipped with the package manager, and which has to be managed wrt versioning the library, so ideally there would be zero. If there’s good reason to add one then we will have to do it, but I want to be conservative.

lib/src/Registry/SRIHash.js
thomashoneyman: The comment is trying to say that “crypto” is part of Node’s core, so there is nothing to npm install or list in the package json file

thomashoneyman:
@f-f I've renamed back to `Sha256` and this should be ready for review again.

app/src/Registry/API.purs
thomashoneyman: Yep! As mentioned there, all of these types have a `print` function. By default all registry modules will have `parse`, `print`, and `codec`, and those which use `Parsing` will also export their `parser` so it can be reused as needed.

lib/src/Registry/SRIHash.purs
```diff
@@ -0,0 +1,97 @@
+-- | Implementation of the `Sha256` data type from the registry spec. The
+-- | registry records the hash of package tarballs in the package metadata so
+-- | that package managers can ensure the source code they download is correct.
+-- | https://github.com/purescript/registry-dev/blob/master/SPEC.md#Sha256
+module Registry.Sha256
```
colinwahl: I see the module name has been changed - should the filepath to these modules be updated as well?

thomashoneyman: Good catch, I forgot to rename the files. Thanks!

colinwahl (commented):
This looks good to me - I'll let Fabrizio have the final approval since there was a bit of discussion above that may not be resolved yet

lib/src/Registry/Sha256.js
```diff
@@ -1,3 +1,4 @@
+// Note: This is a Node runtime dependency, not a library dependency.
```
f-f:
```suggestion
// Note: the `crypto` module is a node builtin, not an external library.
```

f-f (approved):

lib/src/Registry/SRIHash.js
f-f: Ah, this makes sense! I proposed a change to clarify that (feel free to adjust/rewrite it) but otherwise this looks great!

PR #540, opened 2022-10-31
Migrate PackageName to registry library

thomashoneyman:
This implements the `PackageName` from the registry spec (see #533).

### DIFF
--- a/lib/src/Registry/PackageName.purs
+++ b/lib/src/Registry/PackageName.purs
@@ -1,42 +1,59 @@
+-- | Implementation of the `PackageName` data type from the registry spec. A
+-- | package name uniquely identifies a package.
+-- | https://github.com/purescript/registry-dev/blob/master/SPEC.md#packagename
 module Registry.PackageName
   ( PackageName
+  , codec
   , parse
+  , parser
   , print
   ) where
 
-import Registry.Prelude
+import Prelude
 
+import Control.Alt ((<|>))
 import Data.Array as Array
 import Data.Array.NonEmpty as NonEmptyArray
+import Data.Bifunctor (lmap)
+import Data.Codec.Argonaut (JsonCodec)
+import Data.Codec.Argonaut as CA
+import Data.Either (Either)
+import Data.Either as Either
+import Data.Maybe (Maybe(..), isJust)
 import Data.String as String
 import Data.String.CodeUnits as String.CodeUnits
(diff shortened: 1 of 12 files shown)


lib/src/Registry/PackageName.purs
```diff
@@ -1,42 +1,59 @@
...
+codec = CA.prismaticCodec "PackageName" (Either.hush <<< parse) print CA.string
 
+-- | Print a package name as a string
 print :: PackageName -> String
 print (PackageName package) = package
 
-parse :: String -> Either ParseError PackageName
-parse inputStr = Parsing.runParser inputStr do
+-- | Parse a package name from a string, reporting errors on failure
+parse :: String -> Either String PackageName
```
thomashoneyman, line 48: We always print the parser error, so I plan to have the `parse` functions always error as a string. If we want to get the parser error for some reason we can just use `runParser PackageName.parser` directly.

lib/src/Registry/PackageName.purs
```diff
@@ -83,8 +100,9 @@ parse inputStr = Parsing.runParser inputStr do
     allChunks = Array.concatMap NonEmptyArray.toArray (Array.cons firstChunk nextChunks)
     name = String.CodeUnits.fromCharArray allChunks
 
-  -- and check that it's not longer than 50 chars
-  if String.length name > 50 then
-    Parsing.fail "Package name cannot be longer than 50 chars"
+  if String.null name then
+    Parsing.fail "Package name cannot be empty"
```
thomashoneyman, line 104: This is the only significant code change: we previously restricted package names to be at least 2 characters long. However, we based things like the registry index on package name lengths as little as 1 character, following Rust's lead. In the spec I wrote that package names have to be at least 1 character long.

@f-f is there a reason to disallow single-character package names? If so, I'd like to include that rationale in the registry spec. If not, then we can keep this code as-is, and the registry index will continue to work -- it already assumes the existence of single-character packages.

thomashoneyman: Whoops – now that I look at this again, I must be misremembering the "2 characters" limitation. I could have sworn we had it written somewhere, but on review we don't enforce it in the parser and didn't have it in the spec, so I this isn't a change at all!

colinwahl (approved):
I don't remember specifics about a character limit - but other than that the implementation looks good

PR #541, opened 2022-10-31
Implement 'Version' and 'Range' types from the spec

thomashoneyman:
This PR implements the `Version` and `Range` types described in the spec (#533). The change is almost entirely mechanical; move over the data types, parser, and printer, and implement a JSON codec. However, I've made some decisions that I think notably improve how we use these types.

First, `Version` and `Range` are implemented in their own modules instead of sharing a module. This means that we no longer have to disambiguate version and range operations within the namespace, such as `Version.parseVersion` and `Version.parseRange`. It also means that range-specific operations get their own namespace (no more `Version.union` or `Version.intersect`, which doesn't make any sense since those are _range_ operations).

Second, there is no longer a `ParseMode` usable with these types; instead, the registry _only_ uses strict versions and ranges, and the application offers its own `LenientVersion` and `LenientRange` types. This is an improvement for a lot of reasons!

1. We never allow lenient versions in the core registry (things like a `Manifest`) -- but right now it's up to developer discipline to not accidentally use `Version.parseVersion Version.Lenient` somewhere only strict versions should be allowed. A dedicated `LenientVersion` means you _definitely_ expect to be dealing with non-strict versions, which is a big heads-up that this is a place to be careful in the code.
2. We only use lenient versions and ranges when dealing with legacy formats, which is strictly a registry application concern; contorting around these types literally everywhere else in the application feels like a bad code smell. 
3. We already use these types in error. A function like `Range.union` makes no sense in the lenient context; you no longer have one "input" you parsed from. And there are places where we use `Version.rawVersion` to indicate we want to print a raw version, but it could only ever have been possible to have a strict version (ie. `Version.printVersion == Version.rawVersion`); this is confusing, because either we _did_ want a raw version, in which case this could be a bug, or we didn't, in which case `rawVersion` should not have been used.

It's much nicer to just separate these concerns altogether, and so I've done just that. But I've kept all the same tests and ensured they all continue to pass.


app/src/Registry/API/LenientRange.purs
```diff
@@ -0,0 +1,149 @@
+module Registry.API.LenientRange
```
thomashoneyman: Some data types are going to begin to conflict between the registry library (ie. the `Registry` namespace) and the registry application. I've opted to begin placing app-specific modules into `Registry.API` as that's what we generally refer to it as (the "registry API"), but perhaps `Registry.App` would be better -- I'm open to suggestions!

lib/src/Registry/Version.purs
```diff
@@ -0,0 +1,134 @@
...
+  where
+  nonNegativeInt :: Parser String Int
+  nonNegativeInt = do
+    digitChars <- Parsing.Combinators.Array.many1 Parsing.String.Basic.digit
+    let
+      zeroCount = Array.length $ NonEmptyArray.takeWhile (_ == '0') digitChars
+      digitString = CodeUnits.fromCharArray $ NonEmptyArray.toArray digitChars
+      failInteger = Parsing.fail $ "Invalid 32-bit integer: " <> digitString
+    integer <- maybe failInteger pure $ Int.fromString digitString
+    -- We do not accept leading zeros in versions unless we are in lenient mode
```
thomashoneyman:
```suggestion
    -- We do not accept leading zeros in versions
```

scripts/src/LegacyImporter.purs
```diff
@@ -238,6 +246,7 @@ type ImportedIndex =
   , failedVersions :: Map RawPackageName (Map RawVersion VersionValidationError)
   , reservedPackages :: Map PackageName Location
   , registryIndex :: RegistryIndex
+  , packageRefs :: Map PackageName (Map Version RawVersion)
```
thomashoneyman, line 249: We used to use lenient version parsing when parsing a legacy manifest as a `Manifest`. Really, this is conflating two concerns: producing a valid `Manifest` and persisting the `ref` for the package. Since we've got the `ref` information available right here, it's not much trouble to just persist what all the original refs were for the packages we're processing (you can see that I do that in the `packageRefs` on line 274).

scripts/src/LegacyImporter.purs
```diff
@@ -350,23 +371,20 @@ buildLegacyPackageManifests existingRegistry legacyPackageSets rawPackage rawUrl
...
         _ -> do
           let key = "manifest__" <> PackageName.print package.name <> "--" <> tag.name
           liftEffect (Cache.readJsonEntry key cache) >>= case _ of
-            -- We can't just write the version directly, as we need to preserve the _raw_ version.
-            -- So we update the manifest fields before reading/writing the cache.
             Left _ -> ExceptT do
               log $ "CACHE MISS: Building manifest for " <> PackageName.print package.name <> "@" <> tag.name
               manifest <- Except.runExceptT buildManifest
-              liftEffect $ Cache.writeJsonEntry key (map (_ { version = Version.rawVersion version } <<< un Manifest) manifest) cache
+              liftEffect $ Cache.writeJsonEntry key manifest cache
```
thomashoneyman, line 384: Now that we're holding on to the `packageRefs` we don't have to futz around with what we're writing to cache. We just write the manifest as-is.

thomashoneyman:
As far as testing goes, beyond the unit tests I've also run the registry importer and verified that the import statistics are unchanged. This code change does not appear to affect the behavior of the registry.

app/src/Registry/API/LenientRange.purs
colinwahl: I think the "registry API" is a bit conflated, because you have two distinct things:

* The registry service we will run that handles API calls - which uses this module in question
* The registry library which itself exposes an API

Given that, maybe `Registry.App` is more clear so that you can't get it confused with something that's being exported by a library API.

scripts/src/LegacyImporter.purs
colinwahl: Yeah, that's a good improvement.

app/src/Registry/Legacy/Manifest.purs
```diff
@@ -212,12 +214,9 @@ validateLicense licenses = do
 validateDependencies :: Map RawPackageName RawVersionRange -> Either LegacyManifestValidationError (Map PackageName Range)
 validateDependencies dependencies = do
   let
-    parsePackageName = PackageName.parse <<< stripPureScriptPrefix
-    parseVersionRange = lmap Parsing.parseErrorMessage <<< Version.parseRange Version.Lenient
```
colinwahl, line 216: Previously this code parsed the version range in `Lenient` mode and now it uses `Range.parse` - is that intentional?

thomashoneyman: No, this should be using lenient parsing on a raw version range. I've updated it.

app/src/Registry/API.purs
```diff
@@ -52,6 +52,7 @@ import Node.FS.Stats as FS.Stats
 import Node.FS.Sync as FS.Sync
 import Node.Path as Path
 import Node.Process as Node.Process
+import Registry.App.LenientVersion as LenientVersion
```
colinwahl, line 55: Related to this change, I wonder if we should just rename this module to `Registry.App`. I could go either way though.

app/test/Fixture/Manifest.purs
```diff
@@ -28,7 +29,7 @@ instance Fixture PackageName where
   fixture = unsafeFromRight $ PackageName.parse "fixture-package-name"
 
 instance Fixture Version where
-  fixture = unsafeFromRight $ Version.parseVersion Version.Lenient "1.0.0"
+  fixture = unsafeFromRight $ Version.parse "1.0.0"
```
colinwahl, line 32: Similarly these switched from `Lenient` to `Version.parse`, but I assume it's fine since it's just a test?

colinwahl: Oh wait, `LenientVersion` wouldn't even typecheck here.

app/test/Registry/API/LenientVersion.purs
```diff
@@ -0,0 +1,42 @@
+module Test.Registry.App.LenientVersion (spec) where
```
colinwahl: The module names here were changed but the filepaths are still `app/test/Registry/API`

app/test/Fixture/Manifest.purs
thomashoneyman: Yep! These switched because this is for tests, exactly.

app/src/Registry/API.purs
thomashoneyman: Yeah, I think we may want to rename this in the future as we gain a HTTP API and other things running on a server (like the job to compile packages to determine what range of compilers they work with). But I'd like to keep this as-is for the time being.

lib/test/Registry/Version.purs
```diff
@@ -0,0 +1,49 @@
...
+  , "1.2.3"
+  , "1000.1111.1234"
+  ]
+
+invalid :: Array { version :: String, label :: String }
+invalid =
+  [ { version: ".12.0", label: "Malformed" }
+  , { version: "12.0", label: "Too few places" }
+  , { version: "0.12.12.1", label: "Too many places" }
+  , { version: "1.0.1-rc.1", label: "Contains prerelease identiiers" }
```
colinwahl:
```suggestion
  , { version: "1.0.1-rc.1", label: "Contains prerelease identifiers" }
```

app/test/Fixture/Manifest.purs
thomashoneyman: It wouldn't typecheck, but we could do `LenientVersion.version $ unsafeFromRight $ LenientVersion.parse "1.0.0"` if desired. Even so -- this is in the registry library tests, so we aren't using lenient versions.

colinwahl: I think it's good as-is

colinwahl (approved):
Thanks for putting in the effort that this needed - excellent work!

thomashoneyman:
💪 I'll let this sit for a day or two so that @f-f can take a look since it affects his Spago stuff.

f-f (approved):

f-f:
Thanks @thomashoneyman, great work! 👏

PR #542, opened 2022-11-01
Implement 'License' from spec

thomashoneyman:
This PR implements the registry `License` type from the spec (#533). Notably, as with the version & range changes in #541, I have added a small amount of separation between the way the application may choose to deal with malformed licenses and how the stricter registry library does.

Specifically, the `License` parsing no longer tries to bulk up its error message by fuzzy-matching on license identifiers with the Fuse library and simply fails with an 'invalid SPDX license' error message. However, since being able to fuzzy-match and report a suggestion is useful, I've preserved this behavior in the library proper.

You can use `License.parse` for standard license parsing, and you can use `fuzzyMatchLicense` to try and match with valid licenses that the user could have meant. The registry API can look for JSON parsing failures arising from the `license` field and elaborate on the error message if it wants to, and Spago can use this function to suggest alternatives if a config is incorrect.

There is no reason that consumers of the registry library should have to bring in the several libraries necessary to make the fuzzy matcher work when it is only used to add a suggestion to the error message; the fuzzy matching is really best suited for the registry application.

### DIFF
--- a/lib/test/Registry/PackageName.purs
+++ b/lib/test/Registry/PackageName.purs
@@ -23,12 +23,12 @@ spec = do
 
       formatError (Tuple name error) = name <> ": " <> error
 
-      { fail } = Utils.partitionEithers $ map parse goodPackageNames
+      { fail } = Utils.partitionEithers $ map parse valid
 
     unless (Array.null fail) do
       Assert.fail $ String.joinWith "\n"
         [ "Some well-formed package names were not parsed correctly:"
-        , String.joinWith "\n  - " $ map formatError fail
+        , Array.foldMap (append "\n  - " <<< formatError) fail
         ]
 
   Spec.it "Fails to parse malformed package names" do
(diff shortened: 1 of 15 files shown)


lib/test/Registry/PackageName.purs
```diff
@@ -23,12 +23,12 @@ spec = do
 
       formatError (Tuple name error) = name <> ": " <> error
 
-      { fail } = Utils.partitionEithers $ map parse goodPackageNames
+      { fail } = Utils.partitionEithers $ map parse valid
```
thomashoneyman, line 26: The changes in this module are only to match with the `valid` / `invalid` convention for iterating over multiple tests, and to use `foldMap` to slightly improve error output. It's just cosmetic and doesn't affect behavior.

colinwahl (approved):
The reorganization makes sense to me 👍

f-f:
I think it would be useful to ship the fuzzy matching with the library too. It might not be a big deal in this case though since it's only a few lines.

Still, I would like if we could wait a little more than a few hours before merging these PRs - very many things are being shuffled around and this almost feels like a rewrite. I get that we want to move quickly but I would like some time to look at the code and think, to not lose familiarity with it.

PR #544, opened 2022-11-03
Update Nix dependencies

thomashoneyman:
This updates to the latest Spago version and adds PureScript 0.15.6 to the available compilers list. The new Spago version executes from the root directory when you use `spago run`, so I've removed the "rootDir" workaround introduced in #536.

### DIFF
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -1,5 +1,7 @@
 package:
   name: registry-app
+  run:
+    main: Registry.API
   publish:
     version: 0.0.1
     license: BSD-3-Clause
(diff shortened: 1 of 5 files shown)


thomashoneyman:
@f-f Oddly enough, with the new update if I run `spago build -p registry-lib` I get some errors like this:

```console
Error 12 of 12:

  in module Test.Registry
  at lib/test/Registry.purs:13:1 - 13:39 (line 13, column 1 - line 13, column 39)

    Module Test.Spec.Runner was not found.
    Make sure the source file exists, and that it has been provided as an input to the compiler.


  See https://github.com/purescript/documentation/blob/master/errors/ModuleNotFound.md for more information,
  or to contribute content related to this error.
```

...even though the `lib/spago.yaml` file lists this under the `test_dependencies` key. Did that key change, potentially?

f-f (approved):
Looking great!

PR #549, opened 2022-11-04
Implement schema types according to spec (Manifest, Metadata, Owner, Location, PackageSet)

thomashoneyman:
This PR implements the remaining schema types from the spec (#533). In each case I've migrated tests where they existed, or added new tests -- for example, there are new round-trip tests for the package sets and metadata schemas.

Along the way I've needed to make a few adjustments to match the spec.

1. The `RFC3339` type is no longer used in the core library. Instead, we use normal `DateTime` values and have formatters for the simple `ISO8601` format used in the ECMAScript spec. This is what our native backend (JS) provides explicit support for and we switch to use it across the board. Since we already have artifacts using the old format I have provided a function to migrate the formats and I have tested that it successfully migrates timestamps in our metadata files and the cache.
2. The `description` field in the manifest and the `reason` field in the unpublished metadata now have a length limit of 300 characters. This was somewhat arbitrarily chosen, but it's what is in the spec; we should discuss on the spec issue whether this is an appropriate limit.

Speaking of timestamps, I noticed that we were being inconsistent about whether we used the host machine's time or used UTC. Since all of our timestamps assumed UTC, I've switched to use `nowUTC` to explicitly get the correct UTC time any time we get the current time. Otherwise, our timestamps can be off by a number of hours.

One last thing: I've separated the `PackageSet` type from various operations taken on the package sets, such as compiling batches or syncing legacy package sets. The core type has moved to the library, and everything else is in the `App` directory.

To verify these changes I have added a number of unit tests. I've also ensured that the `PublishPursuit` script still works, and I've run the registry importer and registry package set updater to verify they continue to work (and aren't changing the metadata other than fixing the timestamps). Everything looks good.

After this merges, we will need to update the metadata in the `registry` repository to use the new timestamps. That diff looks like this:

```diff
     "0.3.3": {
       "bytes": 61956,
       "hash": "sha256-hW753qS4YbDnqsrMO6DRSkVAKKePf9C7lbToaw63ka4=",
-      "publishedTime": "2022-04-24T17:50:46.0Z",
+      "publishedTime": "2022-04-24T17:50:46.000Z",
       "ref": "v0.3.3"
```


app/src/Registry/API.purs
```diff
@@ -569,8 +567,8 @@ publish source { name, ref, compiler, resolutions } inputMetadata = do
 
   -- Then, we can run verification checks on the manifest and either verify the
   -- provided build plan or produce a new one.
-  verifyManifest { metadata, manifest }
-  eitherBuildPlan <- verifyBuildPlan { source, buildPlan: BuildPlan { compiler, resolutions }, manifest }
+  verifyManifest { metadata: Metadata metadata, manifest }
+  eitherVerifiedResolutions <- verifyResolutions { source, resolutions, manifest }
```
thomashoneyman, line 580: The `BuildPlan` is a bit weird, because several of the functions using it only need the compiler version or the resolutions (for example, no compiler version is needed here). I've unfolded it into separate `compiler` and `resolutions` fields.

app/src/Registry/API.purs
```diff
@@ -674,31 +672,31 @@ verifyManifest { metadata, manifest } = do
   -- are going to fail while parsing from JSON, so we should move them here if we
   -- want to handle everything together
   log "Running checks for the following manifest:"
-  log $ Json.printJson manifestFields
```
thomashoneyman, line 686: This prints the fields in alphabetical order, which we don't really want. Using the registry codec is better, because the fields are ordered.

app/src/Registry/API.purs
```diff
@@ -883,20 +886,21 @@ compilePackage { packageSourceDir, buildPlan: BuildPlan plan } = do
...
 -- | packages where the `purescript-` prefix is still present.
 publishToPursuit :: PublishToPursuit -> RegistryM (Either String String)
-publishToPursuit { packageSourceDir, dependenciesDir, buildPlan: buildPlan@(BuildPlan { compiler }) } = Except.runExceptT do
+publishToPursuit { packageSourceDir, dependenciesDir, compiler, resolutions } = Except.runExceptT do
   log "Generating a resolutions file"
   tmp <- liftEffect Tmp.mkTmpDir
 
   let
-    resolvedPaths = buildPlanToResolutions { buildPlan, dependenciesDir }
+    resolvedPaths = formatPursuitResolutions { resolutions, dependenciesDir }
```
thomashoneyman, line 912: I've renamed this function because this really has to do with formatting a file for Pursuit.

app/src/Registry/Legacy/Manifest.purs
```diff
@@ -375,15 +374,3 @@ fetchLegacyPackageSets = do
         pure contents.value
 
   pure legacySets
-
-filterLegacyPackageSets :: Array GitHub.Tag -> Array String
```
thomashoneyman, line 379: This was dead code.

lib/src/Registry/Location.purs
```diff
@@ -0,0 +1,85 @@
...
+-- | and 'repo' for convenience.
+githubCodec :: JsonCodec GitHubData
+githubCodec = Profunctor.dimap toJsonRep fromJsonRep $ CA.Record.object "GitHub"
+  { githubOwner: CA.string
+  , githubRepo: CA.string
+  , subdir: CA.Record.optional CA.string
+  }
+  where
+  toJsonRep { owner, repo, subdir } = { githubOwner: owner, githubRepo: repo, subdir }
+  fromJsonRep { githubOwner, githubRepo, subdir } = { owner: githubOwner, repo: githubRepo, subdir }
```
thomashoneyman, line 66: I've preserved this since it's in the existing codebase. But I'm happy to just remove it so that the PureScript uses `{ githubOwner, githubRepo }` the same as the input JSON. Let me know what you prefer @f-f, @colinwahl.

lib/src/Registry/Manifest.purs
```diff
@@ -0,0 +1,83 @@
...
+codec = Profunctor.wrapIso Manifest $ CA.object "Manifest"
+  $ CA.recordProp (Proxy :: _ "name") PackageName.codec
+  $ CA.recordProp (Proxy :: _ "version") Version.codec
+  $ CA.recordProp (Proxy :: _ "license") License.codec
+  $ CA.recordPropOptional (Proxy :: _ "description") (Internal.Codec.limitedString 300)
+  $ CA.recordProp (Proxy :: _ "location") Location.codec
+  $ CA.recordPropOptional (Proxy :: _ "owners") (CA.nonEmptyArray Owner.codec)
+  $ CA.recordPropOptional (Proxy :: _ "files") (CA.nonEmptyArray CA.nonEmptyString)
+  $ CA.recordProp (Proxy :: _ "dependencies") (Internal.Codec.packageMap Range.codec)
+  $ CA.record
```
thomashoneyman, line 83: This weird style is done to manually control the order of fields in the resulting JSON.

app/src/Registry/API.purs
```diff
@@ -1496,3 +1501,28 @@ syncLegacyRegistry package location = do
...
+unpublishVersionInMetadata version versionMeta (Metadata metadata) = do
+  let published = Map.delete version metadata.published
+  let unpublished = Map.insert version versionMeta metadata.unpublished
+  Metadata (metadata { published = published, unpublished = unpublished })
+
+isVersionInMetadata :: Version -> Metadata -> Boolean
+isVersionInMetadata version (Metadata metadata) = versionPublished || versionUnpublished
+  where
+  versionPublished = isJust $ Map.lookup version metadata.published
+  versionUnpublished = isJust $ Map.lookup version metadata.unpublished
```
thomashoneyman, line 1537: I've put these in the API for now, but I can see us wanting it to be in the `Registry.Metadata` instead. I'm not sure whether we should keep them, though, or have this be tied in to `RegistryM` so this stays in sync with the metadata in memory.

f-f: Spago does look up in the metadata, so I would say that we could expose the functions for reading things out. On the other hand, this is pretty short and straightforward code, so I don't think we have to necessarily share it.

lib/src/Registry/Location.purs
f-f: I'm fine with having `owner` and `repo` in PureScript, these names with prefixes are just for encoding and we're preserving them so it's all good!

lib/src/Registry/Manifest.purs
f-f: It looks alright - let's add this as a comment! Probably also worth pointing out how things are sorted (I suspect exactly as it is in the code rather than backwards

f-f (approved):

lib/src/Registry/Manifest.purs
thomashoneyman: There's a comment that says why we're using this instead of record sugar ("to have an explicit ordering"). And you're correct, it's sorted exactly as you see in the code!

PR #551, opened 2022-11-05
Migrate Constants into registry library

thomashoneyman:
A small one! This moves the `Constants` module into the registry library.


app/src/Registry/API.purs
```diff
@@ -1085,13 +1088,13 @@ fetchRegistryIndex :: RegistryM Unit
 fetchRegistryIndex = do
   registryIndexPath <- asks _.registryIndex
   log "Fetching the most recent registry index..."
-  liftAff $ fetchRepo Constants.registryIndexRepo registryIndexPath
+  liftAff $ fetchRepo Constants.packageIndex registryIndexPath
```
thomashoneyman, line 1100: I've applied the renaming described in the spec (#533) -- ie. "package index", "package sets", "package metadata" instead of "registry index", "package sets", "metadata". Keeps things a bit more uniform!

lib/src/Registry/Constants.purs
```diff
@@ -0,0 +1,37 @@
+module Registry.Constants where
+
+import Data.Maybe (Maybe(..))
+import Node.Path (FilePath)
+import Registry.Location (GitHubData)
```
thomashoneyman, line 5: This PR relies on #549 and should be merged afterwards.

f-f (approved):

PR #552, opened 2022-11-05
Implement 'Operations' according to spec

thomashoneyman:
This PR migrates the 'Operations' module into the registry library and adds tests to verify the JSON encoding and decoding. I've opted not to move over the fixtures for the GitHub events that decode into particular operations because that's really a concern of the registry application, and we'll stop using GitHub events once we have a legitimate HTTP API. Instead I've just tested the payload body we expect to receive over the wire.

With this change we're also very close to being able to remove the giant `Registry.Json` module altogether, if we want to.

See also the spec (#533) and the schema PR (#549) which must be merged before this one.


app/src/Registry/API.purs
```diff
@@ -122,15 +123,17 @@ main = launchAff_ $ do
...
-          Publish { name, ref } ->
-            "Publishing package `" <> PackageName.print name <> "` at the ref `" <> ref <> "`."
-          PackageSetUpdate _ ->
-            "Processing package set update."
-          Authenticated (AuthenticatedData { payload }) -> case payload of
-            Unpublish { name, version } ->
-              "Unpublishing `" <> PackageName.print name <> "` at version `" <> Version.print version <> "`."
-            Transfer { name } ->
-              "Transferring `" <> PackageName.print name <> "`."
+          Left packageSetOperation -> case packageSetOperation of
```
thomashoneyman, line 126: There is no longer a unified "operation", as detailed in the spec. Instead we have "package set" operations and "package" operations, and the former will be served by GitHub Issues and the latter by a HTTP API, so there's no need for a unified Operation type anymore. This Left / Right encoding only exists until we have the HTTP API, at which point there won't be a unified "runOperation" function anymore. (The equivalent would be the router for the HTTP API, which again won't deal with package sets).

f-f (approved):

PR #553, opened 2022-11-05
Fix subdir reference in fetchRepo

thomashoneyman:
Calls to `fetchRepo` are failing when they include a `subdir` key. The intent was to prevent us from mistakenly calling this and cloning into a subdirectory before there is support for that, or expecting that we _will_ go into a subdirectory when we won't, but alas it's very easy to accidentally crash by supplying a subdir.

Instead, we just don't accept one at all.

### DIFF
--- a/lib/src/Registry/Constants.purs
+++ b/lib/src/Registry/Constants.purs
@@ -1,37 +1,29 @@
 module Registry.Constants where
 
-import Data.Maybe (Maybe(..))
 import Node.Path (FilePath)
-import Registry.Location (GitHubData)
+
+type GitHubRepo = { owner :: String, repo :: String }
 
 -- | The location of the registry GitHub repository
-registry :: GitHubData
-registry = { owner: "purescript", repo: "registry", subdir: Nothing }
-
--- | The location of the package sets within the registry GitHub repository
-packageSets :: GitHubData
-packageSets = registry { subdir = Just packageSetsDirectory }
+registry :: GitHubRepo
+registry = { owner: "purescript", repo: "registry" }
 
 -- | The file path to the package sets within the registry GitHub repository
 packageSetsDirectory :: FilePath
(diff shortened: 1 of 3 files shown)


f-f:
I'm confused by this - why are we using a `subdir` at all? Maybe these constants should not be of type `GitHubData` in the first place?

thomashoneyman:
Oh! Yeesh, I wrote it backwards! This is a terrible patch — I’ll push up a change.

The answer is that we don’t use the subdir key for these two repos since they don’t need it. Other locations — such as the package sets location — do use a subdir since they’re located in a subdirectory.

But you’re right, we probably shouldn’t be using GitHubData because these aren’t packages.

lib/src/Registry/Constants.purs
```diff
@@ -1,37 +1,29 @@
 module Registry.Constants where
 
-import Data.Maybe (Maybe(..))
 import Node.Path (FilePath)
-import Registry.Location (GitHubData)
 
--- | The location of the registry GitHub repository
-registry :: GitHubData
-registry = { owner: "purescript", repo: "registry", subdir: Nothing }
+type GitHubRepo = { owner :: String, repo :: String }
```
thomashoneyman, line 5: I don't love introducing this type here, but we don't depend on foreign GitHub bindings in the library and have no need for the GitHub module, so we don't use the `Address` type it has. It's such a small type that it's easy to put here.

An alternative might be to just have these be:

```purs
registryRepo :: String
registryRepo = "https://github.com/purescript/registry.git"
```

and therefore avoid the extra type definition. We only use this for cloning, anyway. What do you think, @f-f?

f-f: I think introducing this type here is pretty fine

f-f (approved):

PR #554, opened 2022-11-09
Implement manifest index in registry-lib

thomashoneyman:
This pull request implements the `ManifestIndex` type (the old `RegistryIndex`). The difference between the new and old types are that:

1. The `ManifestIndex` is a smart constructor, where if you have a `ManifestIndex` you _know_ that it is valid wrt packages only depending on other packages in the index. There is no more `checkIndex` pass that you have to do before using a given registry index to ensure it's actually valid – if you have a `ManifestIndex` then you know it is valid.
2. There are more pure functions available for working with a manifest index, such as `fromSet` to parse a `Set Manifest` into a `ManifestIndex`, `maximalIndex` to produce the largest possible self-contained index from an input set, `toSortedArray` to produce a topologically-sorted array, `toMap` to produce a `Map PackageName (Map Version Manifest)`, `lookup` to look up a specific package version, `insert` to attempt to insert a package into the index, failing if it would make the index invalid, and so on.
3. There is a separate `topologicalSort` function that works just on `Set Manifest` and does not require an index. It's reused by index-specific functions, but it's also available without having to create that structure.

The biggest design question here is around whether a `ManifestIndex` should consider version bounds or not. Right now I've kept parity with the old version by _not_ considering version bounds: an index is "valid" if all packages depend on other packages in the index, at _any_ version. However, I think it would be nice to adjust this so that:

1. A `ManifestIndex` is valid if all packages depend on other packages in the index _at versions that exist in the index_.
2. We introduce a new `maximalIndexIgnoringBounds` function that ignores version bound when constructing an index, but does not produce a `ManifestIndex` (it would just be a `Map PackageName (Map Version Manifest)`). We need this, for example, to verify that the package sets are self-contained -- they explicitly ignore version bounds.

In this case, we would drop the 38 unsolvable package versions from the registry to allow the index to be more strict. It's a small change to the code and I've left comments describing how to do it, but I think that should be done in a separate PR so we can better evaluate its impact.

**Testing**

This is passing the existing `Registry.Index` tests, as well as the new set of tests in `Test.Registry.ManifestIndex`. I've also confirmed that `registry-importer dry-run` properly reads in the index and executes.


lib/test/Registry/ManifestIndex.purs
```diff
@@ -0,0 +1,256 @@
...
+import Registry.Range as Range
+import Registry.Version (Version)
+import Registry.Version as Version
+import Test.Assert as Assert
+import Test.Spec (Spec)
+import Test.Spec as Spec
+import Test.Utils (unsafeManifest)
+import Test.Utils as Utils
+
+-- TODO: Write, read, insert, remove file system operations.
```
thomashoneyman: Technically this is already checked by `Registry.App`, but I might like to add tests here instead.

lib/src/Registry/ManifestIndex.purs
```diff
@@ -0,0 +1,314 @@
...
+-- | Convert a set of manifests into a `ManifestIndex`. Reports all failures
+-- | encountered rather than short-circuiting.
+fromSet :: Set Manifest -> Either (Map PackageName (Map Version (Map PackageName Range))) ManifestIndex
+fromSet manifests = do
+  let Tuple failed index = maximalIndex manifests
+  if Map.isEmpty failed then Right index else Left failed
+
+-- | Produce the maximal `ManifestIndex` possible for the given set of
+-- | `Manifest`s, collecting failures along the way.
+maximalIndex :: Set Manifest -> Tuple (Map PackageName (Map Version (Map PackageName Range))) ManifestIndex
```
thomashoneyman, line 142: You can compare this function to the obscenity that was `checkIndex` for a comparison.

lib/test/Utils.purs
```diff
@@ -27,14 +47,47 @@ partitionEithers = Array.foldMap case _ of
...
+      , subdir: Nothing
+      }
+  , description: Nothing
+  , owners: Nothing
+  , files: Nothing
+  }
+
+-- | Format a package version as a string in the form 'name@X.Y.Z'
+formatPackageVersion :: PackageName -> Version -> String
+formatPackageVersion name version = PackgaeName.print name <> "@" <> Version.print version
```
f-f:
```suggestion
formatPackageVersion name version = PackageName.print name <> "@" <> Version.print version
```

lib/test/Utils.purs
```diff
@@ -4,14 +4,34 @@ import Prelude
...
+import Data.Maybe (Maybe(..))
+import Data.Tuple (Tuple)
+import Partial.Unsafe (unsafeCrashWith)
 import Partial.Unsafe as Partial
+import Registry.License as License
+import Registry.Location (Location(..))
+import Registry.Manifest (Manifest(..))
+import Registry.PackageName (PackageName)
+import Registry.PackageName as PackageName
+import Registry.PackageName as PackgaeName
```
f-f:
```suggestion
```

f-f (approved):
Just a typo to fix, otherwise looks great!

Issue #523, opened 2022-09-29
Migrate operation validation into library


thomashoneyman:
Currently all validation for the `Operation` type is done via the JSON payload decoding or via checks implemented in the `API` module. However, we'll soon be moving the API module out of the `src` directory so it is not exposed as a library. We still want the operation verification to be runnable locally via Spago, however. With this in mind, below is my proposal for a new `Registry.Operation` module.

---

## Operation Checks

Each registry operation is guarded by validation before its effects are executed (e.g. a package is uploaded, metadata is written). Since the registry runs asynchronously it's critical that this validation can also be run by package managers. And since the registry and package managers will run in different environments, it's critical that effects are injected. Some effects include:

* License detection
* Logging
* Reading metadata, registry index files

The idea is that when you run:

```sh
$ spago publish
$ spago unpublish --version 0.7.0 --reason "Committed credentials."
$ spago transfer
```

then Spago will execute the same validation that the registry would run for each by depending on the registry as a library and providing its own effects. Here I am also making two assumptions:

1. That we are going to unify the Addition and Update operations into a single Publish operation
2. That we are going to preserve package set publishing as something coordinated via GitHub issues, not via an API running on a server (and therefore checks do not need to be exported for Spago, though we could do so).

### Unauthenticated Operations

For unauthenticated operations, the registry begins by verifying the JSON is well-formed. Package managers do not need to do this check, but they need to produce the correct JSON. 

#### Publish

`{ "package": "prelude", "location"?: ..., "ref": ... }`

At first we only have the API payload. We use it to attempt to read the package metadata, which we then verify. To be publishable, either:

1. The package name is not registered (ie. no metadata exists), a location is provided, and the location is not already in use. If the package is new (no metadata exists) then metadata should be created.
3. The package name is registered (ie. metadata exists), and one of:
		1. No location is provided 
		2. The provided location matches the package metadata.
		4. The provided location supports the `subdir` key, and all fields except the `subdir` key matches the package metadata.

At this point we have the API payload and package metadata. We know that the package is publishable. Next, we fetch the package source and verify it.

1. The package must contain a `src` directory containing`.purs` files.
2. The package must contain a valid `purs.json` manifest or `spago.yaml` manifest. In the presence of multiple manifest files the `purs.json` file takes precedence and all others are ignored.

Now we have a valid manifest we can ensure that the API call, manifest, metadata, and package sources are reconciled (these can be in parallel):

1. The manifest package name and API package name must match.
[...]

PR #555, opened 2022-11-10
Migrate sanitizePath into lib

colinwahl:
I need to use `sanitizePath` in my work for operations validations for #523 - it is pure code that was located in the `Foreign.FastGlob` module previously, so I pulled it out into a new lib module `Registry.Internal.Path` for use within the lib directory itself.

@thomashoneyman has said he'll take a look at migrating tests over once this merges. Right now, there are tests exercising the whole `FastGlob.match` but none for this specific utility itself.

### DIFF
--- a/lib/src/Registry/Internal/Path.purs
+++ b/lib/src/Registry/Internal/Path.purs
@@ -5,6 +5,8 @@ module Registry.Internal.Path
 
 import Prelude
 
+import Control.Monad.Except (ExceptT(..), runExceptT)
+import Data.Bifunctor (lmap)
 import Data.Compactable (separate)
 import Data.Either (Either(..))
 import Data.Maybe (Maybe(..))
@@ -15,7 +17,6 @@ import Effect.Aff as Aff
 import Node.FS.Aff as FS
 import Node.Path (FilePath)
 import Node.Path as Path
-import Partial.Unsafe (unsafeCrashWith)
 
 type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array FilePath }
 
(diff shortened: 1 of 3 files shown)


f-f (approved):

lib/src/Registry/Internal/Path.purs
```diff
@@ -0,0 +1,41 @@
...
+import Effect.Aff (Aff)
+import Effect.Aff as Aff
+import Node.FS.Aff as FS
+import Node.Path (FilePath)
+import Node.Path as Path
+import Partial.Unsafe (unsafeCrashWith)
+
+type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array FilePath }
+
+sanitizePaths :: FilePath -> Array FilePath -> Aff SanitizedPaths
```
thomashoneyman: Can we comment this? Specifically, we could comment it to say something like:

```
Given a directory, validate a collection of file paths relative to that root to ensure
no file path traverses out of the directory. Paths are expanded using `FS.realpath`
to follow symlinks.
```

Also, we could perhaps ditch `sanitizePaths` altogether and just expose `sanitizePath`, which allows us to drop the `filterable` dependency. In the application we can use the `partitionEithers` helper function on the result of traversing with `sanitizePath`. Then this comment would just apply to `sanitizePath` and we can test it in isolation.

lib/src/Registry/Internal/Path.purs
```diff
@@ -0,0 +1,41 @@
...
+sanitizePaths baseDirectory paths = do
+  sanitized <- traverse (sanitizePath baseDirectory) paths
+  let { right, left } = separate sanitized
+  pure { succeeded: right, failed: left }
+
+sanitizePath :: FilePath -> FilePath -> Aff (Either String FilePath)
+sanitizePath baseDirectory path = do
+  absoluteRoot <- Aff.attempt (FS.realpath baseDirectory) >>= case _ of
+    Left _ -> unsafeCrashWith $ "sanitizePath provided with a base directory that does not exist: " <> baseDirectory
```
thomashoneyman: I think we should replace this `unsafeCrashWith` with a normal `Left` if it's in library code. (Same with the one used below).

colinwahl: Actually, on second inspection,

`type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array FilePath }`

indicates the failure case is actually a `FilePath` - whereas now it'd be either a file path or a failure string. Let me see how we use the failure cases to see what to do about that.

colinwahl: Ok, it looks like all app code just inspects the `failed` array to check the amount of failed items is 0 - which would still be fine. I'll change this type to be `type SanitizedPaths = { succeeded :: Array FilePath, failed :: Array String }` just so it's more correct and continue with that.

PR #556, opened 2022-11-14
Remove 'RegistryJson' class

thomashoneyman:
This completes the migration to JSON codecs by removing the old `RegistryJson` class altogether and writing codecs and ad-hoc decoders to replace the use of `argonaut-codecs`. This should have zero change to the behavior of the application, and reviewing this is a largely mechanical effort (sorry!).

In this PR I've used two approaches:

1. For the most part I write a `JsonCodec` for whatever type is being encoded and/or decoded.
2. Occasionally I write a decode function `Json -> Either String a` if we only ever decode the value _and_ we're performing some kind of processing (like casing on the result of a previously-parsed field value to determine how to encode another field). For these I could still write the codec, but it would be a boilerplate-y for a one-off decoder.


app/src/Registry/Legacy/PackageSet.purs
```diff
@@ -78,6 +87,13 @@ type LegacyPackageSetEntry =
   , version :: RawVersion
   }
 
+legacyPackageSetEntryCodec :: JsonCodec LegacyPackageSetEntry
```
f-f, line 90: We decode legacy package sets in Spago - would it be worth moving this stuff to the library (e.g. to the `Internal` module), or shall we copy it over?

app/src/Registry/Legacy/PackageSet.purs
```diff
@@ -91,9 +107,14 @@ derive instance Newtype PscTag _
...
 derive instance Ord PscTag
 
-instance RegistryJson PscTag where
-  encode = Json.encode <<< printPscTag
-  decode = Json.decode >=> parsePscTag
+pscTagCodec :: JsonCodec PscTag
+pscTagCodec = CA.codec' decode encode
+  where
+  decode =
+    lmap (CA.Named "PscTag" <<< CA.TypeMismatch) <<< parsePscTag <=< CA.decode CA.string
```
f-f: This looks neat on a single line, but it's a bit dense. Could we use a do-block to make it easier to parse?

app/src/Registry/Legacy/PackageSet.purs
thomashoneyman: What do you think? We could move the legacy package set types to the library, or have Spago rely on the `app` directory. I don't really care about this one, though I'd not use the `RawVersion` newtypes if this moves into e.g. `Registry.Legacy.PackageSet` in the library.

thomashoneyman: Or Spago can reimplement it since it's small. Up to you!

thomashoneyman: (I have a _slight_ preference for having the legacy stuff in the `app` directory, so that we can remove it when we stop generating legacy package sets from the registry, but I'll defer if you want it to be in the `lib` package)

f-f: Oh, if the plan is to eventually remove this code then I'm fine with copying it over!

thomashoneyman: Yea, we won’t make legacy package sets forever, right? Once we stop generating them we have no need for the code.

f-f (approved):

PR #557, opened 2022-11-17
Migrate remaining modules into lib or app namespaces

thomashoneyman:
This completes the reorganization of the registry into a library and application. Specifically:

1. The `Solver` module and its tests (minus the integration test and `BowerInstaller`, which will eventually be removed) have been moved into the library. We can move this back out of the library if it turns out we need the solver to be in its own package.
2. All modules related to the application are under the `App` namespace.

With this, there is nothing else to move around (phew).

I made one other change in here: there is now a `Registry.Types` module that re-exports all the types from the registry schemas (e.g. everything except the `Operation` and `Solver` types). This cuts down on imports quite a bit, and lets us re-export that whole module from `Registry.Prelude` so we always have the registry types available. However, if y'all don't like this then I can revert that part of the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ PureScript needs a way to distribute packages. We used to rely on the Bower regi
 
 Here's a non-comprehensive list of desiderable properties/goals that we'd generally like to
 achieve when talking of a PureScript registry, and that we think this design covers:
-
 - **independent**: we're coming from a situation when relying on a third party registry has
   faulted us, so we'd need something that we can control and possibly host ourselves.
 - **immutable**: published packages are immutable - once a version has been published
@@ -16,7 +15,7 @@ achieve when talking of a PureScript registry, and that we think this design cov
   The only exception is that **unpublishing** will be possible for some time after publishing.
   This goal also directly stems from our experience with Bower, where we were not able to
   prevent packages from disappearing and/or being altered, since we were not storing the source
-  anywhere, but just pointing to the original location of the source, that we had no control over.
+  anywhere, but just pointing to the original location of the source, that we had no control over.  
   This means that with this Registry if your package is building today then you're guaranteed that
   the packages you're depending on will not disappear.
 - **with content hashes**, so that we're able to support independent storage backends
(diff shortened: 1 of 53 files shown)


README.md
```diff
@@ -8,14 +8,15 @@ PureScript needs a way to distribute packages. We used to rely on the Bower regi
 
 Here's a non-comprehensive list of desiderable properties/goals that we'd generally like to
 achieve when talking of a PureScript registry, and that we think this design covers:
+
```
thomashoneyman: These are formatting changes from `prettier`. I can revert if this is too noisy in this PR.

app/src/Registry/App/Auth.purs
```diff
@@ -1,10 +1,10 @@
-module Registry.SSH
+module Registry.App.Auth
```
thomashoneyman, line 1: I noticed that in the spec we always talk about authentication, and this module has to do with how we authenticate payloads. Having spent time working with these modules I think `Auth` is a more sensible module name. But I can go back to SSH if desired.

app/src/Registry/App/PackageIndex.purs
```diff
@@ -1,6 +1,6 @@
-module Registry.App.Index where
+module Registry.App.PackageIndex where
```
thomashoneyman, line 1: This is the file name (and I like it a little better).

app/test/Integration.purs
```diff
@@ -52,9 +44,9 @@ main = launchAff_ do
     Spec.describe "Solves node packages" do
       mkTest solverIndex $ unsafeFromJust $ Map.lookup "purescript-web" solutions
 
-type Owner = String
+type RepoOwner = String
```
thomashoneyman, line 47: This shadows the `Owner` import.

README.md
f-f: I'm ok with prettier formatting stuff, but we should ship this in another PR where we also add it to CI (otherwise we'll have everyone's own formatter shuffling things around all the time)

f-f (approved):

PR #558, opened 2022-11-19
Move GitHub event parsing into script

thomashoneyman:
The next iteration of the Registry is going to be an application serving an HTTP API. However, we'll still have the ability to submit at least package set operations via GitHub Issues. To prepare for this change I have moved the GitHub event trigger out to its own script; when we receive a GitHub event we can call the script, and the script can either process the event directly or call out to our running HTTP server (whichever we prefer).

I haven't touched the functionality. This is a reorganization only. However, I noticed that we're likely to pull in the `Test.Assert` module if we depend on say `httpure` or `httpurple`, and so I've gone ahead and renamed our `Test.Assert` to be `Test.Utils.Assert` to avoid conflicts.


thomashoneyman:
I'm going to close this because with #563 we can more easily have a single entrypoint application that lets you determine what environment you want to run in -- GitHub event, HTTP API, etc.

PR #559, opened 2022-11-21
Add git url parser to Location

thomashoneyman:
As per the SPEC discussion here:
https://github.com/purescript/registry-dev/pull/533#discussion_r1025962326

...this adds a parser that will reject Git urls if they don't follow the accepted format.


lib/src/Registry/Internal/Parsing.purs
```diff
@@ -3,26 +3,43 @@ module Registry.Internal.Parsing where
...
 import Parsing as Parsing
 import Parsing.Combinators as Parsing.Combinators
+import Parsing.Combinators.Array as Parsing.Combinatiors.Array
 import Parsing.Combinators.Array as Parsing.Combinators.Array
 import Parsing.String as Parsing.String
+import Parsing.String.Basic as Parsing.String.Basic
 
+-- | INTERNAL
+-- |
+-- | Parse as many BMP chars as possible
```
f-f: What's BMP? Could we expand the acronym?

thomashoneyman: It’s “Basic Multilingual Plane”. We can expand it.

MonoidMusician (approved):

PR #88, opened 2020-10-18
Add debuggest

ursi:
debug and debugger were taken

PR #560, opened 2022-11-21
More lenient JSON parser for issue comments

thomashoneyman:
We keep seeing terrible errors like this one when no decoder matches a proposed operation:
https://github.com/purescript/registry/issues/88#issuecomment-1321139667

...or seeing no response at all when a payload is not parseable as JSON due to things like trailing commas.

This PR tries to make the situation a bit better by:

1. Using the JsonRepair module to fix things like trailing commas instead of failing to parse the JSON altogether
2. Explicitly checking for object keys we know correspond with specific operations before we decode them, so that we don't fall back to the authenticated case automatically. It's a little hacky, but this really is just for the GitHub Issues workflow. When we have an HTTP API the decoder will match the route, so this is trying to get at the same concept.

### DIFF
--- a/app/src/Registry/App/API.purs
+++ b/app/src/Registry/App/API.purs
@@ -43,7 +43,9 @@ import Foreign.FastGlob as FastGlob
 import Foreign.Git as Git
 import Foreign.GitHub (GitHubToken(..), IssueNumber, PackageURL(..))
 import Foreign.GitHub as GitHub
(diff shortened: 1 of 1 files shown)


thomashoneyman:
I didn't add any tests because we already have tests for GitHub event payloads, by the way.

f-f (approved):

f-f:
Thanks!

PR #561, opened 2022-11-21
Migrate operation validation into library

colinwahl:
#523

This PR takes the operations checks located inline within `Registry.API` and moves them to a separate module within the registry library - it also adds tests for the more complex ones, although most are straightforward.

Note that there is some discussion in #523 about exposing utilities for logging progress of these checks - ultimately @thomashoneyman and I thought that it was adding complexity that was blocking getting these migrated out, so I opted to just move the checks over and can come back to look at progress reporting if it's still wanted in the future. Note that these checks are pretty light (except for possibly the one invoking the solver), so they'd start/complete very quickly even if we were displaying their progress.


lib/test/Utils.purs
```diff
@@ -48,27 +52,31 @@ partitionEithers = Array.foldMap case _ of
...
 
 -- | Unsafely parse a version from a string
 unsafeVersion :: String -> Version.Version
-unsafeVersion = fromRight "Failed to parse Version" <<< Version.parse
+unsafeVersion str = fromRight ("Failed to parse Version: " <> str) (Version.parse str)
 
 -- | Unsafely parse a range from a string
 unsafeRange :: String -> Range.Range
-unsafeRange = fromRight "Failed to parse Range" <<< Range.parse
+unsafeRange str = fromRight ("Failed to parse Range: " <> str) (Range.parse str)
```
colinwahl, line 71: I had an issue where I passed in an invalid range and couldn't figure out where it came from - I modified these functions to print the raw string input to help diagnose failures in the future.

thomashoneyman: Good idea. Thanks!

app/src/Registry/App/API.purs
```diff
@@ -370,27 +368,23 @@ runOperation source operation = case operation of
...
-        Just _, Just _ ->
+        Left Operation.Validation.InternalError ->
           throwWithComment $ String.joinWith "\n"
             [ "Cannot unpublish " <> Version.print version <> "."
             , ""
             , "This version is listed both as published and unpublished. This is an internal error."
             , "cc @purescript/packaging"
             ]
+        Left (Operation.Validation.PastTimeLimit { limit }) ->
+          throwWithComment $ "Packages can only be unpublished within " <> Number.Format.toString (unwrap limit) <> " hours."
```
thomashoneyman: Could we add the `difference` to the error message, noting how many hours have actually elapsed for the version?

app/src/Registry/App/API.purs
```diff
@@ -532,7 +520,7 @@ publish source { name, ref, compiler, resolutions } (Metadata inputMetadata) = d
   log $ "Package available in " <> packageDirectory
 
   log "Verifying that the package contains a `src` directory"
-  whenM (liftAff $ map (Array.null <<< _.succeeded) $ FastGlob.match packageDirectory [ "src/**/*.purs" ]) do
+  whenM (liftAff $ Operation.Validation.containsPursFile (Path.concat [ packageDirectory, "src" ])) do
     throwWithComment "This package has no .purs files in the src directory. All package sources must be in the src directory."
```
thomashoneyman: This error message isn't quite right due to the `files` key. Could we rewrite it to be:

```purs
"This package has no .purs files in the src directory. All package sources must be in the `src` directory, with any additional sources indicated by the `files` key in your manifest.
```

app/src/Registry/App/API.purs
```diff
@@ -617,10 +605,10 @@ publish source { name, ref, compiler, resolutions } (Metadata inputMetadata) = d
...
   log "Checking the tarball size..."
   FS.Stats.Stats { size: bytes } <- liftAff $ FS.Aff.stat tarballPath
-  when (not isLegacyImport && bytes > warnPackageBytes) do
-    if bytes > maxPackageBytes then
+  for_ (Operation.Validation.validateTarballSize bytes) case _ of
+    Operation.Validation.ExceedsMaximum maxPackageBytes ->
       throwWithComment $ "Package tarball is " <> show bytes <> " bytes, which exceeds the maximum size of " <> show maxPackageBytes <> " bytes.\ncc: @purescript/packaging"
-    else
+    Operation.Validation.WarnPackageSize _ ->
       log $ "WARNING: Package tarball is " <> show bytes <> ".\ncc: @purescript/packaging"
```
thomashoneyman: Could we add the `, which exceeds the warning threshold of " <> show maxWarnBytes <> ".` to this error, and change the log to a `comment` so that the cc: on the purescript packaging team actually notifies us?

lib/src/Registry/Operation/Validation.purs
```diff
@@ -0,0 +1,165 @@
...
+  else do
+    results <- traverse (_.path >>> containsPursFile) directories
+    pure $ Array.any (eq true) results
+
+-- | Checks that the manifest package name and the PublishData payload package name match.
+nameMatches :: Manifest -> PublishData -> Boolean
+nameMatches (Manifest manifestFields) { name } =
+  manifestFields.name == name
+
+-- | Checks that the manifest location and JSON payload location match.
```
thomashoneyman: This looks like it's checking the manifest and the metadata, not the manifest and the payload. Although all three are expected to match so long as this is a `Publish` or `Unpublish` (and not expected to match if this is a `Transfer`).

lib/test/Registry.purs
```diff
@@ -34,6 +35,8 @@ main = Aff.launchAff_ $ Spec.Runner.runSpec [ Spec.Reporter.consoleReporter ] do
...
     Spec.describe "Operation" Test.Operation.spec
 
-  Spec.describe "Structures" do
-    Spec.describe "ManifestIndex" Test.ManifestIndex.spec
-    Spec.describe "Solver" Test.Solver.spec
+  Spec.describe "ManifestIndex" Test.ManifestIndex.spec
+  Spec.describe "Solver" Test.Solver.spec
+
+  Spec.describe "Operation Validation" do
+    Test.Operation.Validation.spec
```
thomashoneyman:
```suggestion
  Spec.describe "ManifestIndex" Test.ManifestIndex.spec
  Spec.describe "Solver" Test.Solver.spec
  Spec.describe "Operation Validation" Test.Operation.Validation.spec
```

thomashoneyman:
@f-f Curious to hear what you think, as you'll be using this in Spago.

lib/src/Registry/Operation/Validation.purs
```diff
@@ -0,0 +1,165 @@
...
+  where
+  -- | The absolute maximum bytes allowed in a package
+  maxPackageBytes :: Number
+  maxPackageBytes = 2_000_000.0
+
+  -- | The number of bytes over which we flag a package for review
+  warnPackageBytes :: Number
+  warnPackageBytes = 200_000.0
+
+locationIsUnique :: Location -> Map PackageName Metadata -> Boolean
```
f-f, line 129: This one should only be for legacy packages, so it can stay in the App.

f-f (approved):

f-f:
This looks good! At first read I'd say it's fine for Spago, but we'll only know once we start using it there. Right now spago is on an ancient version of the registry lib though, and I'd like to wait until things have settled (e.g. with the effects migration) before pulling the new version (which will be a source of unsettling by itself already, as things might need refactoring then)

thomashoneyman (approved):

PR #562, opened 2022-11-27
New Solver

MonoidMusician:


lib/src/Registry/Version.purs
```diff
@@ -44,11 +43,8 @@ newtype Version = Version
   , patch :: Int
   }
 
-instance Eq Version where
-  eq = eq `on` (\(Version v) -> [ v.major, v.minor, v.patch ])
-
-instance Ord Version where
-  compare = comparing (\(Version v) -> [ v.major, v.minor, v.patch ])
+derive instance Eq Version
+derive instance Ord Version
```
thomashoneyman: Does this exactly match the existing compare in that it compares by major, then minor, then patch?

lib/test/Registry/Solver.purs
```diff
@@ -2,32 +2,60 @@ module Test.Registry.Solver (spec) where
...
+            then Assert.fail $ "Error(s): " <> intercalateMap "\n" printSolverError solverErrors
+            else Assert.fail $ "Tests expect " <> show expectedErrorCount <> " errors, but received " <> show receivedErrorCount
+
+        let receivedErrors = map (\error -> { error, message: printSolverError error }) solverErrors
+        let combinedErrors = Array.zip errors (Array.fromFoldable receivedErrors)
+
+        for_ combinedErrors \(Tuple expected received) -> do
+          received.message `Assert.shouldEqual` expected.message
+          when (received.error /= expected.error) do
+            log $ unsafeCoerce received.error
```
thomashoneyman: What’s the need for the unsafe coerce here? Could you use unsafeStringify?

app/test/Integration.purs
```diff
@@ -1,147 +0,0 @@
-module Test.Integration (main) where
```
thomashoneyman, line 1: If this module is being removed entirely then I think we can remove the BowerInstaller script too (since it’s only used to produce the solutions for the integration test)

lib/src/Registry/Version.purs
MonoidMusician: Funny you should ask :D It actually was correct (alphabet order is major minor patch), but I'll push a clearer hand-optimized version.

lib/src/Registry/Solver.purs
```diff
@@ -1,336 +1,675 @@
...
+  ]
+
+
+--------------------------------------------------------------------------------
+-- Core algorithm
+--------------------------------------------------------------------------------
+
+type TransitivizedRegistry =
+  SemigroupMap PackageName (SemigroupMap Version (SemigroupMap PackageName Intersection))
+type RR r =
```
thomashoneyman, line 240: I think this is going to run afoul of the formatter

lib/src/Registry/Version.purs
thomashoneyman: Thanks!

lib/test/Registry/Solver.purs
MonoidMusician: Sooo funny story about the unsafeStringify sitting in the test modules: by printing it as json it doesn't distinguish constructor names. So I did have to add that for debugging that issue (in particular it was `Root` vs `Trial`), we can remove it now and consider if there's something better to do later.

lib/src/Registry/Solver.purs
MonoidMusician: I, uh, prefer boutique, artisanal, hand-crafted formatting. ;)

lib/test/Registry/Solver.purs
thomashoneyman: Ah! Ok. Well, perhaps we could add a show class of our own that’s only in the tests? I just don’t want the instances propagating outward and that way we can control it.

lib/src/Registry/Version.purs
```diff
@@ -43,8 +43,23 @@ newtype Version = Version
...
-derive instance Eq Version
-derive instance Ord Version
+instance Eq Version where
+  eq (Version l) (Version r) =
+    case l.major == r.major of
+      true ->
+        case l.minor == r.minor of
+          true -> l.patch == r.patch
+          x -> x
+      x -> x
```
thomashoneyman, line 53: Does this meaningfully change the JS output? I’m curious what the effect is. We should probably leverage purescript-optimizer if performance is becoming a concern, too.

MonoidMusician: My `Ord` instance saves some case splits since it only matches on `EQ` instead of `LT` and `GT`, the `Eq` instance is probably about the same after JITing? idk:
https://gist.github.com/MonoidMusician/a287fb40bbf832edc6c571a9bc4f0f26

But idk, I didn't test the performance of this specifically, or at least not in a while (I think I was originally seeing ~10% speedup, which is in the realm of whether my computer is doing other things or not at the time I run the tests), but it is the most run part of the code.

MonoidMusician: It's certainly quicker than the `on` version, which allocated a lot more with intermediate functions, arrays.

lib/src/Registry/Solver.purs
```diff
@@ -1,336 +1,718 @@
...
+--------------------------------------------------------------------------------
+-- Public API
+--------------------------------------------------------------------------------
+
+-- | Data from the registry index, listing dependencies for each version of
+-- | each package
+type DependencyIndex = Map PackageName (Map Version (Map PackageName Range))
+
+-- | Solve a map of requirements given a registry index.
+solve :: DependencyIndex -> Map PackageName Range -> Either SolverErrors (Map PackageName Version)
```
f-f, line 50: Would it be a big deal for the algorithm to change this `DependencyIndex` to be a `PackageName -> Effect (Map Version (Map PackageName Range))`?
(possibly wrapping the result in an `Either String` in case of errors)

We won't be able to use this in Spago if it requires a whole index to be loaded in memory.

MonoidMusician: Yes, I started that locally and will complete it when I have a chance. It'll just be polymorphic over `Monad m`.

f-f: Great! I thought about just parameterising over any monad, but I suggested `Effect` because it would have zero-cost over the current implementation, while the parametrised Monad might.

MonoidMusician:
My short-term goal is to introduce a public `SolverError` type and a private one. The private type will be what exists currently. But the public type will be meant for consuming by custom printing functions (e.g. colorful output). And then I can gradually introduce better error reporting by changing the private->public mapping without affecting consumers. (E.g. to aggregate errors across failing cases.)

lib/src/Registry/Version.purs
thomashoneyman: How were you benchmarking this? Is there a repo somewhere? Should we add benchmarks to the tests?

lib/test/Registry/Solver.purs
```diff
@@ -144,30 +184,82 @@ spec = do
...
+              [ Tuple (version 1) $ Conflicts $ Map.fromFoldable
+                  [ package "prelude" `Tuple` intersection 2 (solveRoot "prelude") 1 (via "chaotic" 1 [])
+                  ]
+              , Tuple (version 2) $ Conflicts $ Map.fromFoldable
+                  -- TODO: why no global "chaotic" here?
+                  [ package "prelude" `Tuple` intersection 5 (via "chaotic" 2 []) 4 (solveRoot "prelude")
+                  ]
+              , Tuple (version 3) $ Conflicts $ Map.fromFoldable
+                  -- TODO: why no global "chaotic" here?
+                  [ package "prelude" `Tuple` intersection 5 (via "chaotic" 3 []) 4 (solveRoot "prelude")
```
thomashoneyman, line 240: How important is this TODO question? What are you expecting?

lib/test/Registry/Solver.purs
```diff
@@ -258,60 +353,61 @@ prelude =
...
+  { package: package "qaotic"
+  , versions: Map.fromFoldable
+      [ version 1 /\ Just (prelude.package /\ range 0 1)
+      , version 2 /\ Just (prelude.package /\ range 5 6)
       ]
   }
 
 package :: String -> PackageName
-package = Utils.unsafePackageName
+package = fromRight "bad package name" <<< PackageName.parse
```
thomashoneyman, line 383: unsafePackageName ain’t good enough? 😅

thomashoneyman:
I have a couple questions I’ve added as comments, but on the whole this looks good to me — a strict improvement. I’m happy to get it in and keep iterating from there.

lib/src/Registry/Solver.purs
```diff
@@ -1,336 +1,775 @@
...
+  :: forall m
+   . Monad m
+  => Loader m
+  -> Map PackageName Range
+  -> m DependencyIndex
+loadIndex loader required = map _.known <$> go Map.empty (need required)
   where
-  toBeFound :: Set (Tuple PackageName Version)
-  toBeFound = errored # foldMapWithIndex (Set.map <<< Tuple)
+  need = SemigroupMap <<< map pure
```
thomashoneyman, line 77: Do we still have warnings in this module due to use of semigroup instances for Map? Or does theLos avoid it?

thomashoneyman:
I know you had shared some general performance notes in Discord — would you mind updating everyone about the current distribution of results? Time to solve a typical package, time to solve a pathological worst-case?

lib/test/Registry/Solver.purs
MonoidMusician: It's inconsequential – it's a sign that I don't fully understand how the locations are propagated, but it doesn't actually affect what the user sees in error messages or their understanding of what's going on.

lib/test/Registry/Solver.purs
MonoidMusician: whoops, quick hide the evidence I just clobbered all of your work in my merge 😅

lib/src/Registry/Solver.purs
MonoidMusician: I finally fixed those :) but we will have warnings when we hit the next round of the `SemigroupMap` deprecation cycle, when we'll want to remove that and use the normal `Map` again, which can't come soon enough

lib/src/Registry/Version.purs
MonoidMusician: The benchmarking was pretty hacky, inline functions to profile each part. You can see the helpers here and their usage scattered throughout:
https://github.com/purescript/registry-dev/blob/05df62fe0871836de3db31dfcefec3366fc605f9/lib/src/Registry/Solver.purs#L546

We can definitely add proper (external) benchmarks for the `solve` function as a whole, now that I don't need to dissect the various parts of it to see what's wrong, but I'd rather focus on better errors.

MonoidMusician:
Took about an hour to solve the whole registry, package by package. Out of 11519 package versions, 1848 (16%) took more than half a second, only 189 (1.6%) took more than 1 second, and 4 took 10-20 seconds. Those 4 that took the most time are:
```
spec-mocha@4.0.0
justifill@0.2.0
milkis@7.5.0
toppokki@2.5.0
```

Would be interesting to see what makes these particularly slow at some point, but it's not critical.

Using this script in my old unmerged branch: https://github.com/purescript/registry-dev/blob/mm/best-errors/lib/test/Registry/SolveAll.purs

EDIT: if someone wants to crunch the data: https://gist.github.com/MonoidMusician/0d7548c6bd6a606110b3c4c7d9c765f8

lib/src/Registry/Solver.purs
thomashoneyman: Totally agree, those warnings kill me

thomashoneyman (approved):

f-f (approved):

Issue #563, opened 2022-11-30
Implement structured effects in the dev codebase


thomashoneyman:
I plan to rework how we handle effects (like logging, making requests, using a cache, etc.) in the application.

### Background

In the last registry call we discussed some difficulties I've been having when implementing the new HTTP API. Specifically, the current codebase has a few patterns that are difficult to work with:

* The `Foreign` modules are meant to be low-level bindings, but instead we've got quite a lot of registry-specific code that's trickled into it, like the GitHub module relying on our cache effect and making specific authenticated requests to the registry rep, or the S3 FFI bindings looking for specific environment variables to be available. These modules shouldn't have registry-specific or environment-specific code if we can help it, because it dramatically expands the footprint you have to be aware of to make changes to things like GitHub requests.
* Most code is written `Aff` directly and can perform arbitrary effects, like throwing exceptions, writing to the file system, making requests, reading the Node process, logging, committing to GitHub, using the `git`, `purs`, and other command-line tools, and more.
* Code that is in `RegistryM` is barely better than unbounded `Aff`: most of the code is still in `Aff`, with `RegistryM` simply providing access to a tiny handful of functions (`throwWithComment`, etc.) and a minimal environment. All non-trivial code is still implemented in `Aff`, which makes plenty of functions essentially un-testable in a pure environment.
* The `RegistryM` monad itself is for all intents and purposes hardcoded to the GitHub environment -- we've got things like `Octokit` in the environment, functions like `throwWithComment` which assume access to an issue number to write a comment on, and so on. There is no abstraction over the now _many_ environments we execute in -- test fixtures, local dry runs, local runs that commit, GitHub containers triggered by comments, GitHub containers triggered by scheduled jobs, and a HTTP API running on the server.
* Abstracting effects makes it possible for Spago to use more registry functions directly and provide its own methods for things like reading manifests, logging, downloading packages, and so on.
* Implementing `RegistryM` as a record of handlers is problematic because the handlers cannot be mixed and matched or easily rely on each other -- for example, it's quickly becomes messy to implement functions to make GitHub requests that rely on a logging effect that can be handled separately. This is really important for things like a shared logging effect.

[...]

PR #564, opened 2022-12-04
Add base Log and Notify effects

thomashoneyman:
This takes the first step of #563 by abstracting two effects: `MonadLog` for logging information and `MonadNotify` for notifying subscribers of events. I took a little inspiration from Spago with the logging so that they're similar. I separated out `notify` from the logging because they're really separate behaviors. The first helps us diagnose issues and should be informative for registry developers; the second should be helpful and well-written to help registry users understand what has happened in the process. I hope the separation can help us be more conscientious when showing messages to users.

Both effects are fairly bare-bones right now but they demonstrate how to write more of these effects, how to interpret one effect in terms of another, how to write concrete monads that use the various interpreters, and so on. Eventually I am envisioning that we have a `ServerM` for the HTTP API with webhooks for notifying users; a `GitHubM` for events handled from a GitHub event with comments for notifying users; a `LocalM` for running these tools locally when working on registry resources, in which there is no need for `notify`; and a `TestM` for running tests, in which we might make sure particular functions notify what we'd expect.

In each case the monad can be built by mixing & matching the handlers.


app/src/Foreign/Git.purs
```diff
@@ -46,7 +45,7 @@ cloneGitTag url ref targetDir = do
   withBackoff' (Except.runExceptT (runGit args (Just targetDir))) >>= case _ of
     Nothing -> Aff.throwError $ Aff.error $ "Timed out attempting to clone git tag: " <> url <> " " <> ref
     Just (Left err) -> Aff.throwError $ Aff.error err
-    Just (Right _) -> log "Successfully cloned package."
+    Just (Right _) -> Console.log "Successfully cloned package."
```
thomashoneyman, line 48: The majority of changes in this PR are:

* imports of `Effect.Class.Console` and renames of `log` to `Console.log`, since it's not exported by the prelude anymore
* migrations of `log` to `Log.debug`, `Log.warn`, etc. as appropriate, since we now have a logging effect
* migrations of `throwWithComment` to `Notify.die` and `comment` to `Notify.notify`

Those changes can just be glanced at, since they're essentially just renames.

app/src/Foreign/GitHub.purs
```diff
@@ -256,7 +257,7 @@ getRateLimit octokit = do
 printRateLimit :: RateLimit -> Aff Unit
 printRateLimit { limit, remaining, resetTime } = do
   offset <- liftEffect Now.getTimezoneOffset
-  log $ Array.fold
+  Console.log $ Array.fold
```
thomashoneyman, line 260: Everything that's hardcoded to `Aff` has to continue using `Console.log`, but we'll eventually move this all to effects.

app/src/Registry/App/API.purs
```diff
@@ -742,14 +745,19 @@ verifyManifest { metadata, manifest } = do
...
+    Notify.die $ "Some dependencies of your package were not found in the Registry: " <> String.joinWith ", " (map PackageName.print pkgsNotInRegistry)
 
 -- | Verify the build plan for the package. If the user provided a build plan,
 -- | we ensure that the provided versions are within the ranges listed in the
 -- | manifest. If not, we solve their manifest to produce a build plan.
-verifyResolutions :: { source :: Source, resolutions :: Maybe (Map PackageName Version), manifest :: Manifest } -> RegistryM (Either String (Map PackageName Version))
+verifyResolutions
+  :: forall m r
+   . MonadRegistry m
+  => MonadAsk (GitHubEnv r) m
```
thomashoneyman, line 756: This is really ugly, and it only exists because we still have a lot of handlers in the `Env` until they are migrated to actual effects. This will eventually go away and we can just have the `MonadRegistry` constraint.

app/src/Registry/App/Monad.purs
```diff
@@ -0,0 +1,168 @@
...
+  liftAff $ f pkg path
+
+-- | Delete a package from the backend storage provider
+handleDeletePackage :: forall r m. MonadAsk { | HandlersEnv r } m => MonadAff m => PackageInfo -> m Unit
+handleDeletePackage pkg = do
+  f <- asks _.deletePackage
+  liftAff $ f pkg
+
+-- TODO: right now we write this to file separately, but maybe it'd be better
+-- to do everything here so we don't risk to forget this?
```
thomashoneyman, line 172: This is an old TODO. We should have an `Index` effect that handles writing and reading metadata and hides away the implementation detail of the ref in memory backed by the Git repository.

app/src/Registry/App/PackageIndex.purs
```diff
@@ -11,15 +11,17 @@ import Data.String as String
...
+import Registry.Effect.Notify as Notify
 import Registry.ManifestIndex as ManifestIndex
 import Registry.PackageName as PackageName
 import Registry.Range as Range
 import Registry.Version as Version
 
 -- | Attempt to read a manifest index from disk, throwing an exception if the
 -- | index is not valid or cannot be read.
-readManifestIndexFromDisk :: RegistryM ManifestIndex
+readManifestIndexFromDisk :: forall m r. MonadRegistry m => MonadAsk (GitHubEnv r) m => m ManifestIndex
```
thomashoneyman, line 24: Again, the horrible constraints are temporary. In the future this can be:

```purs
readManifestFromDisk :: forall m. MonadRegistry m => m ManifestIndex
```

or, since we're only using notifications and the file system we could only request those effects:

```purs
readManifestIndexFromDisk :: forall m. MonadNotify m => MonadAff m => m ManifestIndex
```

depending on how we want to handle it.

app/src/Registry/App/API.purs
```diff
@@ -885,14 +893,14 @@ compilePackage { packageSourceDir, compiler, resolutions } = do
       filename = PackageName.print packageName <> "-" <> Version.print version <> ".tar.gz"
       filepath = Path.concat [ dir, filename ]
 
-    liftAff (withBackoff' (Wget.wget (Constants.packageStorageUrl <> "/" <> PackageName.print packageName <> "/" <> Version.print version <> ".tar.gz") filepath)) >>= case _ of
-      Nothing -> throwWithComment "Could not fetch tarball."
-      Just (Left err) -> throwWithComment $ "Error while fetching tarball: " <> err
+    liftAff (Wget.wget (Constants.packageStorageUrl <> "/" <> PackageName.print packageName <> "/" <> Version.print version <> ".tar.gz") filepath) >>= case _ of
```
thomashoneyman, line 896: I removed `withBackoff'` here because `wget` already uses it, so it's nested.

app/test/TestM.purs
```diff
@@ -0,0 +1,70 @@
...
+  notify _ = pure unit
+
+instance MonadRegistry TestM where
+  commitMetadataFile = App.handleCommitMetadataFile
+  commitIndexFile = App.handleCommitIndexFile
+  commitPackageSetFile = App.handleCommitPackageSetFile
+  uploadPackage = App.handleUploadPackage
+  deletePackage = App.handleDeletePackage
+  updatePackagesMetadata = App.handleUpdatePackagesMetadata
+  readPackagesMetadata = App.handleReadPackagesMetadata
```
thomashoneyman, line 49: These don't do anything because they just pull from the test env below. When those move into an effect then this duplication can be removed.

app/src/Registry/App/API.purs
```diff
@@ -1139,25 +1144,31 @@ pursPublishMethod :: PursPublishMethod
...
-      LegacyPursPublish -> liftAff do
-        log $ "Cloning repo at tag: " <> show { owner, repo, ref }
-        Git.cloneGitTag (Array.fold [ "https://github.com/", owner, "/", repo ]) ref tmpDir
-        log $ "Getting published time..."
+      LegacyPursPublish -> do
+        Log.debug $ "Cloning repo at tag: " <> show { owner, repo, ref }
+        let address = Array.fold [ "https://github.com/", owner, "/", repo ]
+        liftAff (try (Git.cloneGitTag address ref tmpDir)) >>= case _ of
+          Left error -> Notify.die $ "Could not clone " <> address <> " at ref " <> ref <> " due to a git error: " <> Aff.message error
+          Right _ -> pure unit
```
thomashoneyman: This fixes the issue with the registry silently failing when you try to publish a package with a tag that doesn't exist, by the way (see https://github.com/purescript/registry/issues/91)

PR #565, opened 2022-12-12
Add script to delete packages

thomashoneyman:
As described in #531, we have a lot of packages in the registry that cannot be solved. The full list is here:
https://gist.github.com/MonoidMusician/0f0c4ed70edf7c000a1da012bdca5134

We plan to remove unsolvable packages and update the API so that packages must solve to be added to the registry. However, since there are quite a lot of them, we need a script to help us do this automatically. This PR adds such a script as a CLI tool with two options.

First, you can delete an individual package:

```console
$ registry-package-deleter --package fork@1.0.0
```

Second, you can write a JSON file where keys are package names and values are arrays of versions:

```json
{
  "fork": [ "1.0.0" ],
  "aff-bus": [ "0.0.1", "0.0.2" ]
}
```

and then delete them in bulk:

```console
$ registry-package-deleter --file deletions.json
```

You must have @pacchettibotti credentials in order to run this script (not just anyone can run it).


scripts/src/PackageDeleter.purs
```diff
@@ -0,0 +1,178 @@
...
+          Right _ ->
+            Console.log $ "Successfully removed " <> printed
+
+      commitMetadataFile name >>= case _ of
+        Left err -> Console.log $ "Failed to commit metadata file for " <> PackageName.print name <> ": " <> err
+        Right _ -> pure unit
+
+      commitIndexFile name >>= case _ of
+        Left err -> Console.log $ "Failed to commit registry index file for " <> PackageName.print name <> ": " <> err
+        Right _ -> pure unit
```
thomashoneyman: If we don't want a commit per package and just want to commit in one fell swoop then we'd need to remove these lines and instead add functions to commit the entire registry repo / commit the entire registry index repo (or to `git add` all the filepaths we know we touched and commit them in one go).

MonoidMusician: hm okay, I think one fell swoop is ideal because the deletions don't necessarily make sense in isolation, since deleting package versions could break other package versions that depend on them (but we should also be deleting those in the same batch).

MonoidMusician (approved):

f-f:
Do we even need a script for this in the repo?

If we go by the plan detailed in https://github.com/purescript/registry-dev/issues/531#issuecomment-1342536699, then we should forbid the upload of packages that do not solve before removing all the old packages that do not (otherwise this issue will continue to happen), and this operation is a one-off. Having easy automation available for removing packages makes me uneasy, and I wouldn't like to have it sitting there in the first place.

thomashoneyman:
I can alternately remove this script from the repo and keep it locally; I’ve had to delete and reupload packages several times and it’s super tedious. 

Coming up we at least have removing packages that don’t solve and removing / reuploading packages with adjusted bounds once #567 lands.

The alternative is manually deleting entries from the metadata and index and going to the spaces UI to delete the relevant package version.

As far as it being dangerous to have this script: I agree, and certainly once we leave alpha it should be deleted, but the guardrail is that only people with access to the spaces secrets and pacchettibotti credentials can use it.

f-f:
Yeah. We can leave it in for now, but we'll need to remember to remove it once we are out of alpha. Opened #570 to track this.

PR #567, opened 2022-12-13
Fix ordering to prefer latest package set for legacy manifests

MonoidMusician:
When determining bounds for a legacy package set import, we want to look at the latest package set that included it to get the most recent bounds that would work for it. This is added to the bounds from the packages.dhall and/or bower.json in the published repo.

I verified that this fixes the logic by diffing the resulting values of `legacySets`, and it by and large increased the versions selected. (One version dependency got lower, but it doesn't affect the final manifest, and a few dependencies got added/removed but @thomashoneyman and I agreed that it seems to be correct.)

EDIT: updated the following, since I misread my statistics:
EDIT2: stats in node is a bad idea:

This is the upper bound of package versions that may be affected, there are **1315** of them across **535** packages, but some may still produce the same manifest after merging with the Spago and Bower information (to be determined):
https://gist.github.com/MonoidMusician/0c941247aec7c7b0e007180091709d97


MonoidMusician:
Forgot to say: this should mean that we have fewer packages that fail to solve, but I will have to rerun that script after we push through the changes from this.

thomashoneyman:
Thanks for catching this! We should merge this, delete the affected packages (see #565), reupload them, and then we can re-run the solver over the result. cc: @f-f

thomashoneyman (dismissed):

MonoidMusician:
Update: I was totally misreading my stats, the small number I reported was just of the packages that had dependencies added or removed. There are a lot more actually affected, see updated first comment.

f-f:
Do we have stats on how many more packages are solvable now?

MonoidMusician:
My overnight run of the legacy importer failed, so I need to rerun that and then run the solver before I can see if we made progress on that end.

MonoidMusician:
1097 -> 678 package versions still fail to solve, so it's down by 40% but still significant:
https://gist.github.com/MonoidMusician/0f0c4ed70edf7c000a1da012bdca5134/revisions#diff-b95accf405a6d2528db7e8dbee6bf22fbdf4f41b430708ebd1d5139a283127b0

EDIT: looks like there are some packages that newly fail to solve after the change ...
EDIT2: 1097 -> 678 -> 821

thomashoneyman:
As per the registry call this morning, we're going to adjust this merge to preserve the lowest-seen and highest-seen dependency versions for the specified package version across all package sets it was in. So if aff@1.0.0 was seen in 10 package sets with the dependency const, we'll take the lowest-seen and highest-seen const, such as const@1.0.0 -> const@2.2.1.

Then, we'll re-check how this affects the package solving.

MonoidMusician:
Okay, I FINALLY got a clean run with the new changes. (The importer wasn't working so I made the deleter script into a reimporter, found a codec bug in the process – after running it for an hour I mean.)

Current result is 981 package versions fail to solve. Barely a dent in the 1097 we started with. It is however a strict subset besides one package: react-basic-native@0.1.3.
Same gist: https://gist.github.com/MonoidMusician/0f0c4ed70edf7c000a1da012bdca5134

In light of the difficulty I've had running the importer, I'm more inclined to believe this latest result and disregard my earlier stuff. I'm not sure if I have enough scrollback in my terminal to play historian with what happened. I will have to spend more time investigating though, see if these results look reasonable.

MonoidMusician:
I'm leaning towards transitivizing dependencies before we merge the manifests. Otherwise the merging just doesn't make sense, I fear.

E.g. concur-react@0.5.0 (the latest, which appears in the current package set but no legacy package sets) has an outdated bower.json which lists more dependencies than spago.dhall, and so it doesn't actually solve with the current package set, even though its packages.dhall has a 0.15.4 package set:
https://github.com/purescript-concur/purescript-concur-react/blob/v0.5.0/bower.json
https://github.com/purescript-concur/purescript-concur-react/blob/v0.5.0/spago.dhall
https://github.com/purescript-concur/purescript-concur-react/blob/v0.5.0/packages.dhall

It solves on its own, but the resolved versions are outdated since they are dragged down by the bower deps, compared with what's in the latest package set:
<details>

```json
{
  "aff": [
    "5.1.2",
    "7.1.0"
  ],
  "arrays": [
    "5.3.1",
    "7.1.0"
  ],
  "avar": [
    "3.0.0",
    "5.0.0"
  ],
  "bifunctors": [
    "4.0.0",
    "6.0.0"
  ],
  "catenable-lists": [
    "5.0.1",
    "7.0.0"
  ],
  "concur-core": [
    "0.4.2",
    "0.5.0"
  ],
  "console": [
    "4.4.0",
    "6.0.0"
  ],
  "const": [
    "4.1.0",
    "6.0.0"
  ],
  "contravariant": [
    "4.0.1",
    "6.0.0"
  ],
  "control": [
    "4.2.0",
    "6.0.0"
  ],
  "datetime": [
    "4.1.1",
    "6.1.0"
  ],
  "distributive": [
    "4.0.0",
    "6.0.0"
  ],
  "effect": [
    "2.0.1",
    "4.0.0"
  ],
  "either": [
    "4.1.1",
    "6.1.0"
  ],
  "enums": [
    "4.0.1",
    "6.0.1"
  ],
  "exceptions": [
    "4.0.0",
    "6.0.0"
  ],
  "exists": [
    "4.0.0",
    "6.0.0"
  ],
  "foldable-traversable": [
    "4.1.1",
    "6.0.0"
  ],
  "foreign": [
    "5.0.0",
    "7.0.0"
  ],
  "foreign-object": [
    "2.0.3",
    "4.1.0"
  ],
  "free": [
    "5.2.0",
    "7.0.0"
  ],
  "functions": [
    "4.0.0",
    "6.0.0"
  ],
  "functors": [
    "3.1.1",
    "5.0.0"
  ],
  "gen": [
    "2.1.1",
    "4.0.0"
  ],
  "globals": [
    "5.0.0",
    null
  ],
  "identity": [
    "4.1.0",
    "6.0.0"
  ],
  "integers": [
    "4.0.0",
    "6.0.0"
  ],
  "invariant": [
    "4.1.0",
    "6.0.0"
  ],
  "js-date": [
    "6.0.0",
    "8.0.0"
  ],
  "lazy": [
    "4.0.0",
    "6.0.0"
  ],
  "lists": [
    "5.4.1",
    "7.0.0"
  ],
  "math": [
    "2.1.1",
    null
  ],
  "maybe": [
    "4.0.1",
    "6.0.0"
  ],
  "media-types": [
    "4.0.1",
    "6.0.0"
  ],
  "newtype": [
    "3.0.0",
    "5.0.0"
  ],
  "nonempty": [
    "5.0.0",
    "7.0.0"
  ],
  "now": [
    "4.0.0",
    "6.0.0"
  ],
  "nullable": [
    "4.1.1",
    "6.0.0"
  ],
  "ordered-collections": [
    "1.6.1",
    "3.0.0"
  ],
  "orders": [
    "4.0.0",
    "6.0.0"
  ],
  "parallel": [
    "4.0.0",
    "6.0.0"
  ],
  "partial": [
    "2.0.1",
    "4.0.0"
  ],
  "prelude": [
    "4.1.1",
    "6.0.1"
  ],
  "profunctor": [
    "4.1.0",
    "6.0.0"
  ],
  "profunctor-lenses": [
    "6.3.0",
    "8.0.0"
  ],
  "proxy": [
    "3.0.2",
    null
  ],
  "react": [
    "8.0.0",
    "10.0.1"
  ],
  "react-dom": [
    "6.1.0",
    "8.0.0"
  ],
  "record": [
    "2.0.2",
    "4.0.0"
  ],
  "refs": [
    "4.1.0",
    "6.0.0"
  ],
  "st": [
    "4.1.1",
    "6.2.0"
  ],
  "strings": [
    "4.0.2",
    "6.0.1"
  ],
  "tailrec": [
    "4.1.1",
    "6.1.0"
  ],
  "transformers": [
    "4.2.0",
    "6.0.0"
  ],
  "tuples": [
    "5.1.0",
    "7.0.0"
  ],
  "type-equality": [
    "3.0.0",
    "4.0.1"
  ],
  "typelevel-prelude": [
    "5.0.2",
    "7.0.0"
  ],
  "unfoldable": [
    "4.1.0",
    "6.0.0"
  ],
  "unsafe-coerce": [
    "4.0.0",
    "6.0.0"
  ],
  "web-dom": [
    "4.1.0",
    "6.0.0"
  ],
  "web-events": [
    "2.0.1",
    "4.0.0"
  ],
  "web-file": [
    "2.3.0",
    "4.0.0"
  ],
  "web-html": [
    "2.3.0",
    "4.1.0"
  ],
  "web-storage": [
    "3.0.0",
    "5.0.0"
  ]
}
```

</details>

Actually, transitivizing dependencies won't work well here anyways, since as shown above there are packages that don't exist anymore: globals, math, and proxy.

Besides transitivizing dependencies before merging, other options would be:
(1) look at git to see which of bower.json or spago/packages.dhall was more recently touched
(2) always prefer spago manifests

MonoidMusician:
Okay, I reran the earlier test (take bounds from the latest package set) with the new importer logic and it looks much more like the current result I reported (taking bounds from all package sets), with strictly more failures (1026).

Compared to our current baseline (bounds from oldest package set) it has some different failures, like halogen-formless@4.0.0 does solve but halogen-formless@3.0.0 does not. It makes sense that it would favor later versions.

MonoidMusician:
The set of packages in the current package set that do not solve with the versions from the set is smaller than I feared:
```
Incongruent: barbies@1.0.1
  Version out of range: console@6.0.0 >=5.0.0 <6.0.0
  Version out of range: const@6.0.0 >=5.0.0 <6.0.0
  Version out of range: control@6.0.0 >=5.0.0 <6.0.0
  Version out of range: effect@4.0.0 >=3.0.0 <4.0.0
  Version out of range: prelude@6.0.1 >=5.0.1 <6.0.0
  Version out of range: psci-support@6.0.0 >=5.0.0 <6.0.0
  Version out of range: transformers@6.0.0 >=5.2.0 <6.0.0
Incongruent: bolson@0.3.1
  Version out of range: fast-vect@1.1.0 >=0.7.0 <0.8.0
Incongruent: concur-react@0.5.0
  Version out of range: avar@5.0.0 >=3.0.0 <4.0.0
  Version out of range: foldable-traversable@6.0.0 >=4.1.1 <5.0.0
  Version out of range: free@7.0.0 >=5.2.0 <6.0.0
  Version out of range: nonempty@7.0.0 >=5.0.0 <6.0.0
  Version out of range: profunctor-lenses@8.0.0 >=6.3.0 <7.0.0
  Version out of range: tailrec@6.1.0 >=4.1.1 <5.0.0
Incongruent: deku@0.9.8
  Version out of range: fast-vect@1.1.0 >=0.7.0 <0.8.0
Incongruent: dom-indexed@11.0.0
  Version out of range: web-clipboard@5.0.0 >=4.0.0 <5.0.0
Incongruent: fetch-yoga-json@1.1.0
  Version out of range: yoga-json@4.0.1 >=3.0.2 <4.0.0
Incongruent: httpurple@3.0.1
  Version out of range: routing-duplex@0.7.0 >=0.6.0 <0.7.0
Incongruent: httpurple-yoga-json@1.0.0
  Version out of range: yoga-json@4.0.1 >=2.0.0 <4.0.0
Incongruent: jelly-hooks@0.3.0
  Version out of range: jelly-signal@0.3.0 >=0.3.1 <0.4.0
Incongruent: mdast-util-from-markdown@0.2.1
  Version out of range: aff@7.1.0 >=6.0.0 <7.0.0
  Version out of range: aff-promise@4.0.0 >=3.0.0 <4.0.0
  Version out of range: console@6.0.0 >=5.0.0 <6.0.0
  Version out of range: debug@6.0.2 >=5.0.0 <6.0.0
  Version out of range: foldable-traversable@6.0.0 >=5.0.1 <6.0.0
  Version out of range: psci-support@6.0.0 >=5.0.0 <6.0.0
  Version out of range: spec@7.1.0 >=5.0.1 <6.0.0
  Version out of range: spec-discovery@8.0.1 >=6.0.0 <7.0.0
Incongruent: ocarina@1.5.2
  Version out of range: fast-vect@1.1.0 >=0.7.0 <0.8.0
Incongruent: ps-cst@1.2.0
  Version out of range: spec@7.1.0 >=5.0.0 <6.0.0
Incongruent: rdf@0.1.0
  Version out of range: parsing@10.2.0 >=9.1.0 <10.0.0
Incongruent: react-basic-storybook@2.0.0
  Version out of range: record-studio@1.0.4 >=1.0.1 <1.0.2
Incongruent: rito@0.3.2
  Version out of range: deku@0.9.8 >=0.8.1 <0.9.0
  Version out of range: fast-vect@1.1.0 >=0.7.0 <0.8.0
Incongruent: tecton-halogen@0.1.3
  Version out of range: tecton@0.1.6 >=0.1.2 <0.1.3
```

PR #569, opened 2022-12-13
Move bindings to 'foreign' package and split GitHub module

thomashoneyman:
This PR continues the cleanup effort on the registry code base by taking a few actions necessary to support #563.

My primary motivation in this PR is to split the `Foreign.GitHub` module into two modules -- one for bindings, and one for the specific requests our application needs to make. This is necessary because we need the low-level bindings for cases where we don't have access to our application monad (such as when we fail to decode the workflow event and want to post an error comment), but we don't want to use them when we _do_ have that access (such as when we want to use our `Cache` and `Log` effect when making requests to GitHub).

However, this is really indicative of a deeper issue we have, which I discussed in #563: we've been putting a lot of app-specific logic into the "Foreign" namespace, with the GitHub module being the worst culprit, and haven't been careful about logging and throwing exceptions in that code. Plus, some of the code in that name space doesn't even have foreign bindings!

So in this pull request I've moved code with non-app-specific FFI into the `foreign` project, returning to @f-f's original proposal for the monorepo structure. I've moved code that has no FFI but was in the Foreign namespace into the app, mostly by moving it to a new CLI directory for shelling out to CLI tools like `purs`.

I also made two other quality-of-life changes. First, all modules already implicitly are going to have the `Registry` prefix, so I've omitted the enclosing `Registry` directory in each workspace just to keep the folders a little less nested. Second, I noticed that the `LenientVersion` and `LenientRange` modules are only relevant to the legacy code, so I moved them into the Legacy namespace.

As usual, all tests pass and I've tested that the legacy importer continues to work. In the PR comments I'll describe the changes to the GitHub module.


app/src/App/GitHub.purs
```diff
@@ -0,0 +1,256 @@
+module Registry.App.GitHub
```
thomashoneyman, line 1: We previously had one GitHub module, which implemented bindings and some requests we use in the application. Those requests are going to end up using our effects system, so they have to be pulled into the application. I don't want to pull everything over, however, because there are times when we do still use the low-level bindings.

This module implements requests and data types we plan to use in our application with no promises about them matching the GitHub API -- it's just ad-hoc whatever we end up needing to use. However, this is really just a layer on top of the bindings, as you can see by the liberal use of the Request type from the foreign bindings.

foreign/src/Foreign/GitHub.purs
```diff
@@ -0,0 +1,261 @@
+module Registry.Foreign.GitHub
```
thomashoneyman: This module implements low-level bindings to Octokit, along with helpers for constructing and sending requests via Octokit, which you can then build requests on top of.

foreign/test/Foreign/JsonRepair.purs
```diff
@@ -0,0 +1,57 @@
+module Test.Registry.Foreign.JsonRepair (spec) where
```
thomashoneyman, line 1: I did not change the code for any of these modules that have been moved, other than adjusting imports and adjusting some names. **The only logic that has changed significantly in this PR are the two GitHub modules.**

app/src/App/API.purs
```diff
@@ -1505,3 +1505,26 @@ isVersionInMetadata version (Metadata metadata) = versionPublished || versionUnp
...
+
+newtype WorkflowEvent = WorkflowEvent
+  { issueNumber :: IssueNumber
+  , body :: String
+  , username :: String
+  }
+
+derive instance Newtype WorkflowEvent _
+
+decodeWorkflowEvent :: Json -> Either String WorkflowEvent
```
f-f: This refactoring removes any mention of GitHub, but it seems critical to note that somewhere, since this payload comes from GitHub and one might want to look at the docs.

app/test/Main.purs
```diff
@@ -266,7 +224,7 @@ decodeEventsToOps = do
             ]
         }
 
-    res <- API.readOperation "test/fixtures/package-set-update_issue_created.json"
+    res <- API.readOperation "test/_fixtures/package-set-update_issue_created.json"
```
f-f, line 227: Why this rename?

foreign/src/Foreign/Tar.purs
```diff
@@ -0,0 +1,17 @@
...
+  ) where
+
+import Prelude
+
+import Data.Array as Array
+import Data.Function.Uncurried (Fn1, runFn1)
+import Data.Maybe (Maybe)
+import Effect (Effect)
+
+foreign import getToplevelDirImpl :: Fn1 String (Effect (Array String))
```
f-f, line 12: I don't understand why the Tar module has split in two?

f-f (commented):
I understand the motivation for the refactoring, but I don't understand why some modules have to be split in two? I think we could keep them in one piece and have a section with higher level bindings and one with lower level ones?

app/src/App/API.purs
thomashoneyman: I agree -- in fact, I had a link to the GitHub docs for workflow events, but I must have deleted it. I'll put it in there.

app/test/Main.purs
thomashoneyman: Just to sort the directory apart from the ones containing PureScript modules. It's how the library already named its fixtures folder:
https://github.com/purescript/registry-dev/tree/master/lib/test/_fixtures

so this is bringing them into parity.

foreign/src/Foreign/Tar.purs
thomashoneyman: I put the FFI bindings to the `tar` library into `foreign`, and I put calls to the `tar` / `gzip` CLI into the app, according to the principle that foreign is for bindings. To use the `Registry.Foreign.Tar` module you need to have installed the `node-tar` package, but to use the `Registry.App.CLI.Tar` module you don't; instead, you need to have the `tar` tool installed.

thomashoneyman:
> I understand the motivation for the refactoring, but I don't understand why some modules have to be split in two? I think we could keep them in one piece and have a section with higher level bindings and one with lower level ones?

I split the GitHub requests in two because my longer-term aim is for the app-specific requests to be an effect, though I haven't done it yet. Then the lower-level is for GitHub types and the Octokit-based implementation, and then the higher-level is our effect module. I suppose everything could go into the effect module, modulo any circular imports, but it just felt cleaner to me for them to be separate. If you disagree then I can merge them back together.

As far as the `Tar` module, I explained my reasoning here:
https://github.com/purescript/registry-dev/pull/569#discussion_r1048610980

app/src/App/API.purs
thomashoneyman: Also, the reason this moved to the API module is because ultimately we are going to have separate entrypoints for the server receiving HTTP requests and the GitHub workflow receiving workflow events. I think it makes the most sense for the workflow event decoding to be together with e.g. `decodeOperation`, which is here in the API file for now.

foreign/src/Foreign/Tar.purs
f-f: This makes sense now, thanks!

f-f:
Thanks for the clarifications! I think it would be less confusing to have the two GitHub modules be squashed together in one if nothing else is using the lower level one. I don't have a strong opinion on this though, and we can move forward with this if you'd like so

f-f (approved):

thomashoneyman:
> if nothing else is using the lower level one

There are some places where we use the lower-level one directly, such as responding to a GitHub workflow event before we're able to construct an environment and get into our effects. There we need to make a comment using only the low-level stuff.

We will also use the low-level stuff only when implementing e.g. a `Log` effect that comments to GitHub. Our app-level requests will use the logging and caching effects, so the GitHub module will need those, but then we can't use the low-level GitHub code to implement the logging unless we split out the commenting stuff only. So I think it's best we keep at least some low-level GitHub code separate.

app/src/App/API.purs
```diff
@@ -1506,16 +1506,19 @@ isVersionInMetadata version (Metadata metadata) = versionPublished || versionUnp
   versionPublished = isJust $ Map.lookup version metadata.published
   versionUnpublished = isJust $ Map.lookup version metadata.unpublished
 
-newtype WorkflowEvent = WorkflowEvent
+-- | An event triggered by a GitHub workflow, specifically via an issue comment
+-- | or issue creation.
+-- | https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment
+newtype IssueEvent = IssueEvent
```
thomashoneyman, line 1512: What do you think of renaming this to `IssueEvent`, as we're decoding it expecting fields specific to the issue_comment or issue_created events (where with issue_created we essentially treat it as a comment)? Or do you have other ideas?

f-f:
> So I think it's best we keep at least some low-level GitHub code separate.

Sounds good. Would you mind writing a docstring in the low-level module detailing this reasoning?

app/src/App/API.purs
f-f: Great, I like this!

thomashoneyman:
> Sounds good. Would you mind writing a docstring in the low-level module detailing this reasoning?

Will do! Also, if it will be more clear, I can rename the low-level module to be `Registry.Foreign.Octokit`, since it's specifically binding to that library and providing an interface to its `request` and `paginate` methods. Maybe that's more clear than naming it "GitHub"?

thomashoneyman:
@f-f I applied the Octokit rename in the last commit, but if you want to stick with Foreign.GitHub I can revert it.

f-f:
@thomashoneyman this is all great, thanks!

PR #571, opened 2022-12-14
Do not allow packages to skip solving

thomashoneyman:
As per discussion in #531, we would like to ensure that packages uploaded to the registry are solvable. Accordingly, the `verifyResolutions` check now always applies -- there is no exemption for legacy packages.


app/src/Registry/App/API.purs
```diff
@@ -617,8 +617,8 @@ publish source publishData@{ name, ref, compiler, resolutions } (Metadata inputM
   -- Now that we have the package source contents we can verify we can compile
   -- the package. We skip failures when the package is a legacy package.
   compilationResult <- case eitherVerifiedResolutions of
-    -- We do not throw an exception if we're bulk-uploading legacy packages
-    Left error | source == Importer || (source == API && isLegacyImport) ->
+    -- We do not throw an exception if we're bulk-uploading legacy packages.
+    Left error | source == Importer ->
```
thomashoneyman, line 621: The legacy importer pretends to be the `API` source when being run as a script via workflow events.

app/src/Registry/App/API.purs
```diff
@@ -616,23 +620,15 @@ publish source publishData@{ name, ref, compiler, resolutions } (Metadata inputM
...
-    Left error ->
-      throwWithComment error
-    Right verified ->
-      compilePackage { packageSourceDir: packageDirectory, compiler, resolutions: verified }
+  compilationResult <- compilePackage { packageSourceDir: packageDirectory, compiler, resolutions: verifiedResolutions }
 
   case compilationResult of
     Left error
+      -- We allow bulk-uploaded legacy packages to fail compilation because we
+      -- do not necessarily know what compiler to use with them.
```
thomashoneyman: I'm assuming we aren't going to compile true legacy packages? If we are, we'll need a heuristic for what compiler to use for them.

f-f:
As I detailed in https://github.com/purescript/registry-dev/issues/531#issuecomment-1342536699, we should figure out (and spec properly) how the Trustees are going to work before merging this, because once this is in place then we'll have trouble with folks not being able to publish packages and not being able to do anything about it.

thomashoneyman:
I understand, though that means that packages may be uploaded that cannot be solved, which we may end up having to delete later. So there's some urgency around figuring out the registry trustees situation.

f-f:
Agreed, we should probably have it as first priority spec-wise.

thomashoneyman:
As per the registry call this morning we are going to go ahead and institute this check, and we'll resolve solver issues manually, as they arise, while we complete the registry trustees spec. By "manually" I mean we will either not address the issue, or we will special-case the package in the legacy importer to new bounds.

f-f (approved):
Always great when we get to remove special casing 😄

PR #574, opened 2022-12-26
Introduce structured effects

thomashoneyman:
IT'S A CHRISTMAS MIRACLE!

Yes, at long last, I've finished the `app-effects` pull request. This introduces structured effects to the registry. Along the way I found and fixed a _lot_ of little bugs, added lots of logging and (much better) caching, and set the foundation for new features like the HTTP API.

For this PR I suggest you try the code before you review it. If you have a valid cache, then you can run the full legacy import (including mock publishing) in about 4 minutes on my machine. Without the cache it a while to build missing manifests, but subsequent runs are fast. The legacy importer is broken on `master` right now, so a dry run will "publish" a few packages.

Notably, you only need to have a valid `GITHUB_TOKEN` environment variable for a dry run. None of the rest are even read from the environment.

```console
git clone git@github.com:purescript/registry-dev && cd registry-dev
git checkout trh/app-effects
nix develop
registry-importer dry-run
```

One of the first things you'll notice is that logs are written to the console (with colored output). By default non-debug logs are written, but you can choose the verbosity that suits your task. We also print a link to a logfile containing the full debug logs. Logfiles have names like `legacy-importer-2022-12-25T21:25:27.log` and are created by the `Log` effect if you use the file system handler. They contain timestamped output like this:

```
[2022-12-25T21:26:11.212Z DEBUG] Reading metadata for deku
[2022-12-25T21:26:11.213Z DEBUG] Registry repo up to date, reading from cache...
[2022-12-25T21:26:11.213Z DEBUG] Read cache entry for AllMetadata in memory.
[2022-12-25T21:26:11.213Z DEBUG] Metadata validated. Fetching package source code...
[2022-12-25T21:26:11.214Z DEBUG] Using legacy Git clone to fetch package source at tag: { owner: "mikesol", ref: "v0.9.9", repo: "purescript-deku" }
[2022-12-25T21:26:12.913Z DEBUG] Cloned package source to /tmp/nix-shell.8Ct0ZI/nix-shell.xjJVGX/tmp-6792-bwwsk6lyhsGP/purescript-deku
[2022-12-25T21:26:12.914Z DEBUG] Getting published time...
[2022-12-25T21:26:12.922Z DEBUG] Package downloaded to /tmp/nix-shell.8Ct0ZI/nix-shell.xjJVGX/tmp-6792-bwwsk6lyhsGP/purescript-deku, verifying it contains a src directory...
[2022-12-25T21:26:12.924Z DEBUG] Package contains .purs files in its src directory.
[2022-12-25T21:26:12.924Z INFO] [NOTIFY] Package source does not have a purs.json file. Creating one from your bower.json and/or spago.dhall files...
[2022-12-25T21:26:12.925Z DEBUG] Listing tags for purescript/package-sets
[2022-12-25T21:26:12.925Z DEBUG] Read cache entry for GET /repos/purescript/package-sets/tags in memory.
```

We can use this log as a little tour through the new effects (you can see [all of the effects in the `Registry.App.Effects` directory](https://github.com/purescript/registry-dev/tree/trh/app-effects/app/src/App/Effect)).

**Registry**
At the beginning of the log we are `Reading metadata for deku`. This comes from the `Registry` effect, which handles interacting with registry resources like reading, writing, and deleting manifests, reading and writing metadata, reading and writing package sets, mirroring legacy content, and so on. In user code you generally just have to write `readMetadata "deku"` when you need metadata, and the mechanics of retrieving that data are abstracted away.

**Git**
Next, the log reports that the `Registry repo up to date`. This comes from the `Git` effect for pulling, committing, and so on. I've taken great care to make our git operations reliable, including verifying regularly that we are in sync with the origin (or are ahead if we're pushing, or are behind if we're pulling, and so on). The main handler for the `Registry` effect relies on the `Git` effect to make sure we have registry files available in a local Git checkout, and the `Git` effect makes sure that we haven't fallen out of sync.

Repositories are now configurable! You can choose to point at a fork of the registry, for example. Everything else will work the same.

**TypedCache**
Next, we `Read cache entry for AllMetadata in memory`. This comes from the `TypedCache` effect. This is the most intimidating of the new effects, because I've made the cache well-typed: when you `get` or `put` a value, you receive the actual type you wanted, not JSON that you must decode.

A typed cache is super important for a few reasons. First, our use of cache keys was totally uncontrolled and we could easily run into collisions; this is way harder to do when the values are typed. Second, you can't cache a `ManifestIndex` or other large structure in memory when the cache is untyped, because you have to serialize it on write and deserialize on read and that's too slow. (Or you can coerce it, which is unsafe.). Third, deciding on a single serialization format means the cache isn't flexible enough to accommodate storing most things in JSON, but other things as, say, a `Buffer` (e.g. tarball) on the file system.

The `TypedCache` effect supports many independent caches, which means individual caches can be handled separately and code outside the registry proper (like the `scripts` directory) can still use the typed cache. For example, the `Storage` effect uses the cache to store tarballs on the file system. The `Registry` effect, on the other hand, uses the cache to store manifests and metadata in memory. It never stores any cache data on the file system because we rely on the Git repositories as the source of truth instead. Other caches like the `Importer` cache use an in-memory cache that falls back to a filesystem cache on cache miss.

**Notify**
That `[NOTIFY]` message indicates a human-readable message we want to report to users. It's handled by the `Notify` effect, and when running scripts like the legacy importer it just writes to the console. An API could push a message over a websocket, and on GitHub we comment this message on the relevant issue.

**GitHub**
Towards the end we have `Listing tags for purescript/package-sets`. This comes from the `GitHub` effect, which is for interacting with the GitHub API. We really need to start making use of conditional requests with this effect — I did more digging, and the `If-Last-Modified` header doesn't work, but the `If-None-Match` header given an `etag` does work. One of us ought to modify the `Octokit.js` bindings so that we can read the headers from responses.

The small snippet of the logs doesn't cover all the effects. Here are the others:

**Storage**
The  `Storage` effect manages interactions with the storage backend, such as uploading, downloading, and deleting package tarballs. It has handlers for read-only situations (like the legacy importer dry-run) and for the S3 backend.

**Pursuit**
The `Pursuit` effect manages interactions with Pursuit, like publishing packages. It's a tiny effect for now, but it will grow when we replace `purs publish` with our own implementation, and I think there are some wonderful opportunities to rewrite Pursuit in PureScript and integrate it more closely with the registry.

**PackageSets**
The `PackageSets` effect is for upgrading the package sets. It provides `upgradeAtomic` and `upgradeSequential`.

**Env**
The `Env` effect is for providing `Reader` environments. This is only lightly used, mainly to provide access to things like the pacchettibotti keys or the github event data.

---

OK, so we've got a set of effects, and handlers for those effects. Why have so many files changed? First, when you replace the foundation of the app it's going to affect everything. Second, now that we can be precise with effects, a lot of massive functions in the app are much better off being refactored to be more modular (like the API). Finally, we have historically had the assumption that the whole registry is oriented around GitHub events, and that is no longer true. We can't just assume in the middle of our API pipeline that we can look up if a `username` is on the trustees team to re-sign a payload, for example.

That means we have a big change: the `API` module no longer anticipates that it's being called as part of a GitHub event. Instead, the main operations (`publish`, etc.) know they can be called from a script (like the legacy importer) or from a payload coming over the wire in the upcoming HTTP API, or from a GitHub event. Other operations (`packageSetUpdate`) know they must come from a GitHub event because we rely on GitHub for authentication.

Since the API module no longer presupposes GitHub events, we now have a `Main` module. This module handles all the GitHub stuff: it decodes the operation from the `GITHUB_EVENT_DATA`, re-signs the payload if the GitHub event is an authenticated event sent by a trustee, and does other GitHub-specific initialization before passing things off to the API module.

This lays the groundwork for the HTTP API, which no longer has to work around GitHub assumptions.

#### Next Steps

Once we merge this pull request I believe we only need to do one thing: The `.cache` dir is now part of the `scratch` directory where all other files get put as part of the registry processes. We need to update the GitHub workflows on the `registry` repo to make sure that directory is being cached.

Otherwise, when this merges we just need to run the legacy import and make sure everything works as expected. I've tested the code thoroughly and I feel confident, but you never know when you actually deploy whether some small configuration option is off and needs to be tweaked.

I'll leave some comments in the code to aid review.


app/src/App/Auth.purs
```diff
@@ -27,11 +27,16 @@ sshKeyPath = "id_ed25519"
...
 -- Payload verification is based on this description:
 -- https://www.agwa.name/blog/post/ssh_signatures
-verifyPayload :: NonEmptyArray Owner -> AuthenticatedData -> Aff (Either String String)
-verifyPayload owners { email, signature, rawPayload } = do
+--
+-- We take pacchettibotti as an extra owner because pacchettibotti can always
+-- sign authenticated transfers. This is not sufficient to blanket authorize
+-- things, as someone still needs to use the pacchettibotti credentials to
+-- actually sign the payload.
+verifyPayload :: Owner -> NonEmptyArray Owner -> AuthenticatedData -> Aff (Either String String)
```
thomashoneyman, line 35: We used to look up the owners in metadata, and if there were owners we'd look up the pacchettibotti credentials, and then we'd sign the payload and insert pacchettibotti into the owners. I found it confusing, and it's a little more straightfowrard to pass the pacchettibotti owner as an argument to the verification function.

app/src/App/API.purs
```diff
@@ -2,739 +2,581 @@ module Registry.App.API where
...
+        Log.exit $ "Cannot register " <> printedName <> " because the payload did not include a 'location' field."
+      Just location ->
+        pure $ Metadata { location, owners: Nothing, published: Map.empty, unpublished: Map.empty }
+
+    Just metadata -> case payload.location of
+      -- The user can add a new version of their package if it comes from
+      -- the same location listed in the metadata OR if they do not provide
+      -- a location.
+      Nothing -> pure metadata
+      Just location | (un Metadata metadata).location == location -> pure metadata
```
thomashoneyman, line 335: As discussed in the registry call @f-f we've relaxed the restriction on locations being unique. You can now register a package using a location that is already in use.

app/src/App/API.purs
```diff
@@ -2,739 +2,581 @@ module Registry.App.API where
...
-      comment $ "WARNING: Package tarball is " <> show bytes <> "bytes, which exceeds the warning threshold of " <> show maxWarnBytes <> " bytes.\ncc: @purescript/packaging"
-  log "Hashing the tarball..."
-  hash <- liftAff $ Sha256.hashFile tarballPath
-  log $ "Hash: " <> Sha256.print hash
+      Notify.notify $ "WARNING: Package tarball is " <> show bytes <> "bytes, which exceeds the warning threshold of " <> show maxWarnBytes <> " bytes."
+
+  -- If a package has under ~30 bytes it's about guaranteed that packaging the
+  -- tarball failed. This can happen if the system running the API has a non-
+  -- GNU tar installed, for example.
+  let minBytes = 30.0
```
thomashoneyman, line 509: I was ready to publish this PR an hour earlier than I did, but the publish pipeline was weirdly producing empty tarballs and I couldn't figure it out. Turns out my Mac by default has an old, non-GNU tar installed, and it works somewhat differently. When I opened VS Code it overrode the nix shell for some reason. I've added this check because we definitely don't want to be pushing up empty tarballs!

app/src/App/CLI/Licensee.purs
```diff
@@ -35,24 +33,21 @@ detect directory = do
...
           { licenses: CA.array $ CA.Record.object "SPDXIds"
               { spdx_id: CA.string }
           }
 
       case parse result.stdout of
         Left error -> do
-          Console.log "Licensee failed to decode output: "
-          Console.log error
-          Console.log "arising from the result: "
-          Console.log result.stdout
```
thomashoneyman, line 49: We should not use `Console.log` anymore. We should use the logging effect. If we run something in `Aff` like a CLI tool then we should have it return its output. If we need on-demand logs then we can coordinate with a `Halogen.Subscription` or an `AVar`.

app/src/App/CLI/Wget.purs
```diff
@@ -1,20 +0,0 @@
-module Registry.App.CLI.Wget where
```
thomashoneyman, line 1: We download tarballs from the storage backend using Affjax.Node now.

app/src/App/Effect/Env.purs
```diff
@@ -0,0 +1,159 @@
...
+lookupOptional (EnvKey { key, decode }) = liftEffect $ Process.lookupEnv key >>= case _ of
+  Nothing -> pure Nothing
+  Just "" -> pure Nothing
+  Just value -> case decode value of
+    Left error -> Exception.throw $ "Found " <> key <> " in the environment with value " <> value <> ", but it could not be decoded: " <> error
+    Right decoded -> pure $ Just decoded
+
+-- | Look up a required environment variable, throwing an exception if it is
+-- | missing, an empty string, or present but cannot be decoded.
+lookupRequired :: forall m a. MonadEffect m => EnvKey a -> m a
```
thomashoneyman, line 95: We had a lot of code looking up an arbitrary key like `GITHUB_TOKEN` and throwing if it isn't present. The code here establishes a known set of env vars that can be present.

app/src/App/Effect/Git.purs
```diff
@@ -0,0 +1,634 @@
...
+          Right _ -> do
+            Log.debug $ "Found repo at path " <> path <> ", pulling latest."
+            result <- gitPull key
+            pure result
+
+    now <- Run.liftEffect nowUTC
+    debouncers <- Run.liftEffect $ Ref.read env.debouncer
+    case Map.lookup path debouncers of
+      -- We will be behind the upstream by at most this amount of time.
+      Just prev | DateTime.diff now prev <= Minutes 1.0 ->
```
thomashoneyman, line 241: I've taken a lot of care in this module and elsewhere to make sure the file system remains consistent with the remote git repos. This doesn't matter much for a GitHub event, but it certainly will for a server. Ideally, before we take any operation that relies on the state of a local repository, we sync with the upstream. But of course if you go to read 1000 manifests and sync every time it's way too slow.

So this institutes a little debouncer that sets the minimum interval at which we verify we're still in sync. I thought a minute was reasonable, and the code is quite fast.

app/src/App/Effect/Git.purs
```diff
@@ -0,0 +1,634 @@
...
+-- | Interpret the GIT effect given a handler, eliminating it from the effects.
+runGit :: forall r a. (Git ~> Run r) -> Run (GIT + r) a -> Run r a
+runGit handler = Run.interpret (Run.on _git handler Run.send)
+
+-- | How to sync a locally-checked-out repository.
+-- |
+-- | - Autostash will stash any changes you have, pull, then apply your changes.
+-- | - ForceClean will remove untracked files and hard reset before pulling.
+-- | - OnlyClean will abort in the presence of untracked or modified files.
+data PullMode = Autostash | OnlyClean | ForceClean
```
thomashoneyman, line 146: We can now control how the local checkout should be synced. When doing dry-run dev work it's nice to use `Autostash` so you can apply changes over the remote and have them preserved while the program runs. For things where we want to be sure we're strictly in sync we can use `ForceClean` to forcibly reset the repository in the case it's out of sync. And if we want to be as safe as possible we can use `OnlyClean`, which will exit with an error if we ever get out of sync.

app/src/App/Effect/Log.purs
```diff
@@ -0,0 +1,149 @@
...
+derive instance Eq LogLevel
+derive instance Ord LogLevel
+
+data LogVerbosity = Quiet | Normal | Verbose
+
+derive instance Eq LogVerbosity
+derive instance Ord LogVerbosity
+
+class Loggable a where
+  toLog :: a -> Doc GraphicsParam
```
thomashoneyman, line 36: Borrowed from Spago, although since the whole codebase was just logging strings before I haven't made much use of the type class. We could take another pass through the app sometime to clean up the logs, but it isn't urgent.

app/src/App/Effect/Log.purs
```diff
@@ -0,0 +1,149 @@
...
+        let
+          time = Formatters.DateTime.format Internal.Format.iso8601DateTime now
+
+          level' = case level of
+            Debug -> "DEBUG"
+            Info -> "INFO"
+            Warn -> "WARN"
+            Error -> "ERROR"
+
+          formatted = Dodo.print Dodo.plainText Dodo.twoSpaces
```
thomashoneyman, line 115: I'm happy to change this format if there's something else that would be better. I don't really care — just wanted something timestamped.

app/src/App/Effect/PackageSets.purs
```diff
@@ -0,0 +1,567 @@
+-- | An effect for upgrading package sets, either in batch mode (all suggested
+-- | changes must go in together) or in sequential mode (as many changes as
+-- | possible will be added, and changes that would break the set are recorded.)
+module Registry.App.Effect.PackageSets where
```
thomashoneyman, line 4: This works the same as the old `PackageSets` module wrt the package set upgrades, just using effects now. You can compare them side-by-side.

app/src/App/Effect/Registry.purs
```diff
@@ -0,0 +1,691 @@
...
+        case cache of
+          Nothing -> do
+            Log.info "No cached manifest index, reading from disk..."
+            refreshIndex
+          Just cached -> pure $ reply cached
+      Right Changed -> do
+        Log.info "Manifest index has changed, replacing cache..."
+        refreshIndex
+
+  ReadMetadata name reply -> do
```
thomashoneyman: You'll notice that this file is largely just a layer on top of `Git` and `TypedCache`. I've taken care to keep the local checkout and the cache in sync as much as possible.

First, we no longer cache manifests and metadata to the file system; we use the local checkout as the file system cache. Second, we rely on the `Git` effect to tell if we are in sync if the local checkout has changed, and on change we replace the in-memory cache. Third, we don't just have the manifest index sitting in a ref in memory that can be arbitrarily written to — all writes are controlled by this module.

app/src/App/Effect/TypedCache.purs
```diff
@@ -0,0 +1,345 @@
...
+-- |   Package name version next -> mkExists $ AsBuffer (formatPackageVersion name version) next
+-- | ```
+type FsEncoder key = forall b z. key z b -> Exists (FsEncoding z b)
+
+-- | A box used with `Exists` to capture the encoding associated with values
+-- | of a particular key.
+data FsEncoding :: (Type -> Type -> Type) -> Type -> Type -> Type
+data FsEncoding z b a
+  = AsJson String (JsonCodec a) (z a b)
+  | AsBuffer String (z Buffer b)
```
thomashoneyman: Essentially, these are serialization formats: sometimes we want a cache backed by JSON, sometimes backed by a raw buffer. We can add more if we ever need them, without having to add more handlers.

app/src/App/Effect/TypedCache.purs
```diff
@@ -0,0 +1,345 @@
...
+data CacheValue
+
+toCacheValue :: forall a. a -> CacheValue
+toCacheValue = unsafeCoerce
+
+type MemoryEncoder key = forall b z. key z b -> Exists (MemoryEncoding z b)
+
+-- An encoding for keys so they can be stored in a map.
+data MemoryEncoding :: (Type -> Type -> Type) -> Type -> Type -> Type
+data MemoryEncoding z b a = Key String (z a b)
```
thomashoneyman: This is done because we don't have an `Ord` instance for cache key types, and we can't just derive something because we need to know how something like `MyKey PackageName Version` should be turned into a map key (generally it'll be calling all the `print` functions, but there's no way to tie them together except a class, and I don't want to do that).

app/src/App/Effect/TypedCache.purs
```diff
@@ -0,0 +1,345 @@
...
+      Left _ -> do
+        Log.debug $ "No cache found for " <> id <> " at path " <> path
+        pure $ reply Nothing
+      Right buf -> do
+        Log.debug $ "Read cached buffer for " <> id
+        pure $ reply $ Just buf
+
+  AsJson id codec (Reply reply) -> do
+    let path = Path.concat [ cacheDir, safePath id ]
+    Run.liftAff (Aff.attempt (FS.Aff.readTextFile UTF8 path)) >>= case _ of
```
thomashoneyman: We used to have a lot of file system reads and other `Aff` code that wasn't wrapped in `attempt`. We should really do our best to always catch exceptions when `Aff` code could throw, so that we can properly log the errors and not just kill the process.

app/src/App/Legacy/Manifest.purs
```diff
@@ -375,45 +409,98 @@ fetchLegacyPackageSets = do
...
+        Log.warn $ "Failed to retrieve package sets for some tags: " <> String.joinWith ", " (map (fst >>> un RawVersion) results.fail)
+
+      let
+        convertedSets :: Array LegacyPackageSetUnion
+        convertedSets = map convertPackageSet results.success
+
+        -- TODO: As we've discovered, this union is backwards and needs to be fixed. But that's for
+        -- another PR to take care of.
+        merged :: LegacyPackageSetUnion
+        merged = Array.foldl (\m set -> Map.unionWith Map.union set m) Map.empty convertedSets
```
thomashoneyman, line 447: cc: @MonoidMusician

app/test/App/API.purs
```diff
@@ -0,0 +1,157 @@
+module Test.Registry.App.API (spec) where
```
thomashoneyman, line 1: All the tests are still in place, they've just moved to match the directory structure of the app.

foreign/src/Foreign/Octokit.js
```diff
@@ -12,21 +12,12 @@ export const newOctokitImpl = (authToken) => {
...
-          console.log(`Retrying after ${retryAfter} seconds!`);
           return true;
         }
       },
-      onAbuseLimit: (_, options) => {
-        octokit.log.error(
-          `Abuse detected for request ${options.method} ${options.url}`
-        );
-      },
+      onSecondaryRateLimit: (retryAfter, options) => {},
```
thomashoneyman, line 20: I don't know if something changed in a minor version or something, but I got an exception from Octokit about not having a `onSecondaryRateLimit` handler.

foreign/src/Foreign/Tmp.js
```diff
@@ -2,15 +2,7 @@ import { setGracefulCleanup, dirSync } from "tmp";
...
 
-export function mkTmpDirImpl() {
-  return function () {
-    const tmpobj = dirSync();
-    return tmpobj.name;
-  };
-}
-
-// TODO: we should do proper bracketing when creating the temporary directory
-// so that if the program crashes we still remove it.
```
thomashoneyman, line 13: This is what `setGracefulCleanup()` already does.

lib/src/Operation/Validation.purs
```diff
@@ -42,10 +43,13 @@ import Registry.Version (Version)
 containsPursFile :: FilePath -> Aff Boolean
 containsPursFile parent = flip catchError (\_ -> pure false) do
   children <- FS.Aff.readdir parent
-  stats <- traverse (\path -> { path, stats: _ } <$> FS.Aff.stat path) children
```
thomashoneyman, line 45: cc: @colinwahl this was incorrect and is causing all publishing to fail with a "no purs files" error — sorry I didn't catch it in review. `readdir` returns relative paths, so as we recurse through the tree we need to successively concatenate the parent directory.

scripts/src/LegacyImporter.purs
```diff
@@ -71,224 +83,232 @@ derive instance Eq ImportMode
 
 parser :: ArgParser ImportMode
 parser = Arg.choose "command"
-  [ Arg.argument [ "dry-run" ]
+  [ Arg.flag [ "dry-run" ]
```
thomashoneyman, line 88: Yikes. This is broken on master, and it's why the legacy importer isn't running.

scripts/src/LegacyImporter.purs
```diff
@@ -383,77 +402,88 @@ importLegacyRegistry existingRegistry legacyRegistry = do
...
-              Except.except contents.value
-
-  manifests <- lift $ for package.tags \tag -> do
-    manifest <- Except.runExceptT $ buildManifestForVersion tag
+      -- TODO: This will use the manifest for the package version from the
+      -- registry, without trying to produce a legacy manifest. However, we may
+      -- want to always attempt to produce a legacy manifest. If we can produce
+      -- one we compare it to the existing entry, failing if there is a
+      -- difference; if we can't, we warn and fall back to the existing entry.
+      Registry.readManifest package.name (LenientVersion.version version) >>= case _ of
```
thomashoneyman, line 430: cc: @f-f, @MonoidMusician

app/src/App/Effect/Registry.purs
```diff
@@ -0,0 +1,691 @@
...
+    index <- handleRegistryGit (ReadAllManifests identity)
+    case ManifestIndex.delete name version index of
+      Left error ->
+        Log.exit $ Array.fold
+          [ "Can't delete " <> formatted <> " from manifest index because it would produce unsatisfied dependencies:"
+          , printJson (Internal.Codec.packageMap (Internal.Codec.versionMap (Internal.Codec.packageMap Range.codec))) error
+          ]
+      Right updated -> do
+        commitResult <- Git.writeCommitPush (Git.CommitManifestEntry name) \indexPath -> do
+          Run.liftAff (ManifestIndex.removeFromEntryFile indexPath name version) >>= case _ of
```
thomashoneyman: No more just unconstrained inserting or deleting to the manifest index on disk! We need to validate the changes first. You shouldn't be able to easily break the index by arbitrarily deleting a manifest.

lib/src/ManifestIndex.purs
```diff
@@ -241,17 +259,17 @@ printEntry =
 -- | Given the root of a manifest index on the file system and a package name,
 -- | retrieve the manifests associated with that package, in sorted order from
 -- | lowest to highest version.
-readEntryFile :: forall m. MonadAff m => FilePath -> PackageName -> m (Either String (NonEmptyArray Manifest))
+readEntryFile :: FilePath -> PackageName -> Aff (Either String (NonEmptyArray Manifest))
```
f-f: How is this one better? The previous version runs in any monad that implements `MonadAff` rather than requiring a `liftAff`. I understand this saves a few `liftAff` sprinkled around here, but I would optimise this code for consumers?

scripts/src/LegacyImporter.purs
f-f: So what's happening here is that if this is a legacy package that has a `purs.json` then we take that in?
But well, if it has a `purs.json` then how is it a legacy package? Maybe we should just not publish this package then - I think the moment a package author is opting into having a `purs.json` then they take the publishing into their own responsibility and should not rely on the legacy importer.

app/src/App/API.purs
```diff
@@ -2,739 +2,577 @@ module Registry.App.API where
...
 import Registry.Operation.Validation as Operation.Validation
+import Registry.Owner as Owner
 import Registry.PackageName as PackageName
 import Registry.PackageSet as PackageSet
 import Registry.Range as Range
 import Registry.Sha256 as Sha256
 import Registry.Solver as Solver
 import Registry.Version as Version
-import Sunde as Process
+import Run (AFF, EFFECT, Run)
```
f-f, line 78: I thought we agreed in #563 that this PR would be based on transformers but it looks like it's using Run instead?

app/src/App/Effect/TypedCache.purs
f-f: How is `attempt` different from `try`?

foreign/src/Foreign/Tmp.js
f-f: Are we calling that somewhere?

app/src/App/Effect/TypedCache.purs
f-f: Could you add the above as a comment in the code?

app/src/App/Effect/TypedCache.purs
f-f: Could you add the above as a comment in the code?

scripts/src/LegacyImporter.purs
thomashoneyman: The legacy importer takes in all packages from the legacy registry and attempts to build manifests for them.

As an optimization, this line calls `readManifest` to read the manifest for this package version from the existing manifest index in the `registry-index` repo. If there is one, then we don't bother building the legacy manifest and just return what we found in the manifest index. This doesn't have anything to do with the package's source code, it's just essentially using the manifest index as a cache.

lib/src/ManifestIndex.purs
thomashoneyman: I can put this back. These can be in `MonadAff` if we want to do that for `Aff-` functions in `lib`.

I just removed it here because we in general don't use MTL in the library — for example, the operation validation, and hash creation, and stuff like that is just in `Aff` directly. I'd like to be consistent one way or the other.

app/src/App/API.purs
thomashoneyman: I suggested Run, then a take on MTL, and in that thread said I was going to do the take on MTL. However, we discussed this the same week in the registry call and the consensus was that it doesn't really matter which one we use (we can replace it with the other one if we have to), and because there is no obligation for consumers of the library to use it (the `lib` does not use Run at all).

As far as Spago: it _can_ use functions from the app with either MTL or Run, if you have specific ones you're depending on. With Run you will need to import the handlers (or write your own) for the effects you need. With MTL you will need to implement a monad and import “handlers” for each method of each effect you need. Both are doable. (If you depend on the registry-app, you're getting the Run dependency transitively anyway, so there's no difference in terms of dependencies).

The first cut at implementing this was in the MTL style. But with all the variants we need for testing, dry runs, local scripts, GitHub environment, and the API, and with lots of code not needing all effects, I ended up finding the Run version _much_ nicer to work with. It's way better to be able to swap a specific handler out rather than have to make a whole new monad if you need just one instance to be different. I strongly prefer that we stick with Run.

app/src/App/Effect/TypedCache.purs
thomashoneyman: They're the same, `attempt` is just monomorphic. We don't need to use `attempt` specifically, we just need to catch the errors.

foreign/src/Foreign/Tmp.js
thomashoneyman: It's already set in this file (line 3).

I read through the source code and the node-tmp library keeps a ref with every tmp directory created; it listens for the process exit event and cleans them all up. But only if you call `setGracefulCleanup()`.

scripts/src/LegacyImporter.purs
thomashoneyman: (But because of https://github.com/purescript/registry-dev/pull/574#discussion_r1057108024 it is broken anyways.)

scripts/src/LegacyImporter.purs
f-f: What is this cache used for?

lib/src/ManifestIndex.purs
f-f: I feel this is only tangentially related to MTL - since this is a library I think it's more generally useful for things to be in `MonadAff` rather than `Aff`, so that consumers can use straight `Aff` but also any other monad implementing `MonadAff`

thomashoneyman: That's fine with me! We can have the whole library use `MonadEffect` / `MonadAff` constraints instead of making the functions monomorphic. With regards to MTL — I'm just using that as it's the term for using `MonadX` classes afaik. I know we're not going to introduce our own classes.

scripts/src/LegacyImporter.purs
thomashoneyman: The "cache" (reading from the manifest index) is used to skip building a manifest from spago/bower files for a package at a ref when we already have a manifest in the index for that ref (we lenient-parse the ref to guess at its actual version). Essentially, when you call `buildLegacyManifest`, if we find the tag in the manifest index then we pretend like we successfully built it and return.

As noted in the comment, I don't really think we should be doing this. I feel like if we're going to automatically return success when the manifest exists that we should not have even called `buildLegacyManifest` for it in the first place (we're not really building a legacy manifest!). I think we could update this process to the following:

1. Fetch tags for every package in the legacy registry.
2. Lenient-parse all the tags as versions, then look up each package version in the manifest index; if an entry exists, filter it out of the legacy-import candidates.
3. For all refs that don't lenient-parse or don't have an existing entry, attempt to build them by calling `buildLegacyManifest`. Do not call `readManifest` in `buildLegacyManifest`.
4. I suggest we continue to cache the result of building a manifest for a ref on the file system (for example, this preserves errors we encountered when trying to build e.g `io@0.0.1`, which stops us from constantly trying to build manifests for package tags that are definitely broken). We already do this file system caching today in the legacy importer — it's the `ImportCache`.

cc: @MonoidMusician as you were recently working in this file.

app/src/App/Effect/Git.purs
```diff
@@ -0,0 +1,634 @@
...
+  CommitManifestIndex ->
+    "."
+  CommitMetadataIndex ->
+    Constants.metadataDirectory <> Path.sep <> "*.json"
+  CommitPackageSet version ->
+    Path.concat [ Constants.packageSetsDirectory, Version.print version <> ".json" ]
+  CommitLegacyRegistry ->
+    String.joinWith " " [ "bower-packages.json", "new-packages.json" ]
+  CommitLegacyPackageSets paths ->
+    String.joinWith " " paths
```
thomashoneyman: I've just remembered that this isn't going to work — this will add the string "bower-packages.json new-packages.json", which is invalid. This needs to be fixed so that we return an array of patterns, probably.

app/src/App/Main.purs
```diff
@@ -0,0 +1,307 @@
...
+    -- registry repository.
+    NotJson ->
+      pure Nothing
+
+    MalformedJson issue err -> do
+      let
+        comment = Array.fold
+          [ "The JSON input for this package update is malformed:"
+          , "\n"
+          , "```" <> err <> "```"
```
MonoidMusician: The triple backticks need newlines for markdown to recognize it as a codeblock

thomashoneyman: Thanks! I’ll make sure this gets updated.

app/test/Assert/Run.purs
```diff
@@ -0,0 +1,46 @@
...
+
+shouldContain :: forall f a r. Eq a => Foldable f => f a -> a -> Run (EXCEPT String + r) Unit
+shouldContain container elem =
+  when (elem `Foldable.notElem` container) do
+    Run.Except.throw (Utils.unsafeStringify elem <> "\n\nshould be a member of\n\n" <> Utils.unsafeStringify container)
+
+shouldNotContain :: forall f a r. Eq a => Foldable f => f a -> a -> Run (EXCEPT String + r) Unit
+shouldNotContain container elem =
+  unless (elem `Foldable.notElem` container) do
+    Run.Except.throw (Utils.unsafeStringify elem <> "\n\nshould be a member of\n\n" <> Utils.unsafeStringify container)
```
MonoidMusician: This comment is backwards 🙃

thomashoneyman: 🙃

lib/src/ManifestIndex.purs
thomashoneyman: I've done this throughout the lib, foreign, and app, except in places where we really do want `Aff` (ie. we're using its `Parallel` instance).

scripts/src/LegacyImporter.purs
MonoidMusician: I don't have strong opinions here. I think we should make a reasonable attempt to verify that the legacy packages we've uploaded match the current logic, but I don't care if it lives in the legacy uploader or a one-off script or something. From what I saw earlier on the old branch it was consistently producing indexes that matched. That is, we'll probably be verifying it by hand as we figure out the deleting and reuploading packages that we need to do.

app/src/App/API.purs
thomashoneyman: https://github.com/purescript/registry-dev/pull/575#issue-1518339547

thomashoneyman:
If you haven't reviewed this in a while, I'd recommend starting by running `registry-importer dry-run` and looking at the log file.

thomashoneyman:
I've now verified that this branch fully works in the registry import / package set updater workflows. See:

* Publish metadata: https://github.com/purescript/registry/commit/e084d82440a1769c97bc94cd876fb0ac14b9936a
* Publish manifest: https://github.com/purescript/registry-index/commit/c9f752834e591f2b495edc5acceb4ba6f6f0c362
* Update package set: https://github.com/purescript/registry/commit/b0e56d2ee21328c7308cbc6faf7e85b172832318
* Mirror package set: https://github.com/purescript/package-sets/commit/2ecfd5bc9c893771b7221bfa85f8f00d2db447ab
* Publish docs: https://pursuit.purescript.org/packages/purescript-json-codecs/2.0.0

f-f (approved):

PR #575, opened 2023-01-04
Migrate app effects to MTL

thomashoneyman:
In #574 I briefly mentioned that I had tried doing the effects with MTL, but I ultimately decided that Run provided better ergonomics. In this PR I have migrated our effects to use MTL instead of Run and implemented those changes throughout the app codebase (but not the scripts). That way @colinwahl @f-f @MonoidMusician you can see what an MTL flavor would look like.

I prefer the Run implementation, myself, but if you all think we should do MTL instead then I can finish the job and swap over the effects the rest of the way.


thomashoneyman:
I have a few notes on this implementation vs. the run one.

**Drawbacks**
It's significantly more tedious to implement all the instances and handlers for the various effects in the MTL version for multiple reasons.

The main problem is that the granularity of a handler for the class-based effects are at the function level. You can't just import a handler for the effect as a whole; instead, you have to import implementations for the type class members one-by-one. For the same reason, the handlers have to be implemented as independent functions, which makes it harder to share information among the handlers.

For example, in the `PackageSets` effect we can't just close over the handler's inputs to make a shared helper function in a `where` anymore; we have to pull the helper function out to the top level and do a lot more coordination work.

In total lines of code the MTL version is actually _less_ boilerplate than Run, but it feels harder to manage (to me). First, because there is nothing tying the `handle*` functions to the type class (you won't know you messed up the type signature until you go to use the handler in an instance; in the Run version, you explicitly case on the constructors and can't get it wrong). Second, because the boilerplate is pretty much all in the effect module itself; once you're putting together interpreters you're just importing the single handler you need wholesale. I _really_ like the boilerplate being confined to the module where you define the effect so you don't have to think about it elsewhere.

```purs
class MonadX m where
  a :: ...
  b :: ...
  c :: ...

handleA :: ...
handleA = ...

handleB :: ...
handleB = ...

handleC :: ...
handleC = ...

-- In another module...
instance MonadX MyApp where
  a = handleA
  b = handleB
  c = handleC
```

In Run you have the boilerplate of lifting things into Run, but at least the handlers are grouped together.

```purs
data X a
  = A ...
  | B ...
  | C ...

a :: ...
a = ...

b :: ...
b = ...

c :: ...
c = ...

handleX :: X ~> Aff
handleX = ...

-- In another module...
Run.interpret (Run.on _x handleX Run.send)
```

Another issue is that all information required by the handlers has to be passed via the environment. You have to put all the data the handlers need into a giant record in your `ReaderT`-based monad, and then either a) pull that data out of the environment for every type class member or b) use `MonadAsk` constraints everywhere. For example:

```purs
type FsEnv = { cacheDir :: FilePath }

-- MonadAsk version; this enforces that the environment in your app
-- _must_ have a 'cache' field with an 'FsEnv' in it, which can limit flexibility
-- and lead to duplication if the same info is needed in multiple envs (like a token)
handler1 :: forall m r. MonadAsk { cache :: FsEnv | r } m => ...

-- But it's easy enough to do the instance
instance Cache MyApp where
  method = handler1

-- Normal data version: this doesn't affect the app environment at all other than that
-- it will have to somehow produce an FsEnv in the instance
handler2 :: forall m. FsEnv -> ...

-- But it's tedious to write the instances. Imagine this times 30 for one monad, then times 4
-- for our various environments!
instance Cache MyApp where
  method = do
    env <- ask
    handler2 { cacheDir: env.cacheDir }
```

This is awkward because to implement a full monad with all the effects you've got to keep hopping to the effect modules to see what all the type class methods are, and then what handlers correspond to them, and then figure out how to make your giant environment suitable to cover all of the sub-environments.

**Improvements**

The only significant improvement I feel the MTL version brings to the table is that the typed cache is helped quite a bit by not having to have functor instances on the keys (no more `Functor2` class).

thomashoneyman:
I'm not going to continue any work on this, as I think it's much inferior to the Run version.

PR #576, opened 2023-01-10
Implement a minimal server

thomashoneyman:
This PR implements a server for the registry using HTTPure. You can run it with:

```console
spago run -p registry-app -m Registry.App.Server
```

You can then send requests to `localhost:8080/app/v1/<operation>` to try out different operations. Be careful! This is going to actually do what you ask for: it will publish, unpublish, or transfer in the production environment.

Here's what you should see on startup:

```
 ┌───────────────────────────────────────────┐
 │ Server now up on port 8080                │
 │                                           │
 │ To test, run:                             │
 │  > curl -v localhost:8080/api/v1/publish  │
 └───────────────────────────────────────────┘
```

You can then send requests. For example, to publish a package:

```console
curl --location --request POST 'localhost:8080/api/v1/publish' \
--header 'Content-Type: application/json' \
--data-raw '{
  "compiler": "0.15.6",
  "location": {
    "githubOwner": "purescript",
    "githubRepo": "purescript-prelude"
    },
  "name": "prelude",
  "ref": "v1.0.0"
}'
```

You should see in the terminal:

```console
[NOTIFY] Package source does not have a purs.json file. Creating one from your bower.json and/or spago.dhall files...
[NOTIFY] Verifying package...
[ERROR] You tried to upload a version that already exists: 1.0.0
Its metadata is:
```json
{
  "bytes": 13945,
  "hash": "sha256-SIdCAn2TXRpxz3dHSLeDTCzbjuMpO7xHsKhkCgRgdrw=",
  "publishedTime": "2016-06-01T00:46:17.000Z",
  "ref": "v1.0.0"
}
```

and you should receive the error via curl. Same goes for say a bad JSON payload:

```console
curl --location --request POST 'localhost:8080/api/v1/publish' \
--header 'Content-Type: application/json' \
--data-raw '{
  "compiler": "0.15.6",
  "location": {
    "githuOwner": "purescript",
    "githubRepo": "purescript-prelude"
    },
  "name": "prelude",
  "ref": "v1.0.0"
}'
```

yields the result in the console:

```
An error occurred while decoding a JSON value:
  Under 'Publish':
  At object key location:
  Under 'GitHub':
  At object key githubOwner:
  No value was found.
```

## Next Steps

This is meant as the first step in implementing a server just demonstrating that we can do it. But we should do more:

* We should probably implement a dry-run mode for the server so we can run tests against it.
* We need to specify a machine configuration for NixOS so we can deploy this to Digital Ocean
* We should probably store logs somewhere better than the local `logs` directory, and with better names, too.
* We should probably have requests return an identifier rather than an immediate result, and then allow users to look up the job by that identifier later on (ie. return `12345` if the json is valid, and have a `/status/12345` endpoint that has all its logs, and a `/poll/12345` that just returns 'pending' or the terminal result). Compiling a package, for example, can take a while and cause the request to drop.

That said, it _does_ work, and we can build from here.

### DIFF
--- a/app/src/App/Server.purs
+++ b/app/src/App/Server.purs
@@ -1,0 +1,178 @@
+module Registry.App.Server where
+
+import Registry.App.Prelude hiding ((/))
+
+import Data.Codec.Argonaut as CA
+import Data.Formatter.DateTime as Formatter.DateTime
+import Data.String as String
+import Effect.Aff as Aff
+import Effect.Class.Console as Console
+import HTTPure (Method(..))
+import HTTPure as HTTPure
+import Node.Path as Path
+import Node.Process as Process
+import Registry.App.API (Source(..))
+import Registry.App.API as API
+import Registry.App.Effect.Cache (CacheRef)
+import Registry.App.Effect.Cache as Cache
+import Registry.App.Effect.Env (PACCHETTIBOTTI_ENV)
+import Registry.App.Effect.Env as Env
+import Registry.App.Effect.Git (Debouncer, GIT)
+import Registry.App.Effect.Git as Git
+import Registry.App.Effect.GitHub (GITHUB)
+import Registry.App.Effect.GitHub as GitHub
+import Registry.App.Effect.Log (LOG, LogVerbosity(..))
+import Registry.App.Effect.Log as Log
+import Registry.App.Effect.Notify (NOTIFY)
+import Registry.App.Effect.Notify as Notify
+import Registry.App.Effect.Pursuit (PURSUIT)
+import Registry.App.Effect.Pursuit as Pursuit
+import Registry.App.Effect.Registry (REGISTRY)
+import Registry.App.Effect.Registry as Registry
+import Registry.App.Effect.Storage (STORAGE)
+import Registry.App.Effect.Storage as Storage
+import Registry.App.Legacy.Manifest (LEGACY_CACHE, _legacyCache)
+import Registry.Foreign.FSExtra as FS.Extra
+import Registry.Foreign.Octokit (GitHubToken, Octokit)
+import Registry.Foreign.Octokit as Octokit
+import Registry.Internal.Format as Internal.Format
+import Registry.Operation as Operation
+import Run (AFF, EFFECT, Run)
+import Run as Run
+import Run.Except (EXCEPT)
+import Run.Except as Except
+
+main :: Effect Unit
+main = do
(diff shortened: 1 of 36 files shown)


lib/test/Registry/Manifest.purs
```diff
@@ -8,7 +8,7 @@ import Node.Encoding (Encoding(..))
 import Node.FS.Aff as FS.Aff
 import Node.Path as Path
 import Registry.Manifest as Manifest
-import Test.Assert as Assert
+import Registry.Test.Assert as Assert
```
thomashoneyman, line 11: This is because `Test.Assert` is already a well-known module, and this rename avoids conflicts.

app/src/App/Server.purs
```diff
@@ -0,0 +1,178 @@
...
+  , githubCacheRef :: CacheRef
+  , legacyCacheRef :: CacheRef
+  , registryCacheRef :: CacheRef
+  , octokit :: Octokit
+  , vars :: ServerEnvVars
+  , debouncer :: Debouncer
+  }
+
+createServerEnv :: Aff ServerEnv
+createServerEnv = do
```
thomashoneyman, line 140: This is all information we want to persist from request to request.

f-f:
This is great! 👏 

As I mentioned in https://github.com/purescript/registry-dev/issues/578#issuecomment-1378734348 could we switch to `httpurple`? It uses `routing-duplex`, which I strongly prefer over the way `httpure` does it.

thomashoneyman:
Sure — in fact, I originally used `httpurple`, which is why I had to rename the `Test.Assert` module, but decided against it since none of us have direct maintainer's access and our routing is so simple (we aren't doing much with the extra features). But they are quite similar and I'm cool with either one.

thomashoneyman:
Just leaving this as a note for myself — [npmlock2nix](https://github.com/nix-community/npmlock2nix) now supports package-lock v2 formats (which we use), and [purescript2nix](https://github.com/cdepillabout/purescript2nix) looks like it will support our repository setup with spago.yaml, so I can now nixify the build way more easily.

sigma-andex:
> Sure — in fact, I originally used `httpurple`, which is why I had to rename the `Test.Assert` module, but decided against it since none of us have direct maintainer's access and our routing is so simple (we aren't doing much with the extra features). But they are quite similar and I'm cool with either one.

You mean direct access to httpurple? if so, happy to give you access

f-f (approved):

PR #580, opened 2023-01-11
Fix bounds derived from legacy package sets, bugfixes

### LINKED ISSUE #531: Remove legacy packages that do not solve

MonoidMusician:
Fixes #531. Supersedes #567.

Takes the union of bounds generated from package sets for a particular import. Fixes several bugs we found along the way, some of which were also encountered in the previous PR.

The two PRs generated from this PR (reuploads already went to S3):
- https://github.com/purescript/registry/pull/151
- https://github.com/purescript/registry-index/pull/5

### DIFF
--- a/app/src/App/Effect/Storage.purs
+++ b/app/src/App/Effect/Storage.purs
@@ -94,9 +94,8 @@ parsePackagePath input = do
         <$> PackageName.parse namePart
         <*> Version.parse versionPart
     parts -> Left
-      if Array.length parts > 2
-        then "Too many parts in path: " <> input
-        else "Too few parts in path: " <> input
+      if Array.length parts > 2 then "Too many parts in path: " <> input
+      else "Too few parts in path: " <> input
 
 formatPackageUrl :: PackageName -> Version -> Affjax.Node.URL
 formatPackageUrl name version = Constants.storageUrl <> "/" <> formatPackagePath name version
@@ -132,8 +131,9 @@ handleS3 env = Cache.interpret _storageCache (Cache.handleFs env.cache) <<< case
         Except.throw $ "Could not upload package " <> PackageName.print name <> " due to an error connecting to the storage backend."
       Just objects ->
         pure $ map _.key objects
-    pure $ Set.fromFoldable $
-      resources >>= \resource -> do
+    pure $ Set.fromFoldable
+      $ resources
+      >>= \resource -> do
         { name: parsedName, version } <- Array.fromFoldable $ parsePackagePath resource
         version <$ guardA (name == parsedName)
 
--- a/app/src/App/Effect/Storage.purs
+++ b/app/src/App/Effect/Storage.purs
@@ -181,7 +181,7 @@ handleS3 env = Cache.interpret _storageCache (Cache.handleFs env.cache) <<< case
       Except.throw $ "Could not delete " <> package <> " because it does not exist in the storage backend."
 
 -- | A storage effect that reads from the registry but does not write to it.
-handleReadOnly :: forall r a. FilePath -> Storage a -> Run (LOG + EXCEPT String + AFF + EFFECT + r) a
+handleReadOnly :: forall r a. FilePath -> Storage a -> Run (LOG + AFF + EFFECT + r) a
 handleReadOnly cache = Cache.interpret _storageCache (Cache.handleFs cache) <<< case _ of
   Upload name version _ reply -> do
     Log.warn $ "Requested upload of " <> formatPackageVersion name version <> " to url " <> formatPackageUrl name version <> " but this interpreter is read-only."
--- a/app/src/App/Effect/Storage.purs
+++ b/app/src/App/Effect/Storage.purs
@@ -23,6 +23,7 @@ import Affjax.StatusCode (StatusCode(..))
 import Data.Array as Array
 import Data.Exists as Exists
 import Data.HTTP.Method (Method(..))
+import Data.Set as Set
 import Data.String as String
 import Effect.Aff as Aff
 import Node.Buffer as Buffer
@@ -46,7 +47,7 @@ data Storage a
   = Upload PackageName Version FilePath (Either String Unit -> a)
   | Download PackageName Version FilePath (Either String Unit -> a)
   | Delete PackageName Version (Either String Unit -> a)
-  | Query PackageName (Either String (Array String) -> a)
+  | Query PackageName (Either String (Set Version) -> a)
 
 derive instance Functor Storage
 
@@ -72,7 +73,7 @@ interpret :: forall r a. (Storage ~> Run r) -> Run (STORAGE + r) a -> Run r a
 interpret handler = Run.interpret (Run.on _storage handler Run.send)
 
 -- | Query what tarballs exist for a package in the storage backend
-query :: forall r. PackageName -> Run (STORAGE + EXCEPT String + r) (Array String)
+query :: forall r. PackageName -> Run (STORAGE + EXCEPT String + r) (Set Version)
 query name = Except.rethrow =<< Run.lift _storage (Query name identity)
 
 formatPackagePath :: PackageName -> Version -> String
@@ -82,6 +83,20 @@ formatPackagePath name version = Array.fold
   , Version.print version
   , ".tar.gz"
   ]
+
+parsePackagePath :: String -> Either String { name :: PackageName, version :: Version }
+parsePackagePath input = do
+  filePath <- String.stripSuffix (String.Pattern ".tar.gz") input
+    # note ("Missing .tar.gz suffix: " <> input)
+  case String.split (String.Pattern "/") filePath of
+    [ namePart, versionPart ] ->
+      { name: _, version: _ }
+        <$> PackageName.parse namePart
+        <*> Version.parse versionPart
+    parts -> Left
+      if Array.length parts > 2
+        then "Too many parts in path: " <> input
+        else "Too few parts in path: " <> input
 
 formatPackageUrl :: PackageName -> Version -> Affjax.Node.URL
 formatPackageUrl name version = Constants.storageUrl <> "/" <> formatPackagePath name version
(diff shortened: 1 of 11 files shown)


lib/src/Range.purs
```diff
@@ -128,3 +129,7 @@ intersect (Range r1) (Range r2)
       { lhs: max r1.lhs r2.lhs
       , rhs: min r1.rhs r2.rhs
       }
+
+mk :: Version -> Version -> Maybe Range
```
thomashoneyman, line 136: @f-f you needed this for spago, right?

scripts/src/PackageDeleter.purs
```diff
@@ -138,35 +170,62 @@ main = launchAff_ do
...
   Log.info $ "Updating metadata for " <> formatted
   Registry.readMetadata name >>= case _ of
     Nothing -> Except.throw $ "Could not update metadata for " <> formatted <> " because no existing metadata was found."
-    Just (Metadata old) -> do
-      let new = Metadata $ old { published = Map.delete version old.published, unpublished = Map.delete version old.unpublished }
-      Registry.writeMetadata name new
+    Just (Metadata oldMetadata) -> do
+      let
+        publishment = case Map.lookup version oldMetadata.published, Map.lookup version oldMetadata.unpublished of
+          Just _, Just _ -> unsafeCrashWith $ "Package version was both published and unpublished: " <> formatted
```
thomashoneyman: Could we do `Except.throw` here?

scripts/src/Solver.purs
```diff
@@ -0,0 +1,218 @@
...
+  do
+    unionWithMapThese \_ -> filtered <<< these
+      do map This
+      do map That
+      do
+        unionWithMapThese \_ -> these
+          do Just <<< This
+          do Just <<< That
+          \l r ->
+            if l == r then Nothing else Just (Both l r)
```
thomashoneyman: Well, this is something new 😆

scripts/src/Solver.purs
```diff
@@ -0,0 +1,218 @@
...
+        packageversions <- liftAff (readJsonFile deletePackagesCodec path) >>= case _ of
+          Left err -> Aff.log err *> liftEffect (Process.exit 1)
+          Right values -> pure values
+        pure \registry -> do
+          forWithIndex_ packageversions \package versionsofpackage -> do
+            for_ versionsofpackage \version -> do
+              let
+                deps =
+                  case Map.lookup package registry of
+                    Nothing -> unsafeCrashWith $ "Missing package: " <> PackageName.print package
```
thomashoneyman: Same here wrt `Except.throw`, since we have it available via `PublishEffects`

scripts/src/Solver.purs
```diff
@@ -0,0 +1,218 @@
+module Registry.Scripts.Solver where
```
thomashoneyman, line 10: Do you mind adding a module header with a comment describing what this script does and why it exists? Do you feel there is any point at which we should remove this script in the future?

lib/src/Range.purs
f-f: Not really, see #582

scripts/src/Solver.purs
MonoidMusician: Hey, I copied `unionWithMapThese` from my project that Nate used to stress test purs-tidy, I needed to continue in the tradition ;)
https://github.com/MonoidMusician/dhall-purescript/blob/469c3e10d51a8afb90f2e231cf3e6101f50814eb/src/Dhall/Map.purs#L52-L65

scripts/src/Solver.purs
MonoidMusician: I think we should keep it, it seems like it will be useful to run the solver and see what it resolves to or what the failures are. Especially because we can modify the registry index and test quick fixes locally, as opposed to a spago workflow which wants a clean registry index. And because I'm proud of my solver :D

scripts/src/Solver.purs
thomashoneyman: I'm referring to the `do`s :laughing:

thomashoneyman (dismissed):

scripts/src/Solver.purs
f-f: Yea, I don't understand this code, could we rewrite it to be more explicit? The `do`s are probably doing something here, but also make it very confusing to read

MonoidMusician: The `do`s are basically bullet points, I'm using them to separate multiple arguments using indentation instead of a whole mess of parentheses.

MonoidMusician: Anyways, I'm actually removing this since it isn't really relevant any more and was relying on a temporary checkout of the registry.

app/src/App/Effect/Storage.purs
```diff
@@ -45,6 +46,7 @@ data Storage a
   = Upload PackageName Version FilePath (Either String Unit -> a)
   | Download PackageName Version FilePath (Either String Unit -> a)
   | Delete PackageName Version (Either String Unit -> a)
+  | Query PackageName (Either String (Array String) -> a)
```
thomashoneyman: Since I believe these are package paths, of the form `"package/x.y.z.tar.gz"`, I feel like `Query` should return an `Array Version`

app/src/App/Effect/Storage.purs
```diff
@@ -183,6 +198,9 @@ handleS3 env = Cache.interpret _storageCache (Cache.handleFs env.cache) <<< case
 -- | A storage effect that reads from the registry but does not write to it.
 handleReadOnly :: forall r a. FilePath -> Storage a -> Run (LOG + AFF + EFFECT + r) a
 handleReadOnly cache = Cache.interpret _storageCache (Cache.handleFs cache) <<< case _ of
+  Query _ reply -> do
+    pure $ reply $ Left "Cannot query in read-only mode."
```
thomashoneyman, line 222: You should be able to query in read-only mode, as it does not modify the registry. To share the implementation you can do the same thing done with `downloadS3`.

scripts/src/VerifyIntegrity.purs
```diff
@@ -0,0 +1,187 @@
...
+  s3 <- lift2 { key: _, secret: _ } (Env.lookupRequired Env.spacesKey) (Env.lookupRequired Env.spacesSecret)
+
+  -- Git
+  debouncer <- Git.newDebouncer
+  let
+    gitEnv :: GitEnv
+    gitEnv =
+      { write: ReadOnly
+      , pull: Autostash
+      , repos: Git.defaultRepos { manifestIndex = Git.defaultRepos.manifestIndex { owner = "monoidmusician" }, registry = Git.defaultRepos.registry { owner = "monoidmusician" } }
```
thomashoneyman: I don't think we should hardcode these values. I think we should use default repos, and either take the overrides via the command line, or expect that if you are working on the scripts locally then you override it yourself.

scripts/src/VerifyIntegrity.purs
```diff
@@ -0,0 +1,187 @@
...
+        >>> Storage.interpret (Storage.handleS3 { s3, cache })
+        >>> Git.interpret (Git.handle gitEnv)
+        >>> GitHub.interpret (GitHub.handle { octokit, cache, ref: githubCacheRef })
+        >>> Log.interpret (\log -> Log.handleTerminal Normal log *> Log.handleFs Verbose logPath log)
+        >>> Run.runBaseAff'
+
+  interpret do
+    let
+      sorry e = Log.error e *> liftEffect (Process.exit 1)
+    allMetadata <- Except.runExcept Registry.readAllMetadata >>= either (map absurd <<< sorry) pure
```
thomashoneyman: Why not just add an `Except` interpreter that exits the process instead of this `sorry`?

```purs
>>> Except.catch (\error -> Run.liftEffect (Console.log error *> Process.exit 1))
```

For example, since you're not doing anything with the exception when reading metadata then the two places where you use `sorry` would just become:

```purs
interpret do
  allMetadata <- Registry.readAllMetadata
  allManifests <- Registry.readAllManifests
  ...
```

scripts/src/VerifyIntegrity.purs
MonoidMusician: yeah I removed that already

scripts/src/Solver.purs
f-f: Great!

app/src/App/Effect/Storage.purs
MonoidMusician: It doesn't have tokens to connect to s3 in read-only mode, unless you know of a different way.

app/src/App/Effect/Storage.purs
MonoidMusician: That's correct. I didn't feel like parsing it tbh but I can if you want.

scripts/src/VerifyIntegrity.purs
MonoidMusician: You mean adding that globally? Ideally it wouldn't stop on each package that fails so you could see all the failures.

scripts/src/VerifyIntegrity.purs
```diff
@@ -0,0 +1,187 @@
...
+printDblDiff f names { left, right } = intercalate ", " $ join
+  [ if Set.isEmpty right then mempty
+    else
+      [ names.left <> " missing " <> setOf f right ]
+  , if Set.isEmpty left then mempty
+    else
+      [ names.right <> " missing " <> setOf f left ]
+  ]
+
+verifyPackage :: forall r. Map PackageName Metadata -> ManifestIndex -> Set String -> PackageName -> Run (EXCEPT String + LOG + r) Unit
```
thomashoneyman: Could you add a comment about the rules for verification, e.g. the metadata is the source of truth; there should be zero versions present in S3 / registry-index that are not also present in the metadata's 'published' field, and every version present in the metadata's 'published' field should also be present in s3 / the registry-index?

scripts/src/VerifyIntegrity.purs
thomashoneyman: You're saying you want the process to exit as soon as you encounter one of these exceptions? Or that you don't want to see any exceptions until the end?

app/src/App/Effect/Storage.purs
thomashoneyman: Oh, yea, we have to be connected to S3 to query objects. That's...kind of lame. Well, querying certainly seems like something we would want to have available in a read-only mode, but perhaps you can just add a TODO comment here.

app/src/App/Effect/Storage.purs
thomashoneyman: I would prefer that, thank you!

scripts/src/VerifyIntegrity.purs
MonoidMusician: Oh, I misunderstood your suggestion. Makes sense!

MonoidMusician:
Okie, I think I addressed all your feedback. Thanks for the review, the help in writing it, and especially the new effects! I'm liking how the effects work out, I think it's pretty clean.

thomashoneyman (approved):

PR #582, opened 2023-01-11
A few fixes to make Spago work

f-f:
Mostly shuffling some things back to the `lib`.

### DIFF
--- a/lib/src/Constants.purs
+++ b/lib/src/Constants.purs
@@ -27,3 +27,54 @@ storageUrl = "https://packages.registry.purescript.org"
 -- | The URL of the package operation API
 apiUrl :: String
 apiUrl = "https://registry.purescript.org/api"
+
+ignoredDirectories :: Array FilePath
+ignoredDirectories =
+  [ ".psci"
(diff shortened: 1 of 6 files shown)


lib/src/Constants.purs
```diff
@@ -27,3 +27,54 @@ storageUrl = "https://packages.registry.purescript.org"
...
+  ]
+
+ignoredFiles :: Array FilePath
+ignoredFiles =
+  [ "package-lock.json"
+  , "yarn.lock"
+  , "pnpm-lock.yaml"
+  ]
+
+ignoredGlobs :: Array String
```
thomashoneyman, line 53: Do you think these should be `Array String.Pattern`? Or do you think the "globs" in the name does the job?

thomashoneyman (approved):
I leave the globs question up to you, but otherwise :+1:

lib/src/Constants.purs
f-f: I'm not sure about this yet, let's circle back to this once we have the logic implemented in Spago's publishing pipeline!

PR #583, opened 2023-01-13
Upgrade spago to latest

f-f:


thomashoneyman:
That’s an interesting error — I wonder what changed? If you’re using the version parser you’ll want to switch to lenient parsing for things that start with a ‘v’ prefix.

MonoidMusician:
Hm LegacyVersion is in app, though, and spago isn't depending on app yet (just lib and foreign).

MonoidMusician:
PR: https://github.com/purescript/spago/pull/932

f-f:
Yeah, this broke because we were using lenient version parsing everywhere. Spaghetto has no tests at the moment, so a few things might still break (we didn't get to exercise this code path on our work repo, as we bypass all the Node stuff)

f-f:
This should do a little better now?

thomashoneyman (approved):

PR #586, opened 2023-01-22
Tighten up the dependency list: remove unused and include undeclared transitive packages

f-f:
Spago Next now has an improved version of what used to be the "unused and transitive dependencies check", that allows for tightening up the dependencies declared in the build files.

It can be run with the `--pedantic-packages` flag, sample output before this PR:
image

### DIFF
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -59,7 +59,6 @@ package:
     - registry-lib
     - run
     - safe-coerce
-    - spec
     - strings
     - sunde
     - these
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -7,32 +7,31 @@ package:
     license: BSD-3-Clause
   dependencies:
     - aff
-    - aff-promise
     - affjax
     - affjax-node
+    - ansi
     - argonaut-core
     - arrays
     - b64
     - bifunctors
+    - codec-argonaut
     - console
+    - const
     - control
-    - convertable-options
     - datetime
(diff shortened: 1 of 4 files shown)


app/spago.yaml
```diff
@@ -47,23 +46,29 @@ package:
...
     - prelude
-    - record
+    - profunctor
     - refs
-    - registry-lib
     - registry-foreign
+    - registry-lib
+    - run
     - safe-coerce
+    - spec
```
thomashoneyman: This should be a test dependency

foreign/spago.yaml
```diff
@@ -3,14 +3,43 @@ package:
...
+    - node-fs
+    - node-fs-aff
+    - node-path
     - node-process
+    - nullable
+    - ordered-collections
+    - prelude
+    - profunctor
     - registry-lib
+    - spec
```
thomashoneyman: This is a test dependency and is already listed there — is it possible to detect test dependencies separately? As we import a few packages only for test iirc

f-f: Ah yes, I'm working on that at the moment, will push an update once I'm through

thomashoneyman (approved):

thomashoneyman:
Should we also bump spago in the flake.nix?

f-f:
@thomashoneyman I'll do a release in a few days, will bump it then!

PR #589, opened 2023-02-09
Cache publish failures (solver errors)

MonoidMusician:
So that it doesn't cause the importer to time out after 6 hours.

I just added `scratch/.cache/PublishFailureCache__*`, I think that makes the most sense given how we're handling the other failures?

We can't quite roll it into the existing caches of version failures since those are on raw versions and this is on proper versions.

Once this is merged, we will rerun the package set updater to pick up anything that was missed.

But first we should probably figure out the mysterious phantom dependencies that are in the index ...

### DIFF
--- a/scripts/src/LegacyImporter.purs
+++ b/scripts/src/LegacyImporter.purs
@@ -165,12 +165,13 @@ main = launchAff_ do
     # runAppEffects
     # Cache.interpret Legacy.Manifest._legacyCache (Cache.handleMemoryFs { cache, ref: legacyCacheRef })
     # Cache.interpret _importCache (Cache.handleMemoryFs { cache, ref: importCacheRef })
+    # Cache.interpret _publishFailureCache (Cache.handleFs cache)
     # Notify.interpret Notify.handleLog
     # Except.catch (\msg -> Log.error msg *> Run.liftEffect (Process.exit 1))
     # Log.interpret (\log -> Log.handleTerminal Normal log *> Log.handleFs Verbose logPath log)
     # Run.runBaseAff'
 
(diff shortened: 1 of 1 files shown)


scripts/src/LegacyImporter.purs
```diff
@@ -846,3 +853,22 @@ type IMPORT_CACHE r = (importCache :: Cache ImportCache | r)
 
 _importCache :: Proxy "importCache"
 _importCache = Proxy
+
+-- | A key type for the storage cache. Only supports packages identified by
+-- | their name and version.
+data PublishFailureCache :: (Type -> Type -> Type) -> Type -> Type
+data PublishFailureCache c a = PublishFailure PackageName Version (c String a)
```
thomashoneyman: It looks like we're only going to use this in the importer, so can we add `PublishFailure` as a key to the existing import cache instead of adding a new cache?

```purs
data ImportCache :: (Type -> Type -> Type) -> Type -> Type
data ImportCache c a 
  = ImportManifest PackageName RawVersion (c (Either VersionValidationError Manifest) a)
  | PublishFailure PackageName Version (c String a)

thomashoneyman:
> We can't quite roll it into the existing caches of version failures since those are on raw versions and this is on proper versions.

We do have a way of re-associating versions back to their raw versions:

https://github.com/purescript/registry-dev/blob/ac1c055ff67b050cd8a0db66911b3470b9f59adc/scripts/src/LegacyImporter.purs#L356-L361

That said, we may want this to be in the temporary cache anyways because it's still somewhat in flux (the phantom dependencies).

scripts/src/LegacyImporter.purs
MonoidMusician: Oh right, that's what ADTs are for 😄 Will do that

thomashoneyman (approved):

PR #593, opened 2023-03-11
Read spago.yaml files and use them to create Manifests

f-f:


f-f:
Ha - CI fails because we run `spago install`, and that can't find all the modules that `spago-core` imports from `registry-lib`.
This is because `install` will not include packages from the workspace but only external dependencies, but `spago-core` (an external dependency that depends on `registry-lib`, which is a workspace package) needs them for building.
This is not an issue when calling `spago build` because the compiler just gets all the sources and can find everything.
@thomashoneyman are we good to replace `spago install` with `spago build`?

thomashoneyman:
Yea that’s fine

app/src/App/API.purs
```diff
@@ -371,31 +374,58 @@ publish source payload = do
...
+          Left err' -> Except.throw $ String.joinWith "\n"
+            [ "Could not publish your package - a spago.yaml was present, but it was not possible to read it:"
+            , err'
+            ]
+          Right { yaml: config } -> do
+            -- Once we have the config we are still not entirely sure it fits into a Manifest
+            -- E.g. need to make sure all the ranges are present
+            case spagoToManifest config of
+              Left err -> Except.throw $ String.joinWith "\n"
+                [ "Could not publish your package - there was an error while converting your spago.yaml into a Registry Manifest:"
```
thomashoneyman: Elsewhere we've been referring to this as a "purs.json manifest" rather than a Registry Manifest; could we do the same here (and on line 405)?

app/src/App/API.purs
```diff
@@ -981,3 +1011,28 @@ getPacchettiBotti = do
...
+  location <- note "Did not find a `location` field in the publish config" publishConfig.location
+  let
+    checkRange :: Tuple PackageName (Maybe Range) -> Either PackageName (Tuple PackageName Range)
+    checkRange (Tuple packageName maybeRange) = case maybeRange of
+      Nothing -> Left packageName
+      Just r -> Right (Tuple packageName r)
+  let { fail: failedPackages, success } = partitionEithers $ map checkRange (Map.toUnfoldable deps :: Array _)
+  dependencies <- case failedPackages of
+    [] -> Right (Map.fromFoldable success)
+    errs -> Left $ "The following packages did not have their ranges specified: " <> printJson (CA.array PackageName.codec) errs
```
thomashoneyman: Do you want this to print as an array `[ "aff", "affjax", "algebra" ]` or did you want it to just be comma-separated, like `...ranges specified: aff, affjax, algebra`?

app/src/App/API.purs
```diff
@@ -981,3 +1011,28 @@ getPacchettiBotti = do
...
+  pure $ Manifest
+    { version
+    , license
+    , name
+    , location
+    , description
+    , dependencies
+    , owners: Nothing -- TODO Spago still needs to add this to its config
+    , files: Nothing -- TODO Spago still needs to add this to its config
```
thomashoneyman, line 1037: Can we make an issue about this on the registry-dev repo to track it?

thomashoneyman (commented):
Just a few comments here and there, and then I think this looks good. Glad to have explicit spago.yaml support, even if the owners / files keys aren't yet supported!

app/src/App/API.purs
f-f: The latter, but I forgot to come back to it, good catch 😄

app/src/App/API.purs
f-f: #594

thomashoneyman (approved):

PR #597, opened 2023-04-13
Migrate SSH to ssh2 library

### LINKED ISSUE #591: Do not depend on the system's ssh package for crypto
There are [some operations](https://github.com/purescript/registry-dev/blob/master/SPEC.md#52-authentication) for which we require strong cryptographic authentication.
We do this by piggybacking on the native ssh executable for the cryptographic bits. This seems to be quite fragile however - different versions of ssh on different operating systems seem to do different things, so we should not rely on the native binary.
Fortunately there is [a pure-js package](https://www.npmjs.com/package/ssh2) that implements ssh's protocols, and that includes [reading keys, signing and verifying payloads](https://www.npmjs.com/package/ssh2#utilities), which is all we need here.
We should switch to this package both in the CI here and in the package managers talking to it. The downside is that one would have to use node to perform these operations, but I'd argue that this is a better situation than relying on whatever version one has locally.

thomashoneyman:
This PR migrates the registry away from `ssh-keygen` and over to the [ssh2](https://github.com/mscdex/ssh2) node library as per #591. This is because mismatched ssh-keygen versions have been giving us a lot of headaches in the transfer operation and it would be nicer to be able to sign and verify data using a specific version of the ssh2 library via Spago (and anyone else who wants to sign data using the registry library).

Unfortunately, I'm not exactly sure how to handle translating the signature buffer that we receive from ssh2's `sign` method into a string suitable for use in a JSON payload and back; a first cut with `Buffer.toString` using a UTF8 encoding was no good, and I don't have time to dig in to what encoding should be used just yet.

For that reason you'll see the tests are failing. I'll need to come back and fix that part but everything else should be OK.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -433,7 +433,7 @@ Being a package "owner" grants the ability to take sensitive actions on behalf o
 The registry relies on SSH to authenticate package owners. An `Owner` is a public SSH key, and this SSH key can be used to sign data using OpenSSH versions 8.0 and above. The registry in turn verifies the signed data using the `Owner` and OpenSSH. To submit an authenticated operation, you are required to take the following steps:
 
 1. Construct the JSON payload for the operation
-2. Sign the JSON payload using an SSH key listed in the package's `owners` metadata field.
+2. Sign the JSON payload using an SSH key listed in the package's `owners` metadata field, where the signature is hex-encoded.
 3. Construct the JSON payload for an authenticated operation using the JSON payload from step (1), the signature from step (2), and the email associated with the SSH key used to create the SSH signature.
 
 For example, here's a stringified JSON payload for an `Unpublish` operation unpublishing `prelude@1.0.1` because credentials were accidentally committed:
@@ -442,29 +442,10 @@ For example, here's a stringified JSON payload for an `Unpublish` operation unpu
 "{ \"name\": \"prelude\", \"version\": \"1.0.1\", \"reason\": \"Accidentally committed credentials\" }"
 ```
 
-We can sign this JSON using `ssh-keygen` as described in the post [It's Now Possible to Sign Arbitrary Data With Your SSH Keys](https://www.agwa.name/blog/post/ssh_signatures):
-
-```sh
-ssh-keygen -Y sign -f [ssh_key] -n file [file_to_sign]
-```
-
-where `[ssh_key]` is the path to your private key (such as `~/.ssh/id_ed25519`) and `[file_to_sign]` is the path to the file to be signed (such as `unpublish.json`). This will write the SSH signature to a new file named `[file_to_sign].sig` (for example, `unpublish.sig`), and it looks like this:
-
-```sig
------BEGIN SSH SIGNATURE-----
-U1NIU0lHAAAAAQAAADMAAAALc3NoLWVkMjU1MTkAAAAg2rirQQddpzEzOZwbtM0LUMmlLG
-krl2EkDq4CVn/Hw7sAAAAEZmlsZQAAAAAAAAAGc2hhNTEyAAAAUwAAAAtzc2gtZWQyNTUx
-OQAAAEDyjWPjmOdG8HJ8gh1CbM8WDDWoGfm+TTd8Qa8eua9Bt5Cc+43S24i/JqVWmk98qV
-YXoQmOYL4bY8t/q7cSNeMH
------END SSH SIGNATURE-----
-```
-
-You can also specify `-` for the filename, in which case the file to sign is read from stdin and the signature is written to stdout.
-
-You must then assemble this information into a JSON `object` with three fields:
+Then, assemble this information into a JSON `object` with three fields:
 
 - `payload`: The JSON string representing the `Unpublish` or `Transfer` operation
-- `signature`: An array of strings representing each line of the SSH signature
+- `signature`: A hex-encoded SSH signature
 - `email`: The email address associated with the public key used to sign the payload (this should be in the `owners` metadata field).
 
 For example, in JSON:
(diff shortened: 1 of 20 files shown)


app/test/App/Auth.purs
```diff
@@ -4,6 +4,7 @@ import Registry.App.Prelude
```
thomashoneyman: This module somewhat duplicates the basic SSH tests in the `Test.Registry.SSH` module, but there are some differences — the `verifyPayload` function, for example, does more than your basic sign / verify.

scripts/src/PackageTransferrer.purs
```diff
@@ -122,11 +123,13 @@ transferPackage rawPackageName newLocation = do
...
-  { publicKey, privateKey } <- Env.askPacchettiBotti
+  { privateKey } <- Env.askPacchettiBotti
 
-  signature <- Run.liftAff (Auth.signPayload { publicKey, privateKey, rawPayload }) >>= case _ of
+  signature <- case Auth.signPayload { privateKey, rawPayload } of
     Left _ -> Except.throw "Error signing transfer."
-    Right signature -> pure signature
+    Right signature -> do
+      asString <- Run.liftEffect $ Buffer.toString UTF8 signature
+      pure $ String.split (String.Pattern "\n") asString
```
thomashoneyman: I can just about guarantee this won't work. I'll need to fix this when we figure out the proper way to manage signature buffers.

f-f: A UTF8 string is probably not a good idea, but `base64` encoding should be fine? It would guarantee a big long string of boring characters that are JSON friendly.

thomashoneyman: While looking through the source of the `ssh2` library I see that their `sign` and `verify` methods are thin wrappers over Node's crypto library, which typically uses hex encodings on the output to keep it concise & portable. I've followed their lead in the latest commits and things are working as expected.

lib/src/Operation.purs
```diff
@@ -92,15 +93,15 @@ publishCodec = CA.Record.object "Publish"
 type AuthenticatedData =
   { payload :: AuthenticatedPackageOperation
   , rawPayload :: String
-  , signature :: Array String
+  , signature :: Signature
   , email :: String
```
thomashoneyman: Technically this does not need to be an email address and it can simply be a comment of any kind. Should we change this field name to be `comment` so it can more easily be used as a way to distinguish one owner from another? For example, I can just type `{ email: "thomas key" }` and it will be fine.

thomashoneyman: It's even explicitly called out as a comment regularly in SSH documentation:
https://man.openbsd.org/OpenBSD-current/man8/sshd.8#AUTHORIZED_KEYS_FILE_FORMAT

f-f: I wonder now if we need this field at all?

SPEC.md
```diff
@@ -442,54 +442,28 @@ For example, here's a stringified JSON payload for an `Unpublish` operation unpu
...
-```
-
-You can also specify `-` for the filename, in which case the file to sign is read from stdin and the signature is written to stdout.
-
-You must then assemble this information into a JSON `object` with three fields:
+Then, assemble this information into a JSON `object` with three fields:
 
 - `payload`: The JSON string representing the `Unpublish` or `Transfer` operation
-- `signature`: An array of strings representing each line of the SSH signature
+- `signature`: A hex-encoded SSH signature
```
f-f: The previous instructions detailed how to produce such a signature, I guess the new instructions would include Spago, let's leave a pointer (like "TODO" or something) so we remember to get back to this

SPEC.md
```diff
@@ -442,54 +442,28 @@ For example, here's a stringified JSON payload for an `Unpublish` operation unpu
...
 
 **Verifying SSH Signatures**
 
-When the registry receives an authenticated operation, it takes the following steps:
+When the registry receives an authenticated operation it takes the following steps:
 
-1. The registry will create an [`allowed_signers`](https://man.openbsd.org/ssh-keygen#ALLOWED_SIGNERS) file, which will list all package owners from the package's metadata as well as the @pacchettibotti SSH keys.
-2. The registry will use `ssh-keygen` to verify the provided signature, the provided payload, and the provided email address. If the signature is valid, then `ssh-keygen` will exit with a 0 status and a message beginning with "Good 'file' signature for...". Otherwise, the command will exit with an error.
+1. The registry will retrieve all package owners from the package's metadata
+2. The registry will use each key listed in the package owners to attempt to verify the signature on the authenticated operation.
```
f-f: The previous docs mention that pacchettibotti can also be authenticated as legit, and we should add the trustees too.

lib/src/Operation.purs
thomashoneyman: I think we have to put something there, but it can be anything and it doesn’t affect the key. So the user doesn’t have to supply it. We could make it optional in the manifest (if having a comment would be useful) or remove it altogether.

SPEC.md
thomashoneyman: Yes, there’s really no instruction besides “use spago” or perhaps “use the registry functions”

SPEC.md
thomashoneyman: Technically it’s only pacchettibotti. The trustees use GitHub to request a pacchettibotti signature.

lib/src/Operation.purs
f-f: Isn't this in the Operation though? (i.e. not in the manifest)

thomashoneyman: Oh, psh, I was thinking of the Owner type this whole time — it isn’t necessary there either (but might be useful).

For the authenticated data the only reason we would need it is to detect when a trustee is trying to act as pacchettibotti — we look at the email field.

But we could also say have a convention where if it’s an empty signature we check if you’re a trustee. They’re both just conventions to say “hey a trustee wants to take an action here.”

I think we should remove it from the authenticated data as you suggest since it doesn’t serve any other purpose.

thomashoneyman:
@f-f the latest commits address feedback in these steps:

1. Explicitly mention in the SPEC that pacchettibotti is always considered an owner and trustees can act as pacchettibotti.
2. Direct the user to use their package manager or the registry library functions to sign payloads
2. Remove the 'email' field from the authenticated data type
3. Rename the 'email' field to 'comment' in the owner data type and make it optional

SPEC.md
```diff
@@ -143,17 +143,26 @@ Or, using a monorepo:
...
 
-Note that the email address does not have to be able to send or receive email, but it must be the email address that was associated with the private key. A JSON example:
+Note that the comment is optional. A JSON example:
 
 ```jsonc
 {
   "keytype": "ssh-ed25519",
   "public": "ABCD3FGzaC1lZDI1NTE5AAAAINq4q0EHXacxMzmcG7TNC1DJpSxpK5dhJA6uAlZ",
-  "email": "john@abc.com"
+  "comment": "john@abc"
```
f-f: I am now wondering if this field in the owner is needed at all - is it just for distinguishing the keys?
If so, then maybe `id` could be a more fitting name?

app/src/App/Main.purs
```diff
@@ -272,36 +273,37 @@ decodeIssueEvent json = lmap CA.printJsonDecodeError do
...
+-- pacchettibotti will re-sign it and add itself as an owner before continuing
+-- with the authenticated operation.
 signPacchettiBottiIfTrustee
   :: forall r
    . AuthenticatedData
-  -> Run (GITHUB + PACCHETTIBOTTI_ENV + GITHUB_EVENT_ENV + EXCEPT String + AFF + r) AuthenticatedData
+  -> Run (GITHUB + PACCHETTIBOTTI_ENV + GITHUB_EVENT_ENV + EXCEPT String + AFF + EFFECT + r) AuthenticatedData
 signPacchettiBottiIfTrustee auth = do
-  if auth.email /= pacchettibottiEmail then
+  if auth.signature /= Signature "" then
```
f-f: Why run this check only with an empty comment? We could run it all the time.

SPEC.md
thomashoneyman: Just for distinguishing keys, that's right. We could call it `id` instead. I went with `comment` since that matches the SSH terminology but it doesn't really matter — we don't need the field at all and I've made it optional; it could be nice to keep in there optionally though, so you can distinguish keys.

app/src/App/Main.purs
thomashoneyman: Yes, I suppose we could always run the check to see if a trustee made the comment and always sign as pacchettibotti if so. Presumably then if we get an error looking up the username in the packaging team, or the username isn't part of the packaging team, then we just go ahead and try to process the operation with no re-signing?

f-f: Yep that sounds good

SPEC.md
thomashoneyman: I've renamed the optional field to "id"

app/src/App/Main.purs
thomashoneyman: I've made it so that this check is always run, and on failure we just send the request through un-altered.

f-f (approved):
Thanks!

PR #598, opened 2023-04-16
Exclude Main and Test.Main as module names

### LINKED ISSUE #566: Disallow package publication/updates if modules named `Main` or `Test.Main` are included
See https://github.com/i-am-the-slime/purescript-visx/issues/1, which hindered an update to Try PureScript.

`Main` is typically the name of a module one uses for their application's entry point. Similarly, `Test.Main` is usually the entry point for running tests. If a package includes a module for either of those names, it'll prevent all users who use that library from using these commonly-used module names.

Thus, I propose disallowing package publication and updates if they include such modules in their source globs.

thomashoneyman:
Fixes #566 by disallowing `Main` and `Test.Main` module names in library code due to their high chance of colliding with user module names. These really ought not be published to libraries and we now explicitly disallow it.

This PR uses `parsePartialModule` from `lanugage-cst-parser` as suggested by @natefaubion to reliably parse just the module header of a file:
https://github.com/natefaubion/purescript-language-cst-parser/blob/7a0a4484f8abd91d688d297461209a75667e0da6/src/PureScript/CST.purs#L83-L84

I've adjusted our `containsPursFile` check to instead be `validatePursModules`; instead of walking their `src` directory to ensure there is at least one PureScript file, we now parse the module header of each module in their `src` directory and fail if either a) there are no modules or b) a module uses a reserved name.

### DIFF
--- a/app/test/App/API.purs
+++ b/app/test/App/API.purs
@@ -138,7 +138,6 @@ copySourceFiles = Spec.hoistSpec identity (\_ -> Assert.Run.runTest) $ Spec.befo
 
     for_ acceptedPaths \path -> do
       paths.succeeded `Assert.Run.shouldContain` path
-
   where
   runBefore :: forall r. Run (EFFECT + r) _
   runBefore = do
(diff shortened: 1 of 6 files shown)


app/test/App/API.purs
```diff
@@ -153,6 +152,6 @@ copySourceFiles = Spec.hoistSpec identity (\_ -> Assert.Run.runTest) $ Spec.befo
       writeDirectories = traverse_ (FS.Extra.ensureDirectory <<< inTmp)
 
       writeFiles :: Array FilePath -> _
-      writeFiles = Run.liftAff <<< traverse_ (\path -> FS.Aff.writeTextFile UTF8 (inTmp path) "<test>")
+      writeFiles = Run.liftAff <<< traverse_ (\path -> FS.Aff.writeTextFile UTF8 (inTmp path) "module Module where")
```
thomashoneyman, line 155: New check caught this! Can't be out here with modules containing only the string `<test>`, they have to have a valid module name (which isn't Test.Main or Main).

f-f (approved):
Lovely!

Issue #249, opened 2021-10-11
Handle packages with combined LICENSE files


thomashoneyman:
A number of packages in the PureScript ecosystem have LICENSE files containing multiple licenses. This is for various reasons, but often this happens when a library is ported from another language (ie. Haskell) and needs to retain license contents from that library. For some examples, see:

- https://github.com/slamdata/purescript-markdown/blob/master/LICENSE
- https://github.com/purescript-contrib/purescript-parsing/blob/main/LICENSE

These license files are parsed by `licensee` as NOASSERTION, meaning it can't figure out what the license type is, and eventually they fail the import pipeline.

In the case of these two packages, they're combinations of MIT and BSD-2-Clause, which means we can manually fix them up when importing to the registry. However, a lot of packages are failing this way (~360 versions across ~50 packages), so it may be unrealistic that we fix them manually.

Still, these are packages that ought to be in the registry and which _do_ have valid license files, despite the errors resulting from our attempt to parse them.

---

One possible solution: we could some sort of "resolutions" step, where if we find a license parsed as NOASSERTION, we:

1. Print the LICENSE contents to the console
2. Ask the user to review the LICENSE and manually choose licenses (providing an empty array if it's not a valid license file)
3. Save the user's manual feedback to a `license-resolutions.json` file
4. On subsequent runs, skip detecting from the LICENSE file if the package version is present in the license-resolutions file

This could be a reasonably quick way to run through the ~360 versions that are failing for this reason.

There may also be a more automated solution, but we start getting into diminishing returns vs. manual processing in these last few packages.

mikesol:
Would it make sense for license to ne a nonempty array? That punts but could resolve this issue.

thomashoneyman:
@mikesol We do parse licenses into a `NonEmptyArray`, and we combine the result into a valid SPDX-recognized license. A valid SPDX representation of multiple licenses combines them together using AND or OR, for example "MIT OR BSD3-Clause".

In this case, the issue is that we're not able to automatically parse the LICENSE file using the `licensee` CLI tool because it contains multiple licenses, and so we're going to have to parse it another way -- either writing some new code, or manually reading the LICENSE file and providing a resolution. Either way we'll produce an array of licenses, which will later be combined.

mikesol:
Reading over the [SPDX matching guidelines](https://spdx.dev/license-list/matching-guidelines/), here is the relevant bit:

[...]

PR #600, opened 2023-04-18
Render solver errors with backticks to prevent accidental quoting

thomashoneyman:
Wraps the solver error output in backticks to prevent accidental quoting when showing an error range that begins with `>`. See, for example:

https://github.com/purescript/registry/issues/249#issuecomment-1513535532

@f-f As a brief aside, this is something to keep in mind when recording logs. When notifying via GitHub we want markdown formatting; when storing a log for terminal output we don't.

### DIFF
--- a/app/src/App/API.purs
+++ b/app/src/App/API.purs
@@ -613,11 +613,12 @@ verifyResolutions manifest resolutions = do
         let
           printedError = String.joinWith "\n"
             [ "Could not produce valid dependencies for manifest."
-            , ""
+            , "```"
             , errors # foldMapWithIndex \index error -> String.joinWith "\n"
                 [ "[Error " <> show (index + 1) <> "]"
                 , Solver.printSolverError error
                 ]
+            , "```"
             ]
         Except.throw printedError
       Right solved -> pure solved


f-f:
Good point on the markdown formatting - then maybe we should not be using strings, but something more structured?
This would be akin to what Dodo does, allowing you to use different formatters.

thomashoneyman:
Incidentally, the `Log` and `Notify` effects do use Dodo and can accept a `Doc` directly.

f-f:
Excellent - I now wonder if we can express this Markdown formatting change with a custom formatter

thomashoneyman:
I'm open to exploring that. For the time being, though, I'd like to get this in — it's not a new behavior, we have backticks written into our string logs all over the place — since it's making the solver errors hard to understand right now (they appear to be saying `=1.7.0` for example when it's really `>=1.7.0`). It's all a holdover from when everything was literally strings being thrown as Aff exceptions.

Perhaps we could make an issue to come back through the log messages to format them more reliably?

f-f (approved):

f-f:
I approved, but also just realised there's a nicer way to format this: instead of backticks we could use `Dodo.indent` to indent this block?

In this way:
- the console would just show it indented
- markdown would render it as a code block

thomashoneyman:
That is true, though you would need to turn the exception type from `String` to `Doc` and it may become a more pervasive change than you're thinking (hence why I've suggested a separate PR).

PR #604, opened 2023-05-24
Add ability to republish a package if Pursuit previously failed

### LINKED ISSUE #343: Allow users to reattempt publishing a package
In https://github.com/purescript/registry/issues/327#issuecomment-1034145556 @f-f stated that users should be able to trigger the registry CI to reattempt publication to Pursuit (and presumably package sets) if something unexpectedly went wrong the first time (such as a network error).

However, right now we exit the pipeline if you attempt to re-run it for an existing version:

https://github.com/purescript/registry/blob/fb20bec4af37a96353bc98c3f30391c77e2472d9/ci/src/Registry/API.purs#L277

Instead, it sounds like we should detect the existing version, and then verify that the follow-on steps completed successfully:

1. The package was uploaded to the storage backend
2. The package was uploaded to Pursuit
3. The package was added to the package sets

If we detect that any of these steps did not complete, then we reattempt them, reporting errors to the user as usual.

thomashoneyman:
Fixes https://github.com/purescript/registry-dev/issues/343. In that issue we talked about three possible failure cases:

1. We published metadata but the tarball never made it to the storage backend
2. We published the tarball but the Pursuit docs failed
3. We published the tarball and Pursuit docs but failed to add the package to the package sets

Of these three original concerns, only the second is really relevant.

1. We only publish the metadata if the tarball already made it to the storage backend, so you can already retry.
2. This PR!
3. The package sets are now independent of the publishing workflow, and if a package failed to be added there's the manual endpoint for re-attempting adding the package to the package sets.

We have some cases where a package fails to be added to Pursuit and, other than Pulp, there's currently no way to retry. In this PR I branch the `publish` workflow with the following process:

When you publish a package, we check whether the version has already been published. If not, then we go through the normal publishing workflow. If so, then we check whether the docs made it to Pursuit. If not, then we retry only the Pursuit publishing process (we don't repackage the tarball). If so, then we terminate publishing and provide you a link to your published documentation and post the metadata.

### DIFF
--- a/app/src/App/API.purs
+++ b/app/src/App/API.purs
@@ -1,4 +1,18 @@
-module Registry.App.API where
+module Registry.App.API
+  ( Source(..)
+  , PackageSetUpdateEffects
+  , packageSetUpdate
+  , PublishEffects
+  , publish
+  , AuthenticatedEffects
+  , authenticated
+  , packagingTeam
+  -- The below are exported for tests, but aren't otherwise intended for use
+  -- outside this module.
+  , formatPursuitResolutions
+  , removeIgnoredTarballFiles
+  , copyPackageSourceFiles
+  ) where
 
 import Registry.App.Prelude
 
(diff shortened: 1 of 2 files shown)


f-f (approved):
Beautiful, thanks!

PR #607, opened 2023-05-26
Upgrade to spago 0.93.3 and generate lockfile

thomashoneyman:


spago.lock
```diff
@@ -0,0 +1,1695 @@
+workspace:
+  packages:
+    registry-app:
+      path: node_modules/registry-app
```
thomashoneyman: Nooooooooooooooo!

f-f: Well, that's disappointing 😂

f-f: How did we solve it before?

thomashoneyman: I don't know why node_modules keeps getting picked up as the paths. I'll delete the directory and rebuild the lockfile.

thomashoneyman: It's as-of-yet still unsolved.

f-f: Released a 0.93.4, let's see how that one goes..

PR #608, opened 2023-05-28
Build and run the server via Nix

thomashoneyman:
This PR is a first cut at packaging the registry server and running it via Nix. Doing it via Nix means that we can do Nix-based deployments to our NixOS server, giving us access to all the niceties of a Nix server (namely, easy deployments and rollbacks).

Marking as a draft for now because this is quite ugly and I need to do more work to make it readable and convenient to use. That said: you can run the server with this command (if you have Nix):

```sh
nix run github:purescript/registry-dev/trh/nix-build
```

You should see some build output and then the server will start up:

```
 ┌───────────────────────────────────────────┐
 │ Server now up on port 8080                │
 │                                           │
 │ To test, run:                             │
 │  > curl -v localhost:8080/api/v1/jobs/0   │
 └───────────────────────────────────────────┘
```

And, accordingly, you can `curl` that endpoint to get output. Voila!


flake.nix
```diff
@@ -23,166 +18,205 @@
...
-        compilers = let
-          # Only include the compiler at normal MAJOR.MINOR.PATCH versions.
-          pursOnly = pkgs.lib.filterAttrs
-            (name: _: (builtins.match "^purs-[0-9]_[0-9]+_[0-9]$" name != null))
-            pkgs.pursPackages;
-        in pkgs.symlinkJoin {
+      npmDependencies = pkgs.callPackage ./nix/npm.nix {} {src = ./.;};
+
+      # Spago workspaces present in the repository
+      workspaces = pkgs.spago-nix {src = ./.;};
```
thomashoneyman, line 51: Almost all of this diff is a formatting change, but this part is new. We download NPM dependencies using a new Nix expression in the `nix` directory, and we get access to our Spago workspaces and their dependencies via `spago-nix`. I didn't worry too much about the diff for the time being because this is a proof-of-concept and I plan to break up the flake now that it's become so enormous.

flake.nix
```diff
@@ -23,166 +18,205 @@
...
+            for FILE in $(find ./lib/test/_fixtures/manifests -iname "*.json")
+            do
+              echo "Conforming ''${FILE} to the Manifest type"
+              cat "''${FILE}" | json-to-dhall --records-loose --unions-strict "./types/v1/Manifest.dhall" > /dev/null
+            done
+          '';
+        };
+      };
+    in rec {
+      packages = {
```
thomashoneyman, line 159: This is where we can define each package in the monorepo (the app, lib, etc.) and how to build it. For now I've opted to have the "build" include bundling, but we could just have this compile instead, then run tests and linting.

flake.nix
```diff
@@ -23,166 +18,205 @@
...
-              pursPackages.spago-next
-              pursPackages.psa
-              pursPackages.purs-tidy
-              nodePackages.bower
-            ];
-          };
+      };
+
+      apps = {
+        default = {
```
thomashoneyman, line 180: This is where we can put anything we want to be runnable, like the server, the legacy importer, etc. Each can then be run with e.g. `nix run .#server` or `nix run .#importer`.

nix/npm.nix
```diff
@@ -0,0 +1,70 @@
...
+    "lib"
+    "registry-lib"
+    "node_modules/registry-lib"
+    "foreign"
+    "registry-foreign"
+    "node_modules/registry-foreign"
+  ];
+
+  # Create an array of all (meaningful) dependencies
+  deps =
```
thomashoneyman, line 31: In other words, remove our PureScript workspace packages.

thomashoneyman:
See #618

PR #610, opened 2023-06-18
Update flake

thomashoneyman:
This updates the flake to the latest versions (and formats it). The only significant changes are:

* easy-purescript-nix now supports flakes, so we switch to that and the usual upstream
* purix supports spago-unstable and we'll use it to develop the nix build, so we add that flake input
* easy-purescript-nix refers to "nodejs_18" instead of the nixpkgs "nodejs-18_x", so we add that name to our overlay

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 /output
 /scratch
 
+result
+
 .psc*
 .purs*
 .psa*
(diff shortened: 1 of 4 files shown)


thomashoneyman:
Whoops, forgot to switch to the purix compilers, which includes 0.15.9 — the main point of this PR! This allows us to switch to 0.15.9 for upcoming package sets.

f-f (approved):

PR #612, opened 2023-06-19
Remove easy-purescript-nix

thomashoneyman:
[purix](https://github.com/thomashoneyman/purix) supports all the dev tools we need, namely purs, spago, purs-tidy, and purs-backend-es, and we already use it to gain access to purescript compilers, so this unifies our tooling via purix and drops easy-purescript-nix.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,24 +16,6 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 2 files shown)


f-f (approved):
Amazing!

Issue #602, opened 2022-10-31
Support "Colocation"


i-am-the-slime:
I am a big fan of [colocation](https://kentcdodds.com/blog/colocation). 

<details>
 <summary><h3>Example</h3></summary>
In my specific case this implies I can keep a `Spec.purs` and a `Story.purs` as descendents of `/src/`.

For example, I would like to be able to structure my project similarly to this

```sh
$> tree src
.
├── Yoga
│   ├── Block
│   │   ├── Atom
│   │   │   ├── Button
│   │   │   │   ├── Story.purs
│   │   │   │   ├── Style.purs
│   │   │   │   ├── Types.purs
│   │   │   │   └── View.purs
│   │   │   ├── Button.purs
│   │   ├── Molecule
│   │   │   ├── Sidebar
│   │   │   │   ├── Spec.purs
│   │   │   │   ├── Story.purs
│   │   │   │   ├── Style.purs
│   │   │   │   └── View.purs
│   │   │   ├── Sidebar.purs
```

</details>
<details>
 <summary><h3>Advantages</h3></summary>

I can see quickly which files have a corresponding test (`Sidebar` has one, but `Atom` doesn't). Both have a `Story`.

I do **not** need to keep the module structure between `test` and `src`.
Thus, I can easily delete or move all files relating to some functionality.
</details>

---

From what I can tell, this is necessary:

 - [ ] 1. Allow excluding files from the package (e.g., `*Spec.purs`)
 - [ ] 2. Allow building what's in `src` with more dependencies than what's specified in `spago.yaml`

Possible solution to 1:

Quoting @f-f on Discord:

> Yeah there are no globs to be specified in the new spago
> Everything is by convention: sources are in `src` and tests in `test`
> Oh actually I think this is possible
> We have a `package.files` key that you can use to specify the files to publish

i-am-the-slime:
Here's my current understanding:

spago-next as well as the registry work on a hardcoded `src` directory:

[Example 1 (registry-dev/App/API.purs)](https://github.com/purescript/registry-dev/blob/master/app/src/App/API.purs#LL369C17-L369C30):

 ```purs
   Log.debug $ "Package downloaded to " <> packageDirectory <> ", verifying it contains a src directory with valid modules..."
  Internal.Path.readPursFiles (Path.concat [ packageDirectory, "src" ]) >>= case _ of
  ```

[...]

PR #613, opened 2023-06-25
Support exclude files

i-am-the-slime:
Adds support for exclude files which enables colocation (#602)

`excludeFiles` is an array of globs. The corresponding expanded paths will be subtracted from all included paths (implicit and explicit via `files`).

That way it's possible to keep related files close together but not ship them in the package 🥳.

Right now this depends on a forked version of spaghetto which depends on this fork. I'm not sure in what order this could be merged.


f-f:
Thanks for putting this together @i-am-the-slime! I am on holiday away from screens this week, so I'll be able to review all of this once I'm back on Sunday 🙂

f-f:
@i-am-the-slime the dependency order is `spago-core -> registry -> spago`, so we'll want to merge a PR to `spago-core` first, then one to `registry` (this one), then one to `spago`

thomashoneyman:
I think we should use the same names (include / exclude), which would mean changing the Manifest type (and the associated dhall spec and description in SPEC.md). If you’d like to do that, great! If you’d rather I do it since I know where all the relevant files are then I can — just let me know.

i-am-the-slime:
@thomashoneyman I won't say no to this offer. Especially because I'm having trouble with the new spago in IntelliJ which is what I'm used to using these days.

thomashoneyman:
I ended up going with `includeFiles` and `excludeFiles` — once I was in the code, it felt more obvious what they were. @f-f, do you want to give this a look, especially how I reworded SPEC.md?

SPEC.md
```diff
@@ -199,10 +199,15 @@ All packages in the registry contain a `purs.json` manifest file in their root d
...
 - `dependencies`: dependencies of your package as key-value pairs where the keys are [`PackageName`](#packagename)s and values are [`Range`](#range)s; this is a required field, but if you have no dependencies you can provide an empty object.
-  - All dependencies you provide must exist in the registry, and the dependency ranges must be solvable (ie. it must be possible to produce a single version of each dependency that satisfies the provided version bounds, including any transitive dependencies).
+
+Note:
+
+- Globs you provide at the `includeFiles` and `excludeFiles` keys must contain only `*`, `**`, `/`, `.`, `..`, and characters for Linux file paths. It is not possible to negate a glob (ie. the `!` character), and globs cannot represent a path out of the package source directory.
+- When packaging your project source, the registry will first "include" your `src` directory and always-included files such as your `purs.json` file. Then it will include files which match globs indicated by the `includeFiles` key. Finally, it will apply the excluding globs indicated by the `excludeFiles` key to the included files (always-included files cannot be excluded).
```
f-f: We should link to the list of "always included" files if we have it anywhere

app/src/App/API.purs
```diff
@@ -899,33 +901,51 @@ formatPursuitResolutions { resolutions, dependenciesDir } =
...
+    Nothing -> pure []
+    Just nonEmptyGlobs -> do
+      let globs = map NonEmptyString.toString $ NonEmptyArray.toArray nonEmptyGlobs
+      { succeeded, failed } <- FastGlob.match source globs
+      let succeededAndNotExcluded = succeeded Array.\\ userExcludeFiles
 
       unless (Array.null failed) do
         Except.throw $ String.joinWith " "
-          [ "Some paths matched by globs in the 'files' key are outside your package directory."
+          [ "Some paths matched by globs in the 'include' key are outside your package directory."
```
f-f:
```suggestion
          [ "Some paths matched by globs in the 'includeFiles' key are outside your package directory."
```

app/src/App/API.purs
```diff
@@ -899,33 +901,51 @@ formatPursuitResolutions { resolutions, dependenciesDir } =
...
 
-      case NonEmptyArray.fromArray (Array.filter (Regex.test Internal.Path.pursFileExtensionRegex) succeeded) of
+      case NonEmptyArray.fromArray (Array.filter (Regex.test Internal.Path.pursFileExtensionRegex) succeededAndNotExcluded) of
         Nothing -> pure unit
         Just matches -> do
           let fullPaths = map (\path -> Path.concat [ source, path ]) matches
           Operation.Validation.validatePursModules fullPaths >>= case _ of
             Left formattedError ->
-              Except.throw $ "Some PureScript modules listed in the 'files' section of your manifest contain malformed or disallowed module names." <> formattedError
+              Except.throw $ "Some PureScript modules listed in the 'include' section of your manifest contain malformed or disallowed module names." <> formattedError
```
f-f:
```suggestion
              Except.throw $ "Some PureScript modules listed in the 'includeFiles' section of your manifest contain malformed or disallowed module names." <> formattedError
```

SPEC.md
thomashoneyman: We do, and I've added links to both the always-included and always-ignored files.

thomashoneyman (approved):

f-f (approved):

PR #615, opened 2023-06-26
Add nix cache to CI

thomashoneyman:
See: https://github.com/DeterminateSystems/magic-nix-cache-action

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
           github-token: ${{ env.GITHUB_TOKEN }}
 
       - name: Setup Nix cache
-        uses: DeterminateSystems/magic-nix-cache-action@v1
+        uses: DeterminateSystems/magic-nix-cache-action@v2
 
       # Separates all the cache downloading time from the time required to actually run a step
       # of the workflow
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,15 +32,13 @@ jobs:
           path: |
             .spago
             output
-
       - name: Install Nix
-        uses: cachix/install-nix-action@v16
+        uses: DeterminateSystems/nix-installer-action@v4
         with:
-          extra_nix_config: |
-            access-tokens = github.com=${{ env.GITHUB_TOKEN }}
+          github-token: ${{ env.GITHUB_TOKEN }}
 
       - name: Setup Nix cache
-        uses: DeterminateSystems/magic-nix-cache-action@main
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -39,6 +39,14 @@ jobs:
           extra_nix_config: |
             access-tokens = github.com=${{ env.GITHUB_TOKEN }}
 
+      - name: Setup Nix cache
+        uses: DeterminateSystems/magic-nix-cache-action@main
```
JordanMartinez: Shouldn't this reference a commit specifically on the repo rather than a branch name?

thomashoneyman: Yes, it really should. At the time I introduced this the v1 hadn't been released, but now it has and so we use v1.

.github/workflows/main.yml
```diff
@@ -32,12 +32,18 @@ jobs:
...
             .spago
             output
-
       - name: Install Nix
-        uses: cachix/install-nix-action@v16
+        uses: DeterminateSystems/nix-installer-action@v4
         with:
-          extra_nix_config: |
-            access-tokens = github.com=${{ env.GITHUB_TOKEN }}
+          github-token: ${{ env.GITHUB_TOKEN }}
```
JordanMartinez, line 38: Any reason why the `install-nix-action` is being replaced with DeterminateSystems' one?

thomashoneyman: Just to standardize and make sure they work well together

f-f (approved):

.github/workflows/main.yml
JordanMartinez: @f-f @thomashoneyman Did any of you read through the resulting script(s) that that action runs?

thomashoneyman: Lightly, yes — is there an issue? I’m a fan of the work coming out of Graham & determinate systems, so I’m being trusting.

If the diagnostics are an issue we can disable them.

JordanMartinez: I think what initially stood out to me when I was looking at the scripts was that the script the `action.yaml` file downloads and runs doesn't always use TLS v1.3 to download the installer that's used to install Nix. 

However, I think that only happens if the environment running the script (in this case CI) does not have a `curl` with TLS 1.3 support, which is doubtful.

Regardless, I wanted to make sure we didn't just add this action without further scrutiny.

thomashoneyman: I think it's OK — our version of curl should be fine, and determinate systems is a reliable company in the Nix community.

PR #618, opened 2023-06-28
Add Nix build for the server

thomashoneyman:
This integrates a very simple build of the registry server, integrated with `purs-backend-es`, which is suitable for use for e.g. deploying to a NixOS server or running locally. We can do a lot more with these builds, but I wanted to get something minimal up to start. I've also set up Garnix so we can get fast Nix builds in CI.

Ultimately, I think we should move all of our Nix scripts to be actual derivations which can be built by Garnix, and I think we should include `spago test` in our Flake checks so that's also run by Garnix. Then we can stop using GitHub Actions for our test suite (it's _so slow_).

We'll continue to use GitHub Actions to run the importer daily and to serve as the "registry api".


flake.nix
```diff
@@ -155,7 +143,46 @@
...
+      packages = {
+        default = let 
+          package-lock = pkgs.purix.buildPackageLock {src = ./.;};
+          spago-lock = pkgs.purix.buildSpagoLock {src = ./.; corefn = true;};
+        in pkgs.stdenv.mkDerivation {
+          name = "registry-app";
+          version = "0.0.1";
+          src = ./app;
+          phases = ["buildPhase" "installPhase"];
+          nativeBuildInputs = [pkgs.purs pkgs.purs-backend-es pkgs.esbuild];
```
thomashoneyman: We don't actually need the compiler here, so we can take that out.

```suggestion
          nativeBuildInputs = [pkgs.purs-backend-es pkgs.esbuild];
```

flake.nix
```diff
@@ -155,7 +143,46 @@
...
+          '';
+        };
+      };
+
+      apps = {
+        default = {
+          type = "app";
+          program = "${
+            pkgs.writeShellScriptBin "registry-server" ''
+              ${pkgs.nodejs}/bin/node -e 'require("${packages.default}/app.js").main()'
```
JordanMartinez, line 176: Can you clarify why we're using CJS modules here rather than ES modules? I'd expect something like:
```
'import * as Main from "${packages.default}/app.js"; main();'
```

thomashoneyman: Really I ought to write an `entrypoint.js` file that contains ES modules code like you've described there and then bundle it with esbuild — I'll take care of that later.

thomashoneyman:
Closing in favor of #619.

PR #620, opened 2023-07-14
Include runtime deps in nix builds

thomashoneyman:
In #618 and #619 I added full Nix builds for the registry such that we can run deployments to our NixOS machine via Colmena, and can run any script from this repository with a simple `nix run ...` command. Unfortunately, since you aren't running this from the context of the Nix shell, I forgot to ensure runtime dependencies were included in the builds.

If you were running these commands from within the local Nix shell then you were fine, but in CI and on the server we aren't in a shell and so runtime dependencies like `licensee` are not being picked up.

This PR adds those dependencies to the PATH for each build.

### DIFF
--- a/app/default.nix
+++ b/app/default.nix
@@ -9,6 +9,7 @@
   writeText,
   nodejs,
   compilers,
+  dhall,
   dhall-json,
   git,
   licensee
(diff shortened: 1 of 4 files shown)


colinwahl (approved):
Thanks for working this out 👍

PR #622, opened 2023-07-14
Also add gnutar to builds

thomashoneyman:
The run of the importer showed an issue with empty tarballs being created, so this adds gnutar as well. I won't merge it yet — going to wait for the legacy importer to run and verify this is working as expected, as I expected the last PR to work (tar is available, but something is strange with the tar being used in the runner).

### DIFF
--- a/app/default.nix
+++ b/app/default.nix
@@ -14,7 +14,8 @@
   git,
   licensee,
   coreutils,
-  gzip
+  gzip,
+  gnutar
 }: let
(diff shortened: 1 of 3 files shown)


colinwahl (approved):

thomashoneyman:
Finally!

https://github.com/purescript/registry/commit/aee0f99c6b06b7b0da05b569ebe5d8f0d185bb6d

PR #623, opened 2023-07-15
Add SQLite to Registry Server, it's ready to go now 🚀 

f-f:
As per title, with this PR we are ready to launch the Registry Server as our main entrypoint for package managers to issue package publishing requests.

This patch is quite large, but there are a few main themes:
- we add a SQLite database to the registry server, so that the API is able to work asynchronously.
  When an `Operation` is submitted to the server, the API will create a `Job` in the database, to which the processing of that Operation is associated.
  All the logs from that pipeline will be stored in the database, and package managers (or whatever else issued that Operation request) will be able to request those logs by their `JobId`.
  This enables a "realtime" flow, where package managers can "stream" logs from the remote pipeline as it happens. I'm quite pleased with this 🙂 
- we add a whole bunch of Nix infrastructure (in addition to what was already in place), so that we can deploy the server easily on our DigitalOcean server with Colmena
- we add a `CONTRIBUTING.md` documenting the Nix setup and more generally how to get started with contributing to the codebase
- and finally we add an integration test for the API pipeline itself

(this whole thing was coauthored with @thomashoneyman)


f-f:
@thomashoneyman you might also want to take a look at the conflicts, the nix stuff in trunk does not go well with the new things

f-f:
Also I have no idea why CI is failing - I suspect it's a problem in the npm package lock, but I just generated a fresh one 🤔

package-lock.json
```diff
@@ -50,8 +51,7 @@
     },
     "node_modules/@nodelib/fs.scandir": {
       "version": "2.1.5",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
-      "integrity": "sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==",
+      "license": "MIT",
```
thomashoneyman: The build error is because for some reason these resolved and integrity fields have been deleted in the package-lock.json file, but that’s how Nix knows where to get the package and how to verify the download

thomashoneyman:
I can merge this with master and handle the Nix conflicts today!

package-lock.json
thomashoneyman: Huh, installing with a node_modules directory present deletes the integrity and resolved keys for some packages?? NPM is so weird. A good number of issues about this and other issues with silently missing resolved / integrity fields, such as this one:
https://github.com/npm/cli/issues/4460

Fixable by deleting `node_modules` and `package-lock.json` and then reinstalling.

app/src/App/Effect/Log.purs
```diff
@@ -135,3 +129,48 @@ handleFs verbosity logfile action = case action of
...
+    -- TODO: figure out how to log here
+    -- debug $ "Notifying via a GitHub comment on issue " <> issueNumber
+    Octokit.request env.octokit request >>= case _ of
+      Left _err -> do
+        -- error $ "Could not send comment to GitHub due to an unexpected error."
+        -- debug $ Octokit.printGitHubError err
+        pure unit
+      Right _ ->
+        -- debug $ "Created GitHub comment on issue " <> issueNumber
+        pure unit
```
thomashoneyman: @f-f What was your conclusion here? Do you want to pick off the `Notify` effect in a first pass to send to GitHub and on a second pass ignore it? Do you want to undo merging these effects together? (That way we can interpret `Notify` into `Log`)

f-f: "interpret notify into log" sounds like a good way to get out of this conundrum, but part of the PR is adding `Notify` to the `LogLevel` sum type (that has since moved to `Registry.API.V1`). We could have a separate sum type that the `Log` effect uses, but that might get tedious.

f-f: I am also fine with not having any log here. And fine with getting rid of the GitHub stuff entirely if we want to take this further 😄

thomashoneyman: What's the reason to merge 'Notify' into the 'LogLevel' type? For database filtering?

The reason they were separate before is because we wish to "notify" GitHub API users of events that are not warnings or errors, such as "Starting job...", which aren't necessarily relevant for the server, and which aren't a higher log level than errors. It's kind of an orthogonal thing.

A different option is that we keep them separate and the GitHub API continues to use Notify by sending to GitHub and the server reinterprets Notify into the `Info` log level. We would then ensure in the API pipeline that 'Notify' is used for its actual purpose (to notify users of some information, not errors), and that we throw errors when the point actually is to throw an error.

f-f: The only other place where this is iffy is if we want to log inside the database layer (which is itself part of the logging layer)

thomashoneyman: Yea, ideally the database layer lives above the logging layer, and gets interpreted into `LOG + EFFECT`. Then, the log effect reuses a basic, low-level database function in `EFFECT` for writing logs only (but it doesn't have to). I can put this together if you want.

f-f: > What's the reason to merge 'Notify' into the 'LogLevel' type? For database filtering?

Yes. Though when I get logs from the Registry in Spago I ask for `Info` anyways, as the `Notify` ones are way too sparse.

> A different option is that we keep them separate and the GitHub API continues to use Notify by sending to GitHub and the server reinterprets Notify into the `Info` log level.

So I guess this is a possible way to go.

It doesn't solve the issue of the logging for the Db layer though, which makes me wonder if it's the right way to go

app/src/App/Effect/Db.purs
```diff
@@ -1,174 +1,97 @@
...
+handleSQLite :: forall r a. SQLiteEnv -> Db a -> Run (LOG + EFFECT + r) a
+handleSQLite env = case _ of
+  InsertLog log next -> do
+    Run.liftEffect $ SQLite.insertLog env.db log
+    pure next
+
+  SelectLogsByJob jobId logLevel since reply -> do
+    logs <- Run.liftEffect $ SQLite.selectLogsByJob env.db jobId logLevel since
+    unless (Array.null logs.fail) do
+      Log.warn $ "Some logs are not readable: " <> String.joinWith "\n" logs.fail
```
thomashoneyman, line 80: @f-f This is how you'd now log as part of the DB effect — it's a layer above the logging.

app/src/App/Server.purs
```diff
@@ -173,12 +178,16 @@ createServerEnv = do
   octokit <- Octokit.newOctokit vars.token
   debouncer <- Registry.newDebouncer
 
-  db <- liftEffect $ Db.connect vars.databaseUrl.path
+  db <- liftEffect $ SQLite.connect
+    { database: vars.databaseUrl.path
+    , logger: Run.runBaseEffect <<< Log.interpret (Log.handleTerminal Verbose) <<< Log.info
```
thomashoneyman, line 185: We can't log via the database without a database connection, so I opted for the terminal alone here (but we could also choose a known file on the server to write logs to so they're persistent).

app/src/App/Effect/Comment.purs
```diff
@@ -0,0 +1,68 @@
...
+comment message = Run.lift _comment (Comment (Log.toLog message) unit)
+
+interpret :: forall r a. (Comment ~> Run r) -> Run (COMMENT + r) a -> Run r a
+interpret handler = Run.interpret (Run.on _comment handler Run.send)
+
+-- | Handle a notification by converting it to an info-level LOG
+handleLog :: forall a r. Comment a -> Run (LOG + r) a
+handleLog = case _ of
+  Comment message next -> do
+    Log.info $ Ansi.foreground Ansi.BrightBlue (Dodo.text "[NOTIFY] ") <> message
```
thomashoneyman, line 44: Comments are posted directly to GitHub and not logged as well in the case the user used the GitHub API to talk to the registry. But if they used the server, then we convert comments to info-level logs and log them.

@f-f accordingly, I removed the `Notify` log level. However, you might wish to revert that so that we can do filtering on these logs specifically — we'll log them at the Notify level so they can be filtered on or whatever separately from 'Info'. I'm thinking of that since they have GitHub-flavored markdown in a number of these comments and we may not want to be sending that over the wire to Spago.

thomashoneyman (approved):
We should squash-merge this with both of our usernames attached for the git blame

f-f:
> We should squash-merge this with both of our usernames attached for the git blame

I've been doing this all along for the PRs where we both contributed 🙂

PR #624, opened 2023-07-16
Replace sunde with node-execa

### LINKED ISSUE #616: Replace sunde with node-execa
The wonderful node-execa library by @JordanMartinez is a much better choice for working with child processes in Node, and does not suffer from bugs arising from splitting buffers across code units / code points.

https://github.com/JordanMartinez/purescript-node-execa

We can essentially drop it in anywhere we use sunde now.

pete-murphy:
Fixes #616

### DIFF
--- a/app/src/App/API.purs
+++ b/app/src/App/API.purs
@@ -34,10 +34,10 @@ import Data.String.NonEmpty as NonEmptyString
 import Data.String.Regex as Regex
 import Effect.Aff as Aff
 import Effect.Ref as Ref
-import Node.ChildProcess as ChildProcess
 import Node.FS.Aff as FS.Aff
 import Node.FS.Stats as FS.Stats
 import Node.FS.Sync as FS.Sync
+import Node.Library.Execa as Execa
 import Node.Path as Path
 import Registry.App.Auth as Auth
 import Registry.App.CLI.Purs (CompilerFailure(..))
@@ -93,7 +93,6 @@ import Run.Except as Except
 import Spago.Core.Config as Spago
 import Spago.Core.Prelude as Spago.Prelude
 import Spago.Log as Spago.Log
-import Sunde as Sunde
 
 -- | Operations can be exercised for old, pre-registry packages, or for packages
 -- | which are on the 0.15 compiler series. If a true legacy package is uploaded
@@ -963,12 +962,13 @@ removeIgnoredTarballFiles path = do
 jsonToDhallManifest :: String -> Aff (Either String String)
 jsonToDhallManifest jsonStr = do
   let cmd = "json-to-dhall"
-  let stdin = Just jsonStr
   let args = [ "--records-loose", "--unions-strict", "." <> Path.sep <> Path.concat [ "types", "v1", "Manifest.dhall" ] ]
-  result <- Sunde.spawn { cmd, stdin, args } ChildProcess.defaultSpawnOptions
-  pure $ case result.exit of
-    ChildProcess.Normally 0 -> Right jsonStr
-    _ -> Left result.stderr
+  process <- Execa.execa cmd args identity
+  process.stdin.writeUtf8End jsonStr
+  result <- process.result
+  pure case result of
+    Right _ -> Right jsonStr
+    Left { stderr } -> Left stderr
 
 getPacchettiBotti :: forall r. Run (PACCHETTIBOTTI_ENV + r) Owner
 getPacchettiBotti = do
--- a/app/src/App/CLI/Licensee.purs
+++ b/app/src/App/CLI/Licensee.purs
@@ -30,7 +30,7 @@ detect directory = do
     { exitCode, stdout, stderr } = case result of
       Left { exitCode, stdout, stderr } -> { exitCode, stdout, stderr }
       Right { exitCode, stdout, stderr } -> { exitCode: Just exitCode, stdout, stderr }
-  case exitCode of
+  pure case exitCode of
     -- Licensee will exit with `1` if it didn't parse any licenses,
     -- but we consider this valid Licensee output.
     Just n | n == 0 || n == 1 -> do
@@ -44,12 +44,12 @@ detect directory = do
       case parse stdout of
         Left error -> do
           let printedError = CA.printJsonDecodeError error
-          pure $ Left printedError
+          Left printedError
         Right out -> do
           -- A NOASSERTION result means that a LICENSE file could not be parsed.
           -- For the purposes of the registry we disregard this result, since
           -- we retrieve the license via the package manifest(s) as well.
-          pure $ Right $ Array.filter (_ /= "NOASSERTION") out
+          Right $ Array.filter (_ /= "NOASSERTION") out
 
     _ ->
-      pure $ Left stderr
+      Left stderr
--- a/app/src/App/CLI/Licensee.purs
+++ b/app/src/App/CLI/Licensee.purs
@@ -6,11 +6,10 @@ import Control.Parallel as Parallel
 import Data.Array as Array
 import Data.Codec.Argonaut as CA
 import Data.Codec.Argonaut.Record as CA.Record
-import Node.ChildProcess as NodeProcess
 import Node.FS.Aff as FS
+import Node.Library.Execa as Execa
 import Node.Path as Path
 import Registry.Foreign.Tmp as Tmp
-import Sunde as Process
 
 -- | Attempt to detect the license associated with a set of provided files
 detectFiles :: Array { name :: FilePath, contents :: String } -> Aff (Either String (Array String))
@@ -25,13 +24,16 @@ detectFiles files = do
 detect :: FilePath -> Aff (Either String (Array String))
 detect directory = do
   let cmd = "licensee"
-  let stdin = Nothing
   let args = [ "detect", "--json", directory ]
-  result <- Process.spawn { cmd, stdin, args } NodeProcess.defaultSpawnOptions
-  case result.exit of
-    -- Licensee will exit with `Normally 1` if it didn't parse any licenses,
+  result <- _.result =<< Execa.execa cmd args identity
+  let
+    { exitCode, stdout, stderr } = case result of
+      Left { exitCode, stdout, stderr } -> { exitCode, stdout, stderr }
+      Right { exitCode, stdout, stderr } -> { exitCode: Just exitCode, stdout, stderr }
+  case exitCode of
+    -- Licensee will exit with `1` if it didn't parse any licenses,
     -- but we consider this valid Licensee output.
-    NodeProcess.Normally n | n == 0 || n == 1 -> do
+    Just n | n == 0 || n == 1 -> do
       let
         parse :: String -> Either JsonDecodeError (Array String)
         parse str = map (map _.spdx_id <<< _.licenses) $ flip parseJson str $ CA.Record.object "Licenses"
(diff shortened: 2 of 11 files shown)


app/src/App/CLI/Purs.purs
```diff
@@ -119,14 +119,18 @@ callCompiler compilerArgs = do
...
-  result <- try $ Process.spawn { cmd: purs, stdin: Nothing, args: printCommand compilerArgs.command } (NodeProcess.defaultSpawnOptions { cwd = compilerArgs.cwd })
+  result <- try $ _.result =<< Execa.execa purs (printCommand compilerArgs.command) (_ { cwd = compilerArgs.cwd })
   pure $ case result of
     Left exception -> Left $ case Exception.message exception of
       errorMessage
         | errorMessage == String.joinWith " " [ "spawn", purs, "ENOENT" ] -> MissingCompiler
         | otherwise -> UnknownError errorMessage
-    Right { exit: NodeProcess.Normally 0, stdout } -> Right $ String.trim stdout
-    Right { stdout, stderr } -> Left do
+    Right (Right { exitCode: 0, stdout }) -> Right $ String.trim stdout
```
pete-murphy: Not sure if I'm guaranteed a 0 exit code in the `Right` case using `execa`, so maybe the pattern-match here (and elsewhere in these changes) is unnecessary.

thomashoneyman:
Hey @pete-murphy! Thanks for doing this — I've been quite busy over the past few days so I haven't reviewed beyond a cursory glance, but it looks good so far. I'll get you a proper review this week. @JordanMartinez may also have thoughts.

f-f:
I will also try to give a review asap is no one else gets to it before me (we've been just really busy with the big effort in #623)

app/src/App/API.purs
```diff
@@ -1086,12 +1085,13 @@ removeIgnoredTarballFiles path = do
 jsonToDhallManifest :: String -> Aff (Either String String)
 jsonToDhallManifest jsonStr = do
   let cmd = "json-to-dhall"
-  let stdin = Just jsonStr
   let args = [ "--records-loose", "--unions-strict", "." <> Path.sep <> Path.concat [ "types", "v1", "Manifest.dhall" ] ]
```
JordanMartinez, line 965: Should this be refactored to:
```diff
-let args = [ "--records-loose", "--unions-strict", "." <> Path.sep <> Path.concat [ "types", "v1", "Manifest.dhall" ] ]
+let args = [ "--records-loose", "--unions-strict", Path.concat [ ".", "types", "v1", "Manifest.dhall" ] ]
```

app/src/App/Legacy/Manifest.purs
```diff
@@ -353,12 +352,15 @@ fetchSpagoDhallJson address ref = Run.Except.runExceptAt _spagoDhallError do
...
-    pure $ case result.exit of
-      NodeProcess.Normally 0 -> lmap CA.printJsonDecodeError $ parseJson CA.json result.stdout
-      _ -> Left result.stderr
+    process <- Execa.execa cmd args (_ { cwd = cwd })
+    process.stdin.writeUtf8End dhall
+    result <- process.result
+    pure $ case result of
+      Right { exitCode, stdout, stderr }
+        | exitCode == 0 -> lmap CA.printJsonDecodeError $ parseJson CA.json stdout
+        | otherwise -> Left stderr
```
JordanMartinez: Note: a `Right` case only occurs when `exitCode == 0`. I was following `execa` in how they handled things, but I've been wondering if I should separate that out so you don't need to case on the result to get the `stdout`/`stderr`/`exitCode`.

app/src/App/Effect/Git.purs
```diff
@@ -624,9 +624,11 @@ commitKeyToRepoKey = case _ of
...
-  result <- liftAff $ Sunde.spawn { cmd: "git", args, stdin: Nothing } (ChildProcess.defaultSpawnOptions { cwd = cwd })
-  let stdout = String.trim result.stdout
-  let stderr = String.trim result.stderr
-  pure $ case result.exit of
-    ChildProcess.Normally 0 -> Right stdout
+  result <- liftAff $ _.result =<< Execa.execa "git" args (_ { cwd = cwd })
+  let
+    { stdout, stderr, exitCode } = case result of
+      Left { stderr, stdout, exitCode } -> { stderr, stdout, exitCode }
+      Right { stderr, stdout, exitCode } -> { stderr, stdout, exitCode: Just exitCode }
```
JordanMartinez: Does this still need to be trimmed? `node-execa` will trim newlines at the end of `stdout`/`stderr` but doesn't do any trimming besides that.

JordanMartinez:
Otherwise, LGTM.

app/src/App/API.purs
thomashoneyman: This is intentional due to something weird about Dhall, so we should keep it. I don’t _quite_ remember the details.

app/src/App/Effect/Git.purs
thomashoneyman: The point was to trim new lines so that can go! Tests should catch any issues.

app/src/App/Legacy/Manifest.purs
pete-murphy: Thanks, I was wondering about that. I think with that knowledge I can remove a few redundant pattern matches.

app/src/App/CLI/Purs.purs
```diff
@@ -119,13 +117,13 @@ callCompiler compilerArgs = do
...
       errorMessage
-        | errorMessage == String.joinWith " " [ "spawn", purs, "ENOENT" ] -> MissingCompiler
-        | otherwise -> UnknownError errorMessage
-    Right { exit: NodeProcess.Normally 0, stdout } -> Right $ String.trim stdout
+        | errorMessage == Just (String.joinWith " " [ "spawn", purs, "ENOENT" ]) -> MissingCompiler
+        | otherwise -> UnknownError exception.message
+    Right { exitCode: 0, stdout } -> Right $ String.trim stdout
     Right { stdout, stderr } -> Left do
       case parseJson errorsCodec (String.trim stdout) of
         Left err -> UnknownError $ String.joinWith "\n" [ stdout, stderr, CA.printJsonDecodeError err ]
```
pete-murphy: If `exitCode` is always `0` in the `Right` case, this last branch is unreachable and the `CompilationError`s should be in the first branch. I don't think there's a test case covering that, I'll look into adding one.

app/test/_fixtures/ShouldFailToCompile
```diff
@@ -0,0 +1 @@
+ShouldFailToCompile
```
pete-murphy: Unsure about adding this, but I don't know how else to hit the `CompilationError` case for `Purs.callCompiler`. I could use an existing, non-`.purs` file I guess.

app/src/App/Effect/Git.purs
pete-murphy: I removed some other instances of `String.trim` on `stdout` here: [c31c9e1](https://github.com/purescript/registry-dev/pull/624/commits/c31c9e1b2588aa997d4aac42463328951d50be7a)

app/test/App/CLI/Purs.purs
```diff
@@ -32,3 +34,11 @@ spec = do
       case result of
         Left MissingCompiler -> pure unit
         _ -> Assert.fail "Should have failed with MissingCompiler"
+
+  testCompilationError =
+    Spec.it "Handles compilation error for bad input file" do
+      let fixture = Path.concat [ "test", "_fixtures", "ShouldFailToCompile" ]
+      result <- Purs.callCompiler { command: Purs.Compile { globs: [ fixture ] }, cwd: Nothing, version: Just "0.15.9" }
```
pete-murphy: I'm on an M1 Mac, and in a `nix` shell I only have access to `0.15.9` (I think because the only stable version that `purescript-overlay` supports for `aarch64-darwin` is `0.15.9`, if I'm understanding correctly) so I haven't had a chance to test this on other compilers. (I also have to change line 14 locally to `traverse_ testVersion [ "0.15.9" ]` to pass tests.)

Maybe better to leave this as `version: Nothing` since it should produce the same output regardless of version.

app/src/App/API.purs
f-f: It's because Dhall needs the `./` prefix to import a path relative to the current directory.

Node, however, will remove it:
```
> path.join(".", "foo")
'foo'
```

So we should not refactor it, and probably add the above as a comment so we don't accidentally break it in the future.

app/test/_fixtures/ShouldFailToCompile
thomashoneyman: Another approach we use is to create a temp directory in the test, write a text file, then compile it.

app/src/App/CLI/Licensee.purs
```diff
@@ -25,29 +24,32 @@ detectFiles files = do
...
-  let stdin = Nothing
   let args = [ "detect", "--json", directory ]
-  result <- Process.spawn { cmd, stdin, args } NodeProcess.defaultSpawnOptions
-  case result.exit of
-    -- Licensee will exit with `Normally 1` if it didn't parse any licenses,
+  result <- _.result =<< Execa.execa cmd args identity
+  let
+    { exitCode, stdout, stderr } = case result of
+      Left { exitCode, stdout, stderr } -> { exitCode, stdout, stderr }
+      Right { exitCode, stdout, stderr } -> { exitCode: Just exitCode, stdout, stderr }
```
thomashoneyman, line 32: Perhaps this is a little cleaner?

```suggestion
      Left exit -> exit
      Right exit -> exit { exitCode = Just exitCode }
```

thomashoneyman (dismissed):

app/src/App/CLI/Licensee.purs
pete-murphy: Hm, I can't get that to compile (seeing `TypesDoNotUnify`) I think because the `Left` and `Right` only have _some_ overlapping rows. I see
```
  Could not match type

    ( command :: String
    ...
    )

  with type

    ( failed :: Boolean
    , isCanceled :: Boolean
    , killed :: Boolean
    , message :: String
    , originalMessage :: Maybe String
    , shortMessage :: String
    , signal :: Maybe (Either Int String)
    , signalDescription :: Maybe String
    , timedOut :: Boolean
    ...
    )


while trying to match type { exitCode :: Maybe t1
                           | t2
                           }
  with type t0
while checking that expression let
                                 $1 = exit
                               in $1 { exitCode = Just (exit.exitCode)
                                     }
  has type t0
in value declaration detect

where t0 is an unknown type
      t2 is an unknown type
      t1 is an unknown type
```

pete-murphy:
Sorry for force-push, I rebased on top of latest `master` to resolve conflicts

app/src/App/CLI/Licensee.purs
thomashoneyman: Sorry, I didn't realize that. What you had before is perfect, then.

thomashoneyman (approved):

app/src/App/API.purs
JordanMartinez: Looks like that comment didn't get added in this PR.

pete-murphy: Oops, I overlooked the discussion here, I see it was added in a follow-up though https://github.com/purescript/registry-dev/pull/625

PR #630, opened 2023-07-19
Add `pursVersions` CLI script

colinwahl:
To support https://github.com/purescript/registry-dev/issues/255 we need to know the full list of compilers that the registry supports.

#628 added a script we can run to fetch all supported versions of the `purs` executable.

This PR adds the `Registry.App.CLI.PursVersions` module, which can be used to call that script within the registry codebase.

Errors are handled in the following way:
* An error that occurs calling `purs-versions` is thrown
* An error is thrown with all output of `purs-version` which fail to parse
* An error is thrown if the output of `purs-versions` is empty

I've added a test that makes sure the output of this utility matches what we expect it to be.
This means that when a new compiler version is released and `purescript-overlay` is updated, registry things will start failing until the new version is added to the snapshot. I've discussed this with @thomashoneyman and we think that behavior is desired, as it allows us to make sure we are supporting the full set of compiler versions we expect, without having to hard code them anywhere in registry app code.

### DIFF
--- a/app/src/App/CLI/PursVersions.purs
+++ b/app/src/App/CLI/PursVersions.purs
@@ -1,0 +1,26 @@
+module Registry.App.CLI.PursVersions where
+
+import Registry.App.Prelude
+
+import Data.Array as Array
+import Data.Array.NonEmpty as NEA
+import Data.String as String
+import Node.Library.Execa as Execa
+import Registry.Version as Version
+import Run (AFF, Run)
+import Run as Run
+import Run.Except (EXCEPT)
+import Run.Except as Except
+
+pursVersions :: forall r. Run (EXCEPT String + AFF + r) (NonEmptyArray Version)
+pursVersions = do
+  result <- Run.liftAff $ _.result =<< Execa.execa "purs-versions" [] identity
+  { stdout } <- Except.rethrow $ lmap (\{ stdout, stderr } -> stdout <> stderr) result
+  let { fail, success } = partitionEithers $ map Version.parse (String.split (String.Pattern " ") stdout)
+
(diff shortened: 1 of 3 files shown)


app/test/App/CLI/PursVersions.purs
```diff
@@ -0,0 +1,61 @@
...
+  , "0.15.7"
+  , "0.15.8"
+  , "0.15.9"
+  , "0.15.10"
+  ]
+
+spec :: Spec.Spec Unit
+spec = do
+  Spec.it "All expected versions are present in purs-version" do
+    versionsNonEmpty <- interpret pursVersions
```
thomashoneyman:
```suggestion
    versionsNonEmpty <- Test.Run.runBaseEffects pursVersions
```

app/src/App/CLI/PursVersions.purs
```diff
@@ -0,0 +1,26 @@
...
+import Node.Library.Execa as Execa
+import Registry.Version as Version
+import Run (AFF, Run)
+import Run as Run
+import Run.Except (EXCEPT)
+import Run.Except as Except
+
+pursVersions :: forall r. Run (EXCEPT String + AFF + r) (NonEmptyArray Version)
+pursVersions = do
+  result <- Run.liftAff $ _.result =<< Execa.execa "purs-versions" [] identity
```
thomashoneyman, line 17: `execa` can throw, I expect, so maybe we should use `Aff.attempt` here and then rethrow it into `EXCEPT` so it can be handled appropriately.

thomashoneyman (approved):

PR #631, opened 2023-07-20
Make dhall types directory configurable

thomashoneyman:
We historically have included a relative path to the `./types` directory, assuming that scripts will be executed from the root of the repository. However, now that we've packaged the server for deployment, we no longer run the server from a git checkout of this repository and need to configure where the Dhall type specifications are located on the file system.

This PR makes the directory configurable and puts it into the publishing environment, as well as setting that environment variable where appropriate. For the Nix builds that means reading it from the store; for tests and the Nix shell, that means a local path to the repository itself.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,3 +17,6 @@ PACCHETTIBOTTI_ED25519="YWJjeHl6"
 
 # The location of the sqlite database file.
 DATABASE_URL="sqlite:db/registry.sqlite3"
+
+# The location of the Dhall specifications
+DHALL_TYPES="types"
--- a/.env.example
+++ b/.env.example
@@ -19,4 +19,4 @@ PACCHETTIBOTTI_ED25519="YWJjeHl6"
 DATABASE_URL="sqlite:db/registry.sqlite3"
 
 # The location of the Dhall specifications
-DHALL_TYPES="types"
+DHALL_TYPES="./types"
--- a/app/src/App/Server.purs
+++ b/app/src/App/Server.purs
@@ -26,7 +26,7 @@ import Registry.App.Effect.Comment (COMMENT)
 import Registry.App.Effect.Comment as Comment
 import Registry.App.Effect.Db (DB)
 import Registry.App.Effect.Db as Db
-import Registry.App.Effect.Env (DatabaseUrl, PACCHETTIBOTTI_ENV)
+import Registry.App.Effect.Env (DHALL_ENV, DatabaseUrl, PACCHETTIBOTTI_ENV)
 import Registry.App.Effect.Env as Env
 import Registry.App.Effect.GitHub (GITHUB)
 import Registry.App.Effect.GitHub as GitHub
@@ -65,19 +65,17 @@ router env { route, method, body } = HTTPurple.usingCont case route, method of
   Publish, Post -> do
     publish <- HTTPurple.fromJson (jsonDecoder Operation.publishCodec) body
     lift $ Log.info $ "Received Publish request: " <> printJson Operation.publishCodec publish
-    forkPipelineJob publish.name publish.ref PublishJob
-      \jobId -> do
-        Log.info $ "Received Publish request, job id: " <> unwrap jobId
-        API.publish Current publish
+    forkPipelineJob publish.name publish.ref PublishJob \jobId -> do
+      Log.info $ "Received Publish request, job id: " <> unwrap jobId
+      API.publish Current publish
 
   Unpublish, Post -> do
     auth <- HTTPurple.fromJson (jsonDecoder Operation.authenticatedCodec) body
     case auth.payload of
       Operation.Unpublish { name, version } -> do
-        forkPipelineJob name (Version.print version) UnpublishJob
-          \jobId -> do
-            Log.info $ "Received Unpublish request, job id: " <> unwrap jobId
-            API.authenticated auth
+        forkPipelineJob name (Version.print version) UnpublishJob \jobId -> do
+          Log.info $ "Received Unpublish request, job id: " <> unwrap jobId
+          API.authenticated auth
       _ ->
         HTTPurple.badRequest "Expected unpublish operation."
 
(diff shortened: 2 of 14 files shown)


app/src/App/Server.purs
```diff
@@ -127,9 +124,12 @@ router env { route, method, body } = HTTPurple.usingCont case route, method of
...
         _fiber <- liftAff $ Aff.forkAff $ Aff.attempt $ do
-          void $ runEffects newEnv (action jobId)
-          finishedAt <- nowUTC
-          void $ runEffects newEnv (Db.finishJob { jobId, finishedAt, success: true })
+          result <- runEffects newEnv (action jobId)
+          case result of
+            Left _ -> pure unit
+            Right _ -> do
+              finishedAt <- nowUTC
+              void $ runEffects newEnv (Db.finishJob { jobId, finishedAt, success: true })
```
thomashoneyman, line 132: I also fixed this in this PR: previously, we were catching any errors and writing the job as finished to the DB in runEffects and then re-throwing them and returning in Either. Then, we threw away that result and always set the job to "success".

Now, if we get a Left (indicating an error happened in the pipeline) then we do nothing, since the job was already updated, and only set the job to success on Right.

thomashoneyman:
@JordanMartinez and I tested a deployed version of this PR and it worked:
<https://github.com/purescript/registry/commit/f255411eb1c67549cd0907fa53ae1e6b1a312e74>

Logs are a bit noisy:

```
Calling registry at https: //registry.purescript.org/api/v1/jobs/b618c773-4bad-424f-84a9-50da541b506c?since=2023-07-20T22%3A41%3A48.843Z&level=DEBUG
Calling registry at https: //registry.purescript.org/api/v1/jobs/b618c773-4bad-424f-84a9-50da541b506c?since=2023-07-20T22%3A41%3A48.843Z&level=DEBUG
2023-07-20T22: 42: 26.443Z Pushing to Pursuit...
  2023-07-20T22: 42: 26.577Z Received 201 status, which indicates the upload was successful.
  2023-07-20T22: 42: 26.579Z [NOTIFY
] Successfully uploaded package docs to Pursuit! 🎉 🚀
  2023-07-20T22: 42: 26.580Z Mirroring package node-zlib to location {
  "githubOwner": "JordanMartinez",
  "githubRepo": "purescript-node-zlib"
}
  2023-07-20T22: 42: 26.582Z Reading legacy registry from scratch/registry
Calling registry at https: //registry.purescript.org/api/v1/jobs/b618c773-4bad-424f-84a9-50da541b506c?since=2023-07-20T22%3A42%3A26.582Z&level=DEBUG
2023-07-20T22: 42: 27.349Z Fetching repo at path scratch/registry
  2023-07-20T22: 42: 27.350Z Found repo at path scratch/registry, pulling latest.
  2023-07-20T22: 42: 27.350Z Found the purescript/registry repo locally at scratch/registry, fetching latest changes...
  2023-07-20T22: 42: 27.629Z Local checkout of purescript/registry has no untracked or dirty files, it is safe to pull the latest.
  2023-07-20T22: 42: 27.630Z Local checkout of purescript/registry is up to date, not pulling.
  2023-07-20T22: 42: 27.638Z Committing to the purescript/registry repo at scratch/registry with message Sync node-zlib with legacy registry.
Calling registry at https: //registry.purescript.org/api/v1/jobs/b618c773-4bad-424f-84a9-50da541b506c?since=2023-07-20T22%3A42%3A27.638Z&level=DEBUG
2023-07-20T22: 42: 29.099Z Wrote and committed legacy registry files.
  2023-07-20T22: 42: 29.100Z [NOTIFY
] Mirrored registry operation to the legacy registry.
Job: {
  "createdAt": "2023-07-20T22:39:48.494Z",
  "finishedAt": "2023-07-20T22:42:29.101Z",
  "jobId": "b618c773-4bad-424f-84a9-50da541b506c",
  "jobType": "publish",
  "logs": [
    {
      "jobId": "b618c773-4bad-424f-84a9-50da541b506c",
      "level": "INFO",
      "message": "Wrote and committed legacy registry files.",
      "timestamp": "2023-07-20T22:42:29.099Z"
    },
    {
      "jobId": "b618c773-4bad-424f-84a9-50da541b506c",
      "level": "INFO",
      "message": "[NOTIFY] Mirrored registry operation to the legacy registry.",
      "timestamp": "2023-07-20T22:42:29.100Z"
    }
  ],
  "packageName": "node-zlib",
  "ref": "v0.2.1",
  "success": true
}

✅ Registry finished processing the package. Your package was published successfully!
```

thomashoneyman:
Reverted that last commit because the new Spago (0.93.8) fails to run the tests. Went back to the prior version.

app/src/App/Server.purs
```diff
@@ -182,7 +187,7 @@ createServerEnv = do
 
   db <- liftEffect $ SQLite.connect
     { database: vars.databaseUrl.path
-    , logger: Run.runBaseEffect <<< Log.interpret (Log.handleTerminal Verbose) <<< Log.info
+    , logger: mempty -- Run.runBaseEffect <<< Log.interpret (Log.handleTerminal Normal) <<< Log.info
```
thomashoneyman: This was just _so_ noisy it was hard to see anything else in the logs.

.env.example
```diff
@@ -17,3 +17,6 @@ PACCHETTIBOTTI_ED25519="YWJjeHl6"
 
 # The location of the sqlite database file.
 DATABASE_URL="sqlite:db/registry.sqlite3"
+
+# The location of the Dhall specifications
+DHALL_TYPES="types"
```
f-f: Should this be `./types`?

app/src/App/Server.purs
f-f: Indeed, but we should leave a comment along the lines of "uncomment this line if you'd like to see all the DB queries logged"

thomashoneyman: Fixed

.env.example
thomashoneyman: Since this isn't an executable I don't think it's _necessary_, but I agree it's better to be explicit. I've updated it.

f-f (approved):
Great, thanks!

f-f:
> Logs are a bit noisy:

Yes, _debug_ logs are supposed to be noisy, or they would be useless 😄

thomashoneyman:
This is deployed.

app/src/App/Server.purs
JordanMartinez: Shouldn't this just be another level in the debug logs?

thomashoneyman: an issue here is that we can’t log these to the database because we don’t yet have a connection to the database. We could orchestrate it by setting up a subscription where the logger pushes to the subscription and the log effect subscribes to it, writing to the db when a message is received. I just haven’t bothered putting that together 😄

thomashoneyman: But you’re right this could be set as a debug log. I don’t want them written to the terminal, though, because it makes journalctl almost unusable with all the db insert logs

f-f:
Would it be complicated to set up a deploy-on-merge with colmena+garnix?

thomashoneyman:
I don't think so — we'd do it with a GitHub workflow that runs on `master` on `push` when all checks are successful and uses one of our SSH keys (or maybe the pacchettibotti key) to run `colmena apply`.

Issue #255, opened 2021-10-18
Suggestion: Include compiler ranges in package metadata


thomashoneyman:
A PureScript package normally is compatible with some, but not all, compiler versions. I suggest that we allow package authors to specify this information in their package manifest the same way that dependencies are specified (as described in #43). In other words, a `purs.json` file would include:

```json
{ "name": "my-package"
, "version": "1.2.3"
, "compiler": ">=12.0.0 <14.0.0"
}
```

Field names could be `purs`, `purs-version`, `compiler`, `compiler-version`, or something like that.

#### Why?

A manifest including supported compiler versions opens up a few new possibilities:

* We can verify that a package compiles when added to the registry (or package set) by ensuring it compiles with a sample of compiler versions in the specified range (at least 1 per major version specified, perhaps).
* We can show what compiler versions a package supports in the Pursuit documentation.
* We can add a filter to Pursuit that allows you to look at packages only associated with particular compiler versions. Folks regularly ask to see just 0.14-compatible packages, but it would _also_ be nice to be able to go "back in time" when working on an old project and see only 0.12-compatible packages (for example).

f-f:
I think we considered it in the past - the big issue with allowing users to specify compatibility ranges is that they are often incorrect in two ways:
1. overly restrictive bounds: a package version might be compatible with a compiler version but cannot be used because the bound prevents it
2. overly lax bounds: the package author specifies a bound that includes future versions of the compiler that might end up breaking it - in this way users are allowed to use the package but it would break their build.

To sum up: we wouldn't be able to trust such bounds, which might be worse than having them in the first place!

I think a more useful solution for this issue is something like the [Hackage Matrix Builder](https://matrix.hackage.haskell.org/#), where all the packages that are uploaded to Hackage are built with various versions of the compiler to figure out the compatible versions ([here's e.g. the build matrix for the `purescript` package](https://matrix.hackage.haskell.org/#/package/purescript)).
We could set up something similar, and store all this info in the `Metadata` file for every package, here in this repo.

thomashoneyman:
I agree that compatibility ranges are likely to be incorrect for the reasons you've specified, and I _also_ think they're still useful to include somewhere that tools like Pursuit would be able to access them. So long as it's possible for Pursuit to access the `Metadata` information then it seems sensible to me that the information lives there.

I agree that something like the Hackage Matrix Builder is a good way to determine actual compatibility bounds for a package upon publication. Would a process like this make sense?

**Package Update**

[...]

PR #632, opened 2023-07-20
`compiler-versions` script: Compute supported compiler versions for a single package

colinwahl:
This PR is a first step towards #255. It adds a new script, `compiler-versions`, which will be used to compute the supported compiler versions for packages.

Currently the script can only work on a single package@version that has no dependencies. No attempt is made to modify the `Metadata` type.

In follow up PRs, I will start to introduce more sophisticated routines to get the supported compiler ranges for all existing packages, and eventually get to modifying the `Manifest` type.

Note: This currently computes a `Range` of supported compiler versions. In particular, since we use simple ranges, this means that once the package@version previously compiled for a lower compiler version and stops compiling for a later compiler, we stop checking to see if it becomes supported by a different compiler in the future. In general, I think this is fine, but wanted to call it out explicitly. 

I've run this on various versions of prelude - here's an example of the final output:

```
Found supported compiler versions for prelude@6.0.0: >=0.15.0 <0.15.11
```

### DIFF
--- a/scripts/default.nix
+++ b/scripts/default.nix
@@ -59,4 +59,5 @@ in {
   package-transferrer = build-script "registry-package-transferrer" "PackageTransferrer";
   solver = build-script "registry-solver" "Solver";
   verify-integrity = build-script "registry-verify-integrity" "VerifyIntegrity";
+  compiler-versions = build-script "compiler-versions" "CompilerVersions";
 }
--- a/scripts/default.nix
+++ b/scripts/default.nix
@@ -59,5 +59,5 @@ in {
   package-transferrer = build-script "registry-package-transferrer" "PackageTransferrer";
   solver = build-script "registry-solver" "Solver";
   verify-integrity = build-script "registry-verify-integrity" "VerifyIntegrity";
-  compiler-versions = build-script "compiler-versions" "CompilerVersions";
+  compiler-versions = build-script "registry-compiler-versions" "CompilerVersions";
 }
--- a/scripts/src/CompilerVersions.purs
+++ b/scripts/src/CompilerVersions.purs
@@ -115,19 +115,18 @@ main = launchAff_ do
   Console.log $ "Logs available at " <> logPath
 
   let
-    runDetermineCompilerVersionsForPackage :: PackageName -> Version -> Aff Unit
-    runDetermineCompilerVersionsForPackage package version =
-      determineCompilerVersionsForPackage package version
-        # Except.catch (\error -> Run.liftEffect (Console.log error *> Process.exit 1))
-        # Registry.interpret (Registry.handle registryEnv)
-        # Storage.interpret (Storage.handleReadOnly cache)
-        # GitHub.interpret (GitHub.handle { octokit, cache, ref: githubCacheRef })
-        # Log.interpret (\log -> Log.handleTerminal Normal log *> Log.handleFs Verbose logPath log)
-        # Run.runBaseAff'
+    interpret :: Run _ ~> Aff
+    interpret =
+      Except.catch (\error -> Run.liftEffect (Console.log error *> Process.exit 1))
+        >>> Registry.interpret (Registry.handle registryEnv)
+        >>> Storage.interpret (Storage.handleReadOnly cache)
+        >>> GitHub.interpret (GitHub.handle { octokit, cache, ref: githubCacheRef })
+        >>> Log.interpret (\log -> Log.handleTerminal Normal log *> Log.handleFs Verbose logPath log)
+        >>> Run.runBaseAff'
 
   case arguments of
     File _ -> Console.log "Unsupported at this time." *> liftEffect (Process.exit 1)
-    Package package version -> runDetermineCompilerVersionsForPackage package version
+    Package package version -> interpret $ determineCompilerVersionsForPackage package version
     AllPackages -> Console.log "Unsupported at this time." *> liftEffect (Process.exit 1)
 
 determineCompilerVersionsForPackage :: forall r. PackageName -> Version -> Run (AFF + EFFECT + REGISTRY + EXCEPT String + LOG + STORAGE + r) Unit
--- a/scripts/src/CompilerVersions.purs
+++ b/scripts/src/CompilerVersions.purs
@@ -2,15 +2,22 @@ module Registry.Scripts.CompilerVersions where
 
 import Registry.App.Prelude
 
+import ArgParse.Basic (ArgParser)
+import ArgParse.Basic as Arg
 import Data.Array as Array
+import Data.Array.NonEmpty as NEA
 import Data.Formatter.DateTime as Formatter.DateTime
 import Data.Map as Map
 import Data.String as String
+import Data.Tuple (uncurry)
 import Effect.Class.Console as Console
-import Effect.Ref as Ref
+import Node.FS.Aff as FS.Aff
 import Node.Path as Path
 import Node.Process as Process
 import Registry.App.CLI.Git as Git
+import Registry.App.CLI.Purs as Purs
+import Registry.App.CLI.PursVersions as PursVersions
+import Registry.App.CLI.Tar as Tar
 import Registry.App.Effect.Cache as Cache
 import Registry.App.Effect.Env as Env
 import Registry.App.Effect.GitHub as GitHub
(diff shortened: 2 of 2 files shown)


scripts/default.nix
```diff
@@ -59,4 +59,5 @@ in {
   package-transferrer = build-script "registry-package-transferrer" "PackageTransferrer";
   solver = build-script "registry-solver" "Solver";
   verify-integrity = build-script "registry-verify-integrity" "VerifyIntegrity";
+  compiler-versions = build-script "compiler-versions" "CompilerVersions";
```
thomashoneyman:
```suggestion
  compiler-versions = build-script "registry-compiler-versions" "CompilerVersions";
```

scripts/src/CompilerVersions.purs
```diff
@@ -0,0 +1,202 @@
...
+  | AllPackages
+
+parser :: ArgParser InputMode
+parser = Arg.choose "input (--file or --package or --all)"
+  [ Arg.argument [ "--file" ]
+      """Verify packages from a JSON file like: [ "prelude", "console" ]"""
+      # Arg.unformat "FILE_PATH" pure
+      # map File
+  , Arg.argument [ "--package" ]
+      "Verify the indicated package"
```
thomashoneyman: Maybe "Compute compiler ranges for" instead of "Verify"?

scripts/src/CompilerVersions.purs
```diff
@@ -0,0 +1,202 @@
...
+  now <- nowUTC
+  let logDir = Path.concat [ scratchDir, "logs" ]
+  FS.Extra.ensureDirectory logDir
+  let logFile = "compiler-versions-" <> String.take 19 (Formatter.DateTime.format Internal.Format.iso8601DateTime now) <> ".log"
+  let logPath = Path.concat [ logDir, logFile ]
+  Console.log $ "Logs available at " <> logPath
+
+  let
+    runDetermineCompilerVersionsForPackage :: PackageName -> Version -> Aff Unit
+    runDetermineCompilerVersionsForPackage package version =
```
thomashoneyman: Unless you envision effects differing from branch to branch you could make a generic `runEffects` function that can be used with any function `Run _ ~> Aff` and reuse it. For example:

```purs
case arguments of
  File packages -> runEffects$ traverse determine...
  Package package version -> runEffects $ determineCompilerVersionsForPackage package version
```

scripts/src/CompilerVersions.purs
```diff
@@ -0,0 +1,202 @@
...
+    File _ -> Console.log "Unsupported at this time." *> liftEffect (Process.exit 1)
+    Package package version -> runDetermineCompilerVersionsForPackage package version
+    AllPackages -> Console.log "Unsupported at this time." *> liftEffect (Process.exit 1)
+
+determineCompilerVersionsForPackage :: forall r. PackageName -> Version -> Run (AFF + EFFECT + REGISTRY + EXCEPT String + LOG + STORAGE + r) Unit
+determineCompilerVersionsForPackage package version = do
+  allManifests <- map ManifestIndex.toMap Registry.readAllManifests
+  compilerVersions <- PursVersions.pursVersions
+  Log.debug $ "Checking Manifest Index for " <> formatPackageVersion package version
+  Manifest { dependencies } <- rethrow $ (note "Invalid Version" <<< Map.lookup version <=< note "Invalid PackageName" <<< Map.lookup package) allManifests
```
thomashoneyman: We generally import qualified, so this would be `Except.rethrow`

thomashoneyman (approved):
Minor nits, otherwise looks great! Feel free to merge once addressed.

scripts/src/CompilerVersions.purs
colinwahl: Yeah, I wasn't sure if they'd be different or not. I guess for now I can abstract it out and then we can see what happens.

f-f:
> Note: This currently computes a Range of supported compiler versions. In particular, since we use simple ranges, this means that once the package@version previously compiled for a lower compiler version and stops compiling for a later compiler, we stop checking to see if it becomes supported by a different compiler in the future. In general, I think this is fine, but wanted to call it out explicitly.

Is this the behaviour we wish for? If a generally useful package works at some point, then it's broken by some compiler regression that is then fixed, do we not want to consider it compatible with the new series of compilers?

I would say we should do it the other way around (start checking from the most recent compiler versions rather than the older ones) if we want to keep it a single range, or just use a list of Ranges so we can cover everything.

thomashoneyman:
> if we want to keep it a single range, or just use a list of ranges so we can cover everything

This is the big design question for the feature. Do we use single versions, like `[ "0.13.0", "0.13.1" ]`, where every supported compiler is listed? Do we use ranges instead, like `">=0.13.0 <0.13.8"`? Do we follow your suggestion and use a list of ranges, like `[ ">=0.13.0 <0.13.8", ">=0.13.10 <0.13.12" ]`?

The most correct thing to do is to list out every supported version, either via single versions or via ranges. The benefit is that we have accurately captured every supported compiler for every package. But there are downsides:

1. This isn't compatible with using the solver to figure out what versions of dependencies to download based on what compilers they support. Right now we insert the compiler as a fake dependency into every package version based on what it's known to compile with (for example, `prelude@6.0.0` gets a new dependency `"purs": ">=0.15.0 <0.15.9"`). Then, when solving a new package version, we can find all of its dependencies that are also valid with the indicated compiler. We need to do this to ensure we download dependencies that are compatible with the given compiler too, or else the solver will just download the latest versions regardless of compatibility and then we might get a false compilation failure.

2. We'd no longer be able to set a cutoff for packages (such as "known not to compile from 0.14.0 onward), so when we release a new compiler we would have to compile every package version in the registry again (~13,000 versions). Realistically, most packages with no dependencies would fail right away (like `prelude`). Packages with dependencies only get tried if their dependencies indicate they support the compiler in question, so they'd be filtered out by the solver. But we're still looking at and attempting to solve 13,000 package versions, the vast majority of which is redundant work since we know that if a package version stopped compiling at 0.14.0 it's almost certainly not going to start working again at 0.15.9. 

3. (Weak) If we have a list of ranges or versions, then making the metadata files human readable via pretty-printing will cause these arrays to break over multiple lines, adding multiple lines to every version in a metadata file and ballooning the repo size. We might wish then to make metadata files not formatted, and just stringified instead, like how the manifest index is. Or switch to using `dodo-printer` so we can control this better.

In contrast, using a simple range is less correct — as you said, a compiler regression can cause a package version to falsely record that it doesn't work with some compilers that it does work with. But it is simple, and it allows us to be correct when compiling with the package version's dependencies, and in the vast majority of cases it will be correct too: when a package version stops compiling it almost certainly won't begin compiling again with new releases.

I am happy to use a list of ranges if we can figure out the solver issue. Also, we can migrate this discussion over to #255, as it affects more than this PR.

thomashoneyman:
Oh — and I agree we should prioritize either a) producing the newest range of supported compilers or b) producing the largest range of supported compilers. For example, if you work with >=0.14.0 <0.15.0 and also >=0.15.4 <0.15.5 I feel like we should select the former range as your more "natural" range.

Maybe we could tweak things when we do the bulk version of this script so that we can see, in the registry today, how many packages would end up with multiple ranges if we chose to go that way. If there are quite a few then maybe we prioritize supporting a list of ranges. If there are very few, then maybe we just use a simple range.

f-f:
Good analysis - note that I'm not advocating for a list of ranges, I'm just saying that in all the options that we have available to implement this feature (oldest working range, newest working range, list of ranges), then the current one (oldest working range) is the least correct, so we should really do something else.

While #255 is the general ticket for tracking this change, I do think that this is the right place for this discussion, as this simple assumption is the basis for further work.

colinwahl:
Thank for the discussion - yeah, after reading through it, `oldest working range` is probably not what we want.

Ideally down the line we'd be able to compute whether or not each compiler version works for each package version - maybe for now I can change this script to just determine the full list of supported compiler versions for a single package, and as we look further into implementing the bulk check we can see if that's viable or if we want to go for a different approach with ranges.

When I first was implementing this script I was thinking it'd be nice to cutoff at some point - but the side effect is the range isn't really "correct". I suppose I was optimizing too early - at least for this script, computing the whole list is fine.

thomashoneyman:
I think that's the best approach for now, and we can always switch to list-of-ranges or simple range later if need be. At least this gives us the exact versions that are usable.

colinwahl:
The latest commit changes the script to compute all supported compiler ranges for a single package version. Here's an example of the output:

`Found supported compiler versions for prelude@6.0.0: 0.15.0, 0.15.2, 0.15.3, 0.15.4, 0.15.5, 0.15.6, 0.15.7, 0.15.8, 0.15.9, 0.15.10`
`Found supported compiler versions for prelude@3.0.0: 0.13.0, 0.13.2, 0.13.3, 0.13.4, 0.13.5, 0.13.6, 0.13.8, 0.14.0, 0.14.1, 0.14.2, 0.14.3, 0.14.4, 0.14.5, 0.14.6, 0.14.7, 0.14.8, 0.14.9`

thomashoneyman (approved):

PR #633, opened 2023-07-22
Enable debug database logs

thomashoneyman:
In #631 I disabled the database logger because it was so noisy and we could only log to the terminal. Now, these logs are proper `debug` logs and are written to the database. (Of course, given the sheer number of these things, we might want to be a bit more judicious about how many get written).

The way it works is a little funky because we need the log function at the time we connect to the database, but we can't log to the database until we're connected. So we defer it:

1. We set up an emitter/listener pair
2. The "log" function provided during db connection pushes to the listener
3. We subscribe to the emitter, interpreting messages as calls to `Log.debug`, and running the effect using the Log effect's database handler.
4. Once the operation is complete (or threw an exception) we clean up the subscription

I noticed that we're using a single database connection across all jobs, which means there's no way to distinguish debug logs coming from the database for one operation vs. another one. Now, calls to `runEffect` will set up a new database connection, logs will associate with the given job id (if there is one — to the terminal if not), and the database connection will be cleaned up once the operation completes.

@f-f you may wish to handle these db connections differently. I'm not sure what issues may arise from having too many database connections open at once, or whether we want to be smarter than "any call to runEffects sets up and tears down a database connection".

### DIFF
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -29,6 +29,7 @@ package:
     - foldable-traversable
     - foreign-object
     - formatters
+    - halogen-subscriptions
     - http-methods
     - httpurple
     - identity
--- a/app/src/App/SQLite.js
+++ b/app/src/App/SQLite.js
@@ -2,13 +2,27 @@ import Database from "better-sqlite3";
 
 export const connectImpl = (path, logger) => {
   logger("Connecting to database at " + path);
+
   let db = new Database(path, {
     fileMustExist: true,
     verbose: logger,
   });
+
   db.pragma("journal_mode = WAL");
   db.pragma("foreign_keys = ON");
+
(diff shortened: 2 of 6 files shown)


f-f:
I'm not sure I follow - is this for logging into the DB itself all the queries that the database makes? Why would we want that?

thomashoneyman:
Ha, I thought that was what you and Jordan were getting at with me disabling the db logger in #631 😆 Personally, I don't want to store them in the db, and I don't want to print them to the terminal; we could potentially log to a file, but I don't necessarily want to clutter up the file system either.

f-f:
Yeah this is why I wanted you to leave a comment in the place where one could enable them - I see them being useful locally when implementing database calls, but likely not that useful as general logs

PR #638, opened 2023-07-26
Replace withBackoff with more explicit alternatives

thomashoneyman:
There have been a few times in which I mistakenly assumed `withBackoff` would retry on failure, but that isn't correct: it sets a timeout and cancels the action if it exceeds the timeout, extending the timeout using exponential backoff on each cancellation.

This PR renames `withBackoff` to `withRetryOnTimeout` and introduces a new family of `withRetry` functions that make it easier to retry `Aff`-based actions on failure. For example, `withRetryRequest` lets you make an HTTP request and retry when the request succeeds but an error status code was returned.

### DIFF
--- a/app/src/App/Effect/GitHub.purs
+++ b/app/src/App/Effect/GitHub.purs
@@ -258,18 +258,25 @@ request octokit githubRequest@{ route: route@(GitHubRoute method _ _), codec } =
 requestWithBackoff :: forall a r. Octokit -> Request a -> Run (LOG + AFF + r) (Either Octokit.GitHubError a)
 requestWithBackoff octokit githubRequest = do
   Log.debug $ "Making request to " <> Octokit.printGitHubRoute githubRequest.route
-  let action = Octokit.request octokit githubRequest
-  result <- Run.liftAff $ withBackoff
-    { delay: Duration.Milliseconds 5_000.0
-    , action
-    , shouldCancel: \_ -> Octokit.request octokit Octokit.rateLimitRequest >>= case _ of
-        Right { remaining } | remaining == 0 -> pure false
-        _ -> pure true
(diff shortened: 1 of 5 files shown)


app/src/App/Effect/GitHub.purs
```diff
@@ -258,18 +258,25 @@ request octokit githubRequest@{ route: route@(GitHubRoute method _ _), codec } =
 requestWithBackoff :: forall a r. Octokit -> Request a -> Run (LOG + AFF + r) (Either Octokit.GitHubError a)
 requestWithBackoff octokit githubRequest = do
   Log.debug $ "Making request to " <> Octokit.printGitHubRoute githubRequest.route
-  let action = Octokit.request octokit githubRequest
-  result <- Run.liftAff $ withBackoff
-    { delay: Duration.Milliseconds 5_000.0
-    , action
-    , shouldCancel: \_ -> Octokit.request octokit Octokit.rateLimitRequest >>= case _ of
```
thomashoneyman, line 265: This, for example, is intended to retry requests unless we’ve exceeded our rate limit. But shouldCancel just checks whether we should let the timeout kill the request, and on a rate limit exceeded failure we get a quick 403. So it looks like this is coming from a misunderstanding of how withBackoff previously worked.

f-f (approved):

f-f:
I have copied the previous iteration of this code in Spago as I didn't want to depend on the Registry app, but maybe I should depend on it? Not sure if we'd like this in `lib`

thomashoneyman:
I'd rather not put this in `lib` because I'd like to be able to arbitrarily change it in the future. It could potentially go in `foreign` (despite not really being a foreign binding) as `Registry.Foreign.Aff` and be a collection of Aff utils or something. But I think the best thing is to copy it into Spago and tweak it to your needs.

PR #639, opened 2023-07-26
`compiler-versions` script: Compute supported compiler versions for all packages

colinwahl:
This PR builds on #632 and is a step towards #255. It extends to `compiler-versions` script with the option to compute the supported compiler versions for all packages in the manifest index.

The algorithm for finding the supported packages for a specific compiler version is the following:

1. Topologically sort the current manifest index, initialize an empty `DependencyIndex`
2. For each `Manifest`, try to solve via the previous `DependencyIndex` and then compile - if it succeeds, add the entry to the `DependencyIndex`, else record failure and continue

The end result is a `DependencyIndex` with as many entries as we could solve and then compile with that compiler version. Any `Manifest` not found there either didn't solve (meaning its dependencies are not supported by that compiler version) or didn't compiler (meaning its dependencies are supported by that compiler version, but that package is not). We record all entries and continue to the next compiler.

I am currently dumping results to a file and logging progress messages per entry we check. There are comprehensive debug logs. The failure results are sorted in topological order so that we can see how failures propagate throughout.

### Usage:

For all packages, all compilers:
`GITHUB_TOKEN=xxx spago run -p registry-scripts -m Registry.Scripts.CompilerVersions -- --all-packages --all-compilers`

For all packages, with a single compiler:
`GITHUB_TOKEN=xxx spago run -p registry-scripts -m Registry.Scripts.CompilerVersions -- --all-packages --compiler 0.13.0`

For a single package, checking all compilers:
`GITHUB_TOKEN=xxx spago run -p registry-scripts -m Registry.Scripts.CompilerVersions -- --package prelude@3.0.0 --all-compilers`

For a single package, checking a single compiler:
`GITHUB_TOKEN=xxx spago run -p registry-scripts -m Registry.Scripts.CompilerVersions -- --package prelude@3.0.0 --compiler 0.13.0`

### Future Work:

I've ran the script on 0.15.0 and 0.13.0 and the output looked fine at a glance. In the future we will need to add some verification checks (perhaps making sure the package sets for that compiler version are all included in this output), but I've left that as follow on work.

I've ran into spurious API failures while installing packages quite often, which should be alleviated by #638.

### DIFF
--- a/app/src/App/CLI/Purs.purs
+++ b/app/src/App/CLI/Purs.purs
@@ -123,7 +123,7 @@ callCompiler compilerArgs = do
       | originalMessage == Just (String.joinWith " " [ "spawn", purs, "ENOENT" ]) -> Left MissingCompiler
     Left { stdout, stderr } -> Left do
       case parseJson errorsCodec stdout of
-        Left err -> UnknownError $ String.joinWith "\n" [ stdout, stderr, CA.printJsonDecodeError err ]
+        Left err -> UnknownError $ String.joinWith "\n" [ "\n[STDOUT]", stdout, "[STDERR]", stderr, "[DECODE ERROR]", CA.printJsonDecodeError err ]
         Right ({ errors } :: { errors :: Array CompilerError })
           | Array.null errors -> UnknownError "Non-normal exit code, but no errors reported."
           | otherwise -> CompilationError errors
--- a/app/src/App/CLI/Purs.purs
+++ b/app/src/App/CLI/Purs.purs
@@ -71,12 +71,15 @@ printCompilerErrors errors = do
   String.joinWith "\n" printed
   where
   printCompilerError :: CompilerError -> String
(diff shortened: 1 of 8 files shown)


thomashoneyman:
I think the algorithm is sound and I've been able to run this with the 0.15.x series and 0.14.x series of the compiler (trying 0.13.x now). Once I've verified the various CLI combinations I'll take a deeper code review. At a glance there are some lingering `Debug` imports we should remove.

thomashoneyman:
I like the progress output, thanks for doing that!

```
[0.13.0 1343/10830] Checking aff-promise@2.0.0
```

f-f (approved):

thomashoneyman:
It's taking a very long time to verify the 0.13.0 compiler, but so far I'm about 7,000 package versions in and no random failures (with #638 included on top of this code).

PR #640, opened 2023-07-27
Add caching to compile-versions

thomashoneyman:
The `compile-versions` script has no caching right now, which means every run will try solving and compilation anew for, in the worst case, every compiler * every package version, or hundreds of thousands of compiler / package / version combinations.

This PR adds caching to `compile-versions` based on the package, version, and compiler version, so we don't recompile already-visited package versions. In the future we could take this back out by using the metadata files themselves as the "cache", but until we add compiler versions to the metadata files this is the only way we have to save run results.

### DIFF
--- a/scripts/src/CompilerVersions.purs
+++ b/scripts/src/CompilerVersions.purs
@@ -7,13 +7,15 @@ import ArgParse.Basic as Arg
 import Data.Array as Array
 import Data.Array.NonEmpty as NEA
 import Data.Codec.Argonaut as CA
+import Data.Codec.Argonaut.Common as Codec.Common
 import Data.Codec.Argonaut.Record as CA.Record
 import Data.Codec.Argonaut.Variant as CA.Variant
+import Data.Exists as Exists
 import Data.Formatter.DateTime as Formatter.DateTime
 import Data.Map as Map
-import Data.Maybe as Maybe
 import Data.Profunctor as Profunctor
 import Data.Semigroup.Foldable as Semigroup.Foldable
+import Data.Set as Set
 import Data.String as String
 import Data.Variant as Variant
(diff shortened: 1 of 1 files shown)


thomashoneyman:
@f-f I requested your review because Colin is on PTO until next week.

scripts/src/CompilerVersions.purs
```diff
@@ -135,57 +144,66 @@ main = launchAff_ do
...
+   . PackageName
+  -> Version
+  -> TargetCompiler
+  -> Run (REGISTRY + STORAGE + LOG + EXCEPT String + AFF + EFFECT + r) Unit
+compilersForPackageVersion package version target = do
+  allManifests <- Registry.readAllManifests
+  supportedCompilers <- PursVersions.pursVersions
+  Log.debug $ "Checking manifest index for " <> formatPackageVersion package version
+  Manifest { dependencies } <- Except.rethrow (note "No entry found in manifest index." (ManifestIndex.lookup package version allManifests))
+  -- FIXME: Support packages with dependencies once we have compilers versions
```
f-f, line 175: I'm confused - so if a package has any dependency we don't check it? But then we're only looking at Prelude?

thomashoneyman: This is for the single-package CLI flag. We don’t yet store what compilers a package works with in metadata so we don’t know what of the package’s dependencies work with the given compiler. So for now this CLI arg only works with zero dependency packages. (It’s been that way for the life of the script.)

The bulk mode of the script works because we start with the zero dependency packages and build from there.

Once we do store compiler support in package metadata, then we can fix this single-package option so it works for any package by looking at its dependents’ metadata.

f-f (approved):

PR #654, opened 2023-08-09
Add healthchecks to server

### LINKED ISSUE #648: Add healthchecks to server
We don't have any server monitoring right now, so if the registry server goes down we have no idea. However, we can use healthchecks.io service to provide monitoring and alert the infrastructure team.

If we wanted to self-host on a server apart from the registry server:
* There is a [module in nixpkgs](https://github.com/NixOS/nixpkgs/blob/fef2dfa379a6fae356a589974e5266ff16b0573b/nixos/modules/services/web-apps/healthchecks.nix) for healthchecks we can set up on the server
* The [self-hosted documentation](https://healthchecks.io/docs/self_hosted) describes the [settings we can use to configure the service](https://healthchecks.io/docs/self_hosted_configuration)
* We can use the `DISCORD_CLIENT_ID` and `DISCORD_CLIENT_SECRET` to configure an integration with our Discord server and send notifications to the #infrastructure (or #registry) channels

Note: we should only do this once #641 has merged, as this will lead to conflicts in the Nix configuration.

thomashoneyman:
Closes #648 by registering the server with both uptimerobot.com and healthchecks.io. Uptime Robot pings our server, and we ping Health Checks.

In the future we can build out the health check to be more than just a GET request; it could notify a 'start' event, then do some integrity checking on the server (we're in sync with the upstream Git repositories, the storage is in sync with the manifest index, etc.), then notify a 'verified' event. But for now it's just notifying that it's alive.

As seen earlier today, these checks are integrated into the #infrastructure channel in Discord as well as the pacchettibotti email, so we'll be notified if the server runs into any hiccups.


thomashoneyman:
(Note: the code as seen here has already been deployed, for testing.)

app/src/App/Effect/Env.purs
```diff
@@ -202,13 +205,20 @@ githubApiUrl = EnvKey { key: "GITHUB_API_URL", decode: pure }
...
 
 defaultPursuitApiUrl :: URL
 defaultPursuitApiUrl = "https://pursuit.purescript.org"
 
+-- | The URL of the health checks endpoint
+healthchecksUrl :: EnvKey URL
+healthchecksUrl = EnvKey { key: "HEALTHCHECKS_URL", decode: pure }
+
+defaultHealthchecksUrl :: URL
+defaultHealthchecksUrl = "https://hc-ping.com/9094e12b-399a-4a6a-8758-418be7520550"
```
f-f: This token should stay secret, as it's the unique identified for the check. We should have a clearly fake uuid here

app/src/App/Server.purs
```diff
@@ -224,6 +235,17 @@ main = do
       Console.log $ "Failed to start server: " <> Aff.message error
       Process.exit 1
     Right env -> do
+      _fiber <- Aff.launchAff do
```
f-f:
```suggestion
      _healthcheck <- Aff.launchAff do
```

thomashoneyman:
I've made that a required (secret) env var, and set the actual values elsewhere. The server is updated.

f-f (approved):

PR #656, opened 2023-08-19
Use current time as published date

### LINKED ISSUE #652: Incorrect time measurements in unpublish workflow prevents unpublishing
As seen in:
https://github.com/purescript/registry/issues/346#issuecomment-1665808201

...a package published just minutes prior cannot be unpublished because, and I quote:

> Cannot unpublish web-fetch@4.0.0 because it was published 6063.176484722222 hours ago. Packages can only be unpublished for 48 hours after publication.

This error arises here:
https://github.com/purescript/registry-dev/blob/3166e0982d37d96cfc586fcd0791de937822d2c1/app/src/App/API.purs#L243-L256

But the real implementation is deeper, in the registry library itself:
https://github.com/purescript/registry-dev/blob/3166e0982d37d96cfc586fcd0791de937822d2c1/lib/src/Operation/Validation.purs#L133-L158

We're comparing the current time in the UTC time zone to the published time associated with the package, which is calculated here:
https://github.com/purescript/registry-dev/blob/3166e0982d37d96cfc586fcd0791de937822d2c1/app/src/App/Effect/Source.purs#L94-L108

The problem in this specific case:

1. A commit to the repository was made in 2022
2. The tag associated with that commit wasn't made until today
3. We published the package, looked up the ref time, and voila: we see 2022 and assign that as the 'published time'
4. Now the publish time is long in the past, and you can't unpublish.

The reason why we use the ref time as the publish time is to support the legacy importer, in which we need to associate old publish times so the registry doesn't say everything was published in August 2023. But we probably should be using the current time in the publish pipeline instead.

pete-murphy:
Fixes #652

### DIFF
--- a/scripts/src/PackageDeleter.purs
+++ b/scripts/src/PackageDeleter.purs
@@ -237,7 +237,7 @@ deleteVersion arguments name version = do
           Just (Left _) -> Log.error "Cannot reimport a version that was specifically unpublished"
           Just (Right specificPackageMetadata) -> do
             -- Obtains `newMetadata` via cache
-            API.publish PackageSource'Legacy
+            API.publish LegacyPackage
               { location: Just oldMetadata.location
               , name: name
               , ref: specificPackageMetadata.ref
--- a/scripts/src/PackageDeleter.purs
+++ b/scripts/src/PackageDeleter.purs
@@ -237,7 +237,7 @@ deleteVersion arguments name version = do
           Just (Left _) -> Log.error "Cannot reimport a version that was specifically unpublished"
           Just (Right specificPackageMetadata) -> do
             -- Obtains `newMetadata` via cache
-            API.publish API.Legacy
+            API.publish PackageSource'Legacy
               { location: Just oldMetadata.location
               , name: name
               , ref: specificPackageMetadata.ref
--- a/app/src/App/API.purs
+++ b/app/src/App/API.purs
@@ -1,6 +1,5 @@
 module Registry.App.API
-  ( Source(..)
-  , PackageSetUpdateEffects
+  ( PackageSetUpdateEffects
   , packageSetUpdate
   , PublishEffects
   , publish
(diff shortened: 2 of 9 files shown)


scripts/src/PackageDeleter.purs
```diff
@@ -237,7 +237,7 @@ deleteVersion arguments name version = do
           Just (Left _) -> Log.error "Cannot reimport a version that was specifically unpublished"
           Just (Right specificPackageMetadata) -> do
             -- Obtains `newMetadata` via cache
-            API.publish API.Legacy
+            API.publish PackageSource'Legacy
```
thomashoneyman: We typically only use quotes in this codebase to represent types or functions that are a tweak on another; could we use something like “LegacyPackage” and “CurrentPackage” as constructors instead?

pete-murphy: Yeah, sorry was intentionally leaving this as "ugly"/unconventional to remind myself to comment about alternatives and ask for suggestions about the naming, since I'm not sure what to call the type/constructors here. I'll just paste what I had started typing up in another comment:

This was originally
```haskell
data Source = Legacy | Current
```
but after moving to `Prelude` that would conflict with `Source` here https://github.com/purescript/registry-dev/blob/282a4f049af74df888c9871b340b2c4100650491/app/src/App/Effect/Source.purs#L32 and with `Current` constructor here https://github.com/purescript/registry-dev/blob/282a4f049af74df888c9871b340b2c4100650491/app/src/App/CLI/Git.purs#L304

Another alternative would be keeping the original constructor names but hide `Current` when importing in the `CLI.Git` module, or renaming the `GitStatus` constructor.

pete-murphy: But also happy rename things to
```haskell
data PackageSource = LegacyPackage | CurrentPackage
```
if that's preferable

pete-murphy: Updated [706ea23](https://github.com/purescript/registry-dev/pull/656/commits/706ea2380513df2454274b3ad54e716a3a816400)

pete-murphy:
Not sure why the Garnix checks are failing, don't see any output in the logs

thomashoneyman (approved):

thomashoneyman:
Thank you @pete-murphy!

PR #657, opened 2023-08-24
Replace Affjax with Fetch 

### LINKED ISSUE #643: Replace Affjax with Fetch
The [fetch](https://github.com/rowtype-yoga/purescript-fetch) library by @sigma-andex and @i-am-the-slime takes advantage of Node's support of the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) since Node 18. It's a little more comfortable for sending buffer data in Node (see #642), and it has no JS library dependencies — Affjax relies on the elderly `xhr2` library.

This issue tracks replacing our use of Affjax with Fetch. Fortunately we don't make many requests; they're all in the following modules:

* [App.Prelude](https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/src/App/Prelude.purs#L159), used in the `withRetryRequest` helper
* [App.Effect.Source](https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/src/App/Effect/Source.purs#L133-L137), used to download tarballs from GitHub into a buffer
* [App.Effect.Storage](https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/src/App/Effect/Storage.purs#L242-L247), used to download tarballs from the registry backend into a buffer
* [App.Effect.Pursuit](https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/src/App/Effect/Pursuit.purs#L65-L78), used to POST json data to Pursuit to upload packages

We'll also want to drop the dependencies in `spago.yaml`:
https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/spago.yaml#L10-L11

As well as drop the `xhr2` dependency from `package.json`:
https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/package.json#L10

CharlesTaylor7:
Resolves #643

Replaces Affjax with Fetch.
I recommend reviewing `Fetch.Retry` first, the rest of changes are pretty mechanical. 
There's probably a case to be made for moving some or all of the `Fetch.Retry` module into the actual fetch package.
Just let me know what you want to do.

### DIFF
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -7,8 +7,6 @@ package:
     version: 0.0.1
   dependencies:
     - aff
-    - affjax
-    - affjax-node
     - ansi
     - argonaut-core
     - arrays
@@ -26,6 +24,7 @@ package:
     - exceptions
     - exists
     - filterable
+    - fetch
     - foldable-traversable
     - foreign-object
     - formatters
@@ -35,6 +34,7 @@ package:
     - integers
     - js-date
     - js-uri
+    - js-promise-aff
     - lists
     - maybe
     - media-types
--- a/app/src/App/CLI/Git.purs
+++ b/app/src/App/CLI/Git.purs
@@ -2,7 +2,6 @@ module Registry.App.CLI.Git where
 
 import Registry.App.Prelude
 
-import Affjax (URL)
 import Data.Array as Array
 import Data.Array.NonEmpty as NonEmptyArray
 import Data.String as String
--- a/app/src/App/Effect/Pursuit.purs
+++ b/app/src/App/Effect/Pursuit.purs
@@ -3,19 +3,15 @@ module Registry.App.Effect.Pursuit where
 
 import Registry.App.Prelude
 
-import Affjax.Node (URL)
-import Affjax.Node as Affjax.Node
-import Affjax.RequestBody as RequestBody
-import Affjax.RequestHeader as RequestHeader
-import Affjax.ResponseFormat as ResponseFormat
-import Affjax.StatusCode (StatusCode(..))
 import Data.Argonaut.Core as Argonaut
 import Data.Array as Array
 import Data.Codec.Argonaut as CA
 import Data.HTTP.Method as Method
 import Data.Map as Map
-import Data.MediaType.Common as MediaType
 import Data.Profunctor as Profunctor
+import Effect.Exception as Exception
+import Fetch.Retry as Fetch
+import Foreign (unsafeFromForeign)
 import Registry.App.Effect.Env (RESOURCE_ENV)
 import Registry.App.Effect.Env as Env
 import Registry.App.Effect.Log (LOG)
@@ -65,37 +61,34 @@ handleAff (GitHubToken token) = case _ of
     { pursuitApiUrl } <- Env.askResourceEnv
     Log.debug "Pushing to Pursuit..."
 
-    result <- Run.liftAff $ withRetryRequest'
-      { content: Just $ RequestBody.json payload
-      , headers:
-          [ RequestHeader.Accept MediaType.applicationJSON
-          , RequestHeader.RequestHeader "Authorization" ("token " <> token)
-          ]
-      , method: Left Method.POST
-      , username: Nothing
-      , withCredentials: false
-      , password: Nothing
-      , responseFormat: ResponseFormat.string
-      , timeout: Nothing
-      , url: Array.fold [ pursuitApiUrl, "/packages" ]
-      }
+    result <- Run.liftAff $
+      Fetch.withRetryRequest (Array.fold [ pursuitApiUrl, "/packages" ])
+        { method: Method.POST
+        , body: Argonaut.stringify payload
+        , headers:
+            { "Accept": "application/json"
+            , "Authorization": "token " <> token
+            }
+        }
 
     result' <- case result of
       Cancelled -> do
         Log.error $ "Pursuit failed to connect after several retries."
         pure $ Left $ "Expected to receive a 201 status from Pursuit, but failed to connect after several retries."
       Failed reqError -> case reqError of
-        AffjaxError err -> do
-          pure $ Left $ "Pursuit publishing failed with an HTTP error: " <> Affjax.Node.printError err
-        StatusError { body, status: StatusCode status } -> do
-          Log.error $ "Pursuit publishing failed with status " <> show status <> " and body\n" <> body
+        Fetch.FetchError err -> do
+          pure $ Left $ "Pursuit publishing failed with an HTTP error: " <> Exception.message err
+        Fetch.StatusError { text: textAff, status } -> do
+          text <- Run.liftAff textAff
+          Log.error $ "Pursuit publishing failed with status " <> show status <> " and body\n" <> text
           pure $ Left $ "Expected to receive a 201 status from Pursuit, but received " <> show status <> " instead."
-      Succeeded { body, status: StatusCode status }
+      Succeeded { text: textAff, status }
         | status == 201 -> do
             Log.debug "Received 201 status, which indicates the upload was successful."
             pure $ Right unit
         | otherwise -> do
-            Log.error $ "Pursuit publishing failed with status " <> show status <> " and body\n" <> body
+            text <- Run.liftAff textAff
+            Log.error $ "Pursuit publishing failed with status " <> show status <> " and body\n" <> text
             pure $ Left $ "Expected to receive a 201 status from Pursuit, but received " <> show status <> " instead."
 
     pure $ reply result'
@@ -105,37 +98,33 @@ handleAff (GitHubToken token) = case _ of
     let name = PackageName.print pname
     let url = Array.fold [ pursuitApiUrl, "/packages/purescript-" <> name <> "/available-versions" ]
     Log.debug $ "Checking if package docs for " <> name <> " are published on Pursuit using endpoint " <> url
-    result <- Run.liftAff $ withRetryRequest'
-      { content: Nothing
-      , headers: [ RequestHeader.Accept MediaType.applicationJSON ]
-      , method: Left Method.GET
-      , username: Nothing
-      , withCredentials: false
-      , password: Nothing
-      , responseFormat: ResponseFormat.json
-      , timeout: Nothing
-      , url
+    result <- Run.liftAff $ Fetch.withRetryRequest url
+      { headers: { accept: "application/json" }
       }
 
     case result of
       Cancelled -> do
         Log.error $ "Could not reach Pursuit after multiple retries at URL " <> url
         pure $ reply $ Left $ "Could not reach Pursuit to determine published versions for " <> name
-      Failed (AffjaxError httpError) -> do
-        let printedError = Affjax.Node.printError httpError
+      Failed (Fetch.FetchError httpError) -> do
+        let printedError = Exception.message httpError
         Log.error $ "Pursuit publishing failed because of an HTTP error: " <> printedError
         pure $ reply $ Left "Could not reach Pursuit due to an HTTP error."
-      Failed (StatusError { body, status: StatusCode status }) -> do
-        Log.error $ "Could not fetch published versions from Pursuit (received non-200 response) " <> show status <> " and body\n" <> Argonaut.stringify body
+      Failed (Fetch.StatusError { text: textAff, status }) -> do
+        text <- Run.liftAff textAff
+        Log.error $ "Could not fetch published versions from Pursuit (received non-200 response) " <> show status <> " and body\n" <> text
         pure $ reply $ Left $ "Received non-200 response from Pursuit: " <> show status
-      Succeeded { body } -> case CA.decode availableVersionsCodec body of
-        Left error -> do
-          let printed = CA.printJsonDecodeError error
-          Log.error $ "Failed to decode body " <> Argonaut.stringify body <> "\n with error: " <> printed
-          pure $ reply $ Left $ "Received a response from Pursuit, but it could not be decoded:\n\n" <> printed <> "\n\ncc: @purescript/packaging"
-        Right versions -> do
-          Log.debug "Found versions from Pursuit!"
-          pure $ reply $ Right versions
+      Succeeded { text: textAff, json: jsonAff } -> do
+        json <- Run.liftAff jsonAff
+        case CA.decode availableVersionsCodec (unsafeFromForeign json) of
+          Left error -> do
+            let printed = CA.printJsonDecodeError error
+            text <- Run.liftAff textAff
+            Log.error $ "Failed to decode body " <> text <> "\n with error: " <> printed
+            pure $ reply $ Left $ "Received a response from Pursuit, but it could not be decoded:\n\n" <> printed <> "\n\ncc: @purescript/packaging"
+          Right versions -> do
+            Log.debug "Found versions from Pursuit!"
+            pure $ reply $ Right versions
 
 -- The Pursuit /available-versions endpoint returns versions as a tuple of the
 -- version number and documentation URL, represented as a two-element array.
--- a/app/src/App/Prelude.purs
+++ b/app/src/App/Prelude.purs
@@ -2,9 +2,9 @@ module Registry.App.Prelude
   ( LogVerbosity(..)
   , PackageSource(..)
   , PursPublishMethod(..)
-  , RetryRequestError(..)
   , Retry
   , RetryResult(..)
+  , URL
   , class Functor2
   , defaultRetry
   , formatPackageVersion
@@ -32,16 +32,11 @@ module Registry.App.Prelude
   , unsafeFromRight
   , withRetry
   , withRetryOnTimeout
-  , withRetryRequest
-  , withRetryRequest'
   , writeJsonFile
   ) where
 
 import Prelude
 
-import Affjax as Affjax
-import Affjax.Node as Affjax.Node
-import Affjax.StatusCode (StatusCode(..))
 import Control.Alt ((<|>)) as Extra
 import Control.Alternative (guard) as Extra
 import Control.Monad.Except (ExceptT(..)) as Extra
@@ -97,6 +92,8 @@ import Registry.Types (License, Location(..), Manifest(..), ManifestIndex, Metad
 import Registry.Version as Version
 import Type.Proxy (Proxy(..)) as Extra
 import Type.Row (type (+)) as Extra
+
+type URL = String
 
 class Functor2 (c :: Type -> Type -> Type) where
   map2 :: forall a b z. (a -> b) -> c z a -> c z b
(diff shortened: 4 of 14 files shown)


app/spago.yaml
```diff
@@ -35,6 +34,7 @@ package:
     - integers
     - js-date
     - js-uri
+    - js-promise-aff
```
CharlesTaylor7, line 37: This is a transitive dependency of `fetch`, so it incurs no overhead.

app/src/App/CLI/Git.purs
```diff
@@ -2,7 +2,6 @@ module Registry.App.CLI.Git where
 
 import Registry.App.Prelude
 
-import Affjax (URL)
```
CharlesTaylor7, line 5: `URL` is just a type alias for `String`, defined by Affjax.
Since the alias was used throughout the app, I've redefined the same alias in `Registry.App.Prelude`.

app/src/App/Prelude.purs
```diff
@@ -155,36 +152,6 @@ mapKeys k = Map.fromFoldable <<< map (Extra.lmap k) <<< (Map.toUnfoldable :: _ -
...
-withRetryRequest
-  :: forall e a b
-   . Retry e
-  -> (Affjax.Error -> e)
-  -> (Affjax.Response a -> Either.Either e b)
-  -> Affjax.Request a
-  -> Extra.Aff (RetryResult e b)
-withRetryRequest retry onAffjaxError onAffjaxResponse request =
-  withRetry retry (map (Either.either (Either.Left <<< onAffjaxError) onAffjaxResponse) (Affjax.Node.request request))
-
```
CharlesTaylor7, line 189: Lines 158-187, were lifted and rewritten for use in the module `Fetch.Retry`

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,88 @@
...
+  -> { | input }
+  -> Aff (RetryResult RetryRequestError Response)
+withRetryRequest url r = do
+  request <- liftEffect $ new url $ Request.convert r
+
+  let req :: CoreRequest.Request
+      req = request
+
+      fetch :: Effect (Promise CoreResponse.Response)
+      fetch = Core.fetch req
```
CharlesTaylor7: This gives me an error saying:

```purs
[1/1 TypesDoNotUnify] app/src/Fetch/Retry.purs:54:15

  54        fetch = Core.fetch req
                    ^^^^^^^^^^^^^^

  Could not match type
    Promise
  with type
    Promise
  while trying to match type Promise Response
    with type Promise Response
  while checking that expression fetch req
    has type Effect (Promise Response)
  in value declaration withRetryRequest
```

So i'm presuming incompatible versions of `js-promise` are being pulled in, but that is a bit surprising. I thought using a package set would prevent this kind of diamond dependency problem.

app/src/Fetch/Retry.purs
```diff
@@ -45,27 +45,19 @@ withRetryRequest
...
-  let req :: CoreRequest.Request
-      req = request
-
-      fetch :: Effect (Promise CoreResponse.Response)
-      fetch = Core.fetch req
-
-      runRequest :: Aff CoreResponse.Response
-      runRequest = Promise.Aff.toAffE fetch
+  let runRequest :: Aff Response
+      runRequest = Fetch.fetch url r
```
CharlesTaylor7: The type signature of `withRetryRequest` was copied and adapted directly from the `fetch` source code. The only difference in these signatures is the return types. 

The error message seems to be about ambiguous type variables, but I don't have a clue about to constrain the type any futher.

Error Message:

```purs
[1/1 NoInstanceFound] app/src/Fetch/Retry.purs:49:20

  49        runRequest = Fetch.fetch url r
                         ^^^^^^^^^^^^^^^

  No type class instance was found for
    Prim.Row.Union input5
                   t1
                   ( body :: String
                   , cache :: RequestCache
                   , credentials :: RequestCredentials
                   , headers :: Record t2
                   , integrity :: Integrity
                   , method :: Method
                   , mode :: RequestMode
                   , referrer :: Referrer
                   , referrerPolicy :: ReferrerPolicy
                   )
  The instance head contains unknown type variables. Consider adding a type annotation.
  while applying a function fetch
    of type Union @Type t0 t1
              ( body :: String
              , cache :: RequestCache
              , credentials :: RequestCredentials
              , headers :: Record t2
              , integrity :: Integrity
              , method :: Method
              , mode :: RequestMode
              , referrer :: Referrer
              , referrerPolicy :: ReferrerPolicy
              )
             => Union @Type t3 t4
                  ( body :: RequestBody
                  , cache :: String
                  , credentials :: String
                  , headers :: Headers
                  , integrity :: String
                  , method :: String
                  , mode :: String
                  , referrer :: String
                  , referrerPolicy :: String
                  )
                 => ToCoreRequestOptions t0 t3 => String
                                                  -> Record t0
                                                     -> Aff
                                                          { arrayBuffer :: ...
                                                          , blob :: ...
                                                          , body :: ...
                                                          , headers :: Headers
                                                          , json :: ...
                                                          , ok :: Boolean
                                                          , redirected :: Boolean
                                                          , status :: Int
                                                          , statusText :: String
                                                          , text :: ...
                                                          , url :: String
                                                          }
    to argument url
  while inferring the type of fetch url
  in value declaration withRetryRequest
  where input5 is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        t0 is an unknown type
        t3 is an unknown type
        t1 is an unknown type
        t4 is an unknown type
        t2 is an unknown type
```

CharlesTaylor7:
@thomashoneyman I've tried a few different approaches to reimplementing `withRetryRequest` and I could use some help. I'm running into very arcane type errors, and pretty much stuck at this point.

See these 2 commits for the two main approaches I tried:

- [Using Fetch.Core](https://github.com/purescript/registry-dev/pull/657/commits/474852caf5e7beba6143a9c3098da36915e82fe0?diff=unified&w=0#r1304948892)
- [Using just fetch](https://github.com/purescript/registry-dev/pull/657/commits/e9b1a5c900d1fc32fcef4d292a961a89bc090577?diff=unified&w=0#r1304950084)


On the bright side, I was able to port the rest of the application to use the new fetch based api, just haven't figured out how to implement the retry helper in a way to satisfy the type checker.

app/src/Fetch/Retry.purs
pete-murphy: There's a relevant thread on the Discourse: https://discourse.purescript.org/t/specialize-function-via-annotation/3616. The fix there also applies here, this compiles

```diff
  withRetryRequest
    :: forall input output thruIn thruOut headers
     . Union input thruIn (HighlevelRequestOptions headers String)
    => Union output thruOut CoreRequest.UnsafeRequestOptions
    => ToCoreRequestOptions input output
    => String
    -> { | input }
    -> Aff (RetryResult RetryRequestError Response)
  withRetryRequest url r = do
    let
      runRequest :: Aff Response
-     runRequest = fetch url r
+     runRequest = fetchWithProxy (Proxy :: Proxy thruIn) url r
  
    retryRequest runRequest
  
+ fetchWithProxy
+   :: forall input output thruIn thruOut headers
+    . Union input thruIn (HighlevelRequestOptions headers String)
+   => Union output thruOut CoreRequest.UnsafeRequestOptions
+   => ToCoreRequestOptions input output
+   => Proxy thruIn
+   -> String
+   -> { | input }
+   -> Aff Response
+ fetchWithProxy _ url r = do
+   request <- liftEffect $ new url $ Request.convert r
+   cResponse <- Promise.Aff.toAffE $ Core.fetch request
+   pure $ Response.convert cResponse
```

Visible type applications (https://github.com/purescript/purescript/releases/tag/v0.15.10) is a new feature in PureScript which lets us leave out the `Proxy` argument, so this also compiles

```diff
  withRetryRequest
    :: forall input output thruIn thruOut headers
     . Union input thruIn (HighlevelRequestOptions headers String)
    => Union output thruOut CoreRequest.UnsafeRequestOptions
    => ToCoreRequestOptions input output
    => String
    -> { | input }
    -> Aff (RetryResult RetryRequestError Response)
  withRetryRequest url r = do
    let
      runRequest :: Aff Response
-     runRequest = fetchWithProxy (Proxy :: Proxy thruIn) url r
+     runRequest = fetch @thruIn url r
  
    retryRequest runRequest
  
- fetchWithProxy
-   :: forall input output thruIn thruOut headers
+ fetch
+   :: forall input output @thruIn thruOut headers
     . Union input thruIn (HighlevelRequestOptions headers String)
    => Union output thruOut CoreRequest.UnsafeRequestOptions
    => ToCoreRequestOptions input output
-   => Proxy thruIn
-   -> String
+   => String
    -> { | input }
    -> Aff Response
- fetchWithProxy _ url r = do
+ fetch url r = do
    request <- liftEffect $ new url $ Request.convert r
    cResponse <- Promise.Aff.toAffE $ Core.fetch request
    pure $ Response.convert cResponse
```

The only difference from the `fetch` source code is the `@thruIn`.

Sorry it doesn't help much with this PR, but it seems like it would be worthwhile to expose `@thruIn`  in the `fetch` implementation upstream to use with VTAs (and maybe the other type parameters as well? not sure what the heuristic is).

app/src/Fetch/Retry.purs
pete-murphy: I think the package set maybe hasn't been updated in a while? If I change this line https://github.com/purescript/registry-dev/blob/282a4f049af74df888c9871b340b2c4100650491/spago.yaml#L4
to point to a more recent package set (`37.0.0`) `spago build` succeeds without the `promiseToPromise` function. This PR seems relevant https://github.com/rowtype-yoga/purescript-fetch-core/commit/4d74acdd83f373604557f06be42724cee82ae91c

CharlesTaylor7:
@pete-murphy Thanks for the help!

@thomashoneyman
This is ready for review now.

foreign/package.json
```diff
@@ -7,13 +7,15 @@
...
     "fast-glob": "^3.2.11",
     "fs-extra": "^10.0.0",
     "fuse.js": "^6.6.2",
     "jsonrepair": "^2.2.1",
     "registry-lib": "file:../lib",
     "semver": "^7.5.2",
     "tar": "^6.1.11",
-    "tmp": "^0.2.1"
+    "tmp": "^0.2.1",
+    "yaml": "^2.3.1"
```
CharlesTaylor7, line 19: `tmp`, `yaml`, and `better-sqlite3` are all required by foreign modules, and the test suite wouldn't run in my nix shell without adding these.

app/src/App/Effect/Pursuit.purs
```diff
@@ -65,37 +60,34 @@ handleAff (GitHubToken token) = case _ of
     { pursuitApiUrl } <- Env.askResourceEnv
     Log.debug "Pushing to Pursuit..."
 
-    result <- Run.liftAff $ withRetryRequest'
-      { content: Just $ RequestBody.json payload
+    result <- Run.liftAff $ 
+      Fetch.withRetryRequest (Array.fold [ pursuitApiUrl, "/packages" ])
```
thomashoneyman:
```suggestion
    result <- Run.liftAff $ Fetch.withRetryRequest (Array.fold [ pursuitApiUrl, "/packages" ])
```

app/src/App/Effect/Pursuit.purs
```diff
@@ -65,37 +60,34 @@ handleAff (GitHubToken token) = case _ of
...
-      , method: Left Method.POST
-      , username: Nothing
-      , withCredentials: false
-      , password: Nothing
-      , responseFormat: ResponseFormat.string
-      , timeout: Nothing
-      , url: Array.fold [ pursuitApiUrl, "/packages" ]
+        { "Accept": "application/json"
+        , "Authorization": "token " <> token
+        }
```
thomashoneyman: You'll need to run `purs-tidy` over this code to get through CI — things like this record should be indented two spaces.

```suggestion
          { "Accept": "application/json"
          , "Authorization": "token " <> token
          }
```

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+          StatusError { status } ->
+            -- We retry on 500-level errors in case the server is temporarily
+            -- unresponsive, and fail otherwise.
+            if status >= 500 then
+              attempt < 3
+            else false
+      }
+  
+
+-- | Copied and adapted from fetch source code
```
thomashoneyman: I think we should link to this issue, as it describes more in-depth why this proxy is necessary:
https://github.com/rowtype-yoga/purescript-fetch/issues/9

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+withRetryRequest
+  :: forall input output thruIn thruOut headers
+   . Union input thruIn (HighlevelRequestOptions headers String)
+  => Union output thruOut CoreRequest.UnsafeRequestOptions
+  => ToCoreRequestOptions input output
+  => String
+  -> { | input }
+  -> Aff (RetryResult RetryRequestError Response)
+withRetryRequest url r = retryRequest $ 
+  fetchWithProxy (Proxy :: _ thruIn) url r
```
thomashoneyman:
```suggestion
withRetryRequest url input = retryRequest $ fetchWithProxy (Proxy :: _ thruIn) url input
```

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+import Promise.Aff as Promise.Aff
+
+
+type URL = String
+
+data RetryRequestError
+  = FetchError FetchError
+  | StatusError Response
+
+newtype FetchError = WrapFetchError Error
```
thomashoneyman: I don't think we need to wrap `Error` in a newtype; we can just have `RetryRequestError` take `Error` directly in the `FetchError` constructor, e.g.

```purs
data RetryRequestError
  = FetchError Error
  | StatusError Response
```

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+
+foreign import printError :: FetchError -> String
```
thomashoneyman: We don't need this function, because [`Exception.message`](https://pursuit.purescript.org/packages/purescript-exceptions/6.0.0/docs/Effect.Exception#v:message) already provides this for `Error`. When we need to print a JS error we should just use `Exception.message` directly.

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
+module Fetch.Retry 
+  ( withRetryRequest
+  , printError
+  , FetchError
+  , URL
+  , RetryRequestError(..)
+  , module Fetch
```
thomashoneyman: I would rather not re-export the whole Fetch module, and just export the functions we've written in this module. We should remove this re-export.

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+import Fetch (HighlevelRequestOptions, Response, class ToCoreRequestOptions, new)
+import Fetch.Internal.Request as Request
+import Fetch.Core as Core
+import Fetch.Core.Request as CoreRequest
+import Fetch.Internal.Response as Response
+import Prim.Row (class Union)
+import Promise.Aff as Promise.Aff
+
+
+type URL = String
```
thomashoneyman: You've already defined `URL` in the `App.Prelude`, so this should be removed.

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+      (Either.either
+        (Either.Left <<< FetchError <<< WrapFetchError)
+        onFetchResponse
+      )
+  where
+    onFetchResponse :: Response -> Either RetryRequestError Response
+    onFetchResponse response = 
+      if response.status >= 400 
+        then Either.Left $ StatusError response 
+        else Either.Right response
```
thomashoneyman: Typically we just `import Data.Either (Either(..))` and use these constructors directly. We don't do that in the Prelude module because of all the re-exports, but in other modules it's fine to use `Left` and `Right` without being qualified.

```suggestion
        then Left $ StatusError response 
        else Right response
```

app/src/Fetch/Retry.purs
```diff
@@ -0,0 +1,90 @@
...
+  => Union output thruOut CoreRequest.UnsafeRequestOptions
+  => ToCoreRequestOptions input output
+  => Proxy thruIn
+  -> String
+  -> { | input }
+  -> Aff Response
+fetchWithProxy _ url r = do
+  request <- liftEffect $ new url $ Request.convert r
+  cResponse <- Promise.Aff.toAffE $ Core.fetch request
+  pure $ Response.convert cResponse
```
thomashoneyman:
```suggestion
fetchWithProxy _ url input = do
  request <- liftEffect $ new url $ Request.convert input
  coreResponse <- Promise.Aff.toAffE $ Core.fetch request
  pure $ Response.convert coreResponse
```

thomashoneyman (commented):
This is a great start! Thanks for your work to puzzle out a good way to get fetch working with the retry function. I have a few comments, but overall this looks quite close.

thomashoneyman:
@MonoidMusician A bit of an odd error in the tests, coming from the solver:

```
   ✗ Nested dependency of target package has no versions in range.:

  "No versions found in the registry for does-not-exist in range\n  >=0.0.0 seen in fixed-broken@2.0.0\n  <5.0.0 seen in fixed-broken@2.0.0" ≠ "No versions found in the registry for does-not-exist in range\n  >=0.0.0 seen in fixed-broken@2.0.0 from declared dependencies transitive-broken\n  <5.0.0 seen in fixed-broken@2.0.0 from declared dependencies transitive-broken"
```

The only thing that would affect the solver here are dependency updates; I know that there were recently changes to `ordered-collections` — do you see anything that would be causing this change to the error message?

app/src/Fetch/Retry.purs
sigma-andex: We can add this to fetch so we can use vta here

thomashoneyman: We're using 0.15.10 in the registry so feel free to use VTAs if you'd like.

sigma-andex: great. I have just tagged a new version (v3.0.0) of fetch with the vta change.

CharlesTaylor7:
@thomashoneyman Alright, I resolved the comments you left & squashed down to 1 commit.

CharlesTaylor7:
@thomashoneyman I manually tested the one failing test against different registry versions, and found the last version that works is 35.1.0, and the first failing version is 35.2.0.

thomashoneyman:
Looks like that corresponds with ordered-collections being updated to 3.1.0:
https://github.com/purescript/registry/blob/main/package-sets/35.2.0.json

Release is here, looks like a pretty major rewrite of the internals:
https://github.com/purescript/purescript-ordered-collections/releases/tag/v3.1.0

I'll have to review separately and see if the change to behavior is benign or not.

app/src/App/Effect/Pursuit.purs
```diff
@@ -105,37 +98,33 @@ handleAff (GitHubToken token) = case _ of
...
-      , headers: [ RequestHeader.Accept MediaType.applicationJSON ]
-      , method: Left Method.GET
-      , username: Nothing
-      , withCredentials: false
-      , password: Nothing
-      , responseFormat: ResponseFormat.json
-      , timeout: Nothing
-      , url
+    result <- Run.liftAff $ Fetch.withRetryRequest url
+      { headers: { accept: "application/json" }
```
thomashoneyman, line 102: Should this be “Accept” the way it is above?

thomashoneyman:
The code looks good — I’ll give this a try later today and hopefully we can get it merged!

MonoidMusician:
@thomashoneyman I don't see an obvious reason why the error message would have changed, assuming both the old Data.Map and new one are both correct. The API surface I used is pretty small. Maybe traversal order changed somewhere? But it's a pretty benign change to that error message so I don't think it's a big deal. The algorithm should still be correct.

thomashoneyman (approved):

natefaubion:
> Maybe traversal order changed somewhere?

This is certainly possible. Both old and new `traverse` order effects structurally (ie, equivalent to `toUnfoldableUnordered`), and there's no reason to assume that the trees internally would produce the same effect order.

PR #659, opened 2023-09-04
Gzip package documentation when publishing to Pursuit

### LINKED ISSUE #642: Gzip package documentation when sending to Pursuit
The registry publishes packages to Pursuit by sending the JSON output of the compiler's `purs publish` command in the body of a POST request:

https://github.com/purescript/registry-dev/blob/64e3a6c444ba5a82d61cd0522f1300e5d719992a/app/src/App/Effect/Pursuit.purs#L59-L78

We're consistently seeing 502 error codes from Pursuit for some packages on upload. These are really 413 Content Too Large errors (they're getting garbled on their way through nginx). Pursuit won't accept package uploads larger than 1.5mb. See, for example:

* https://github.com/purescript/registry/issues/317
* https://github.com/purescript/registry/issues/295
* https://github.com/purescript/registry/issues/152

However, I've noticed that some packages fail when being published via the registry, but succeed when published via Pulp. The difference: Pulp sends a gzipped response to Pursuit, and we don't.

First Pulp gzips the JSON from `purs publish`

https://github.com/purescript-contrib/pulp/blob/78f7aa6ae76337ea6f8362cac03fb1c8ee1858cd/src/Pulp/Publish.purs#L68

https://github.com/purescript-contrib/pulp/blob/78f7aa6ae76337ea6f8362cac03fb1c8ee1858cd/src/Pulp/Publish.purs#L170-L171

Then it passes that to the publish docs function:

https://github.com/purescript-contrib/pulp/blob/78f7aa6ae76337ea6f8362cac03fb1c8ee1858cd/src/Pulp/Publish.purs#L85

https://github.com/purescript-contrib/pulp/blob/78f7aa6ae76337ea6f8362cac03fb1c8ee1858cd/src/Pulp/Publish.purs#L326-L327

In contrast, we're just sending the JSON directly. We should gzip the JSON as well.

**Implementation**

We have a few options for the implementation. To actually gzip the data we could:

1. Adapt the Pulp implementation
    This probably isn't our best bet because Pulp relies on a number of NPM libraries we don't want to bring in (`concat-stream`, etc.). I'd like to avoid pulling new JS dependencies unless absolutely necessary. But we can at least take a look at what they're doing.

2. Use [purescript-node-zlib](https://github.com/jordanmartinez/purescript-node-zlip) by @JordanMartinez for the `createGzip` function, or write our own pure JS gzip function using the underlying Node code.
    This is a recent library Jordan published which gives us access to the zlib Node module (same as is used by Pulp) and has functions like [createGzip](https://github.com/JordanMartinez/purescript-node-zlib/blob/7dc9eda45cf427bf68458ff6293d31de6ccb4c57/src/Node/Zlib.purs#L157-L162) already available. We could potentially write some glue around this to gzip the input JSON. Or we could write a thin FFI around createGzip that just gzips an input string.

3. Use the `gzip` CLI
    We have access to the `gzip` CLI via Nix; this is already used, for example, in the implementation of the [`Registry.App.CLI.Tar` module](https://github.com/purescript/registry-dev/blob/64e3a6c444ba5a82d61cd0522f1300e5d719992a/app/src/App/CLI/Tar.purs#L24-L28). We could potentially introduce a `Registry.App.CLI.Gzip` module that exposes a `gzipCLI` function and a `gzip :: String -> Aff String Buffer` function that uses gzip with the `"-"` argument to read/write over stdin/stdout.

Once we have the gzipped data we can send it with an appropriate Content-Encoding header in the Affjax request. (Presumably we'd send it as a `RequestBody.string` or `RequestBody.blob` — not sure which we should use.)

pete-murphy:
Fixes #642.

### DIFF
--- a/app/src/App/Effect/Pursuit.purs
+++ b/app/src/App/Effect/Pursuit.purs
@@ -12,12 +12,14 @@ import Data.Profunctor as Profunctor
 import Effect.Exception as Exception
 import Fetch.Retry as Fetch
 import Foreign (unsafeFromForeign)
+import Node.Buffer as Buffer
 import Registry.App.Effect.Env (RESOURCE_ENV)
 import Registry.App.Effect.Env as Env
 import Registry.App.Effect.Log (LOG)
 import Registry.App.Effect.Log as Log
 import Registry.App.Legacy.LenientVersion (LenientVersion(..))
 import Registry.App.Legacy.LenientVersion as LenientVersion
+import Registry.Foreign.Gzip as Gzip
 import Registry.Foreign.Octokit (GitHubToken(..))
 import Registry.PackageName as PackageName
 import Registry.Version as Version
@@ -61,12 +63,15 @@ handleAff (GitHubToken token) = case _ of
     { pursuitApiUrl } <- Env.askResourceEnv
     Log.debug "Pushing to Pursuit..."
 
-    result <- Run.liftAff $
+    result <- Run.liftAff do
+      gzipped <- Gzip.compress (Argonaut.stringify payload)
+      body <- liftEffect (Buffer.toString UTF8 gzipped)
       Fetch.withRetryRequest (Array.fold [ pursuitApiUrl, "/packages" ])
         { method: Method.POST
-        , body: Argonaut.stringify payload
+        , body
         , headers:
             { "Accept": "application/json"
+            , "Content-Encoding": "gzip"
             , "Authorization": "token " <> token
             }
         }
--- a/app/src/App/Effect/Pursuit.purs
+++ b/app/src/App/Effect/Pursuit.purs
@@ -101,7 +101,7 @@ handleAff (GitHubToken token) = case _ of
     let name = PackageName.print pname
     let url = Array.fold [ pursuitApiUrl, "/packages/purescript-" <> name <> "/available-versions" ]
     Log.debug $ "Checking if package docs for " <> name <> " are published on Pursuit using endpoint " <> url
-    result <- Run.liftAff $ Fetch.withRetryRequest @String url
+    result <- Run.liftAff $ Fetch.withRetryRequest url
       { headers: { accept: "application/json" }
       }
 
--- a/app/src/App/Effect/Pursuit.purs
+++ b/app/src/App/Effect/Pursuit.purs
@@ -12,7 +12,6 @@ import Data.Profunctor as Profunctor
 import Effect.Exception as Exception
 import Fetch.Retry as Fetch
 import Foreign (unsafeFromForeign)
-import Node.Buffer as Buffer
 import Registry.App.Effect.Env (RESOURCE_ENV)
 import Registry.App.Effect.Env as Env
 import Registry.App.Effect.Log (LOG)
@@ -65,10 +64,9 @@ handleAff (GitHubToken token) = case _ of
 
     result <- Run.liftAff do
       gzipped <- Gzip.compress (Argonaut.stringify payload)
-      body <- liftEffect (Buffer.toString UTF8 gzipped)
       Fetch.withRetryRequest (Array.fold [ pursuitApiUrl, "/packages" ])
         { method: Method.POST
-        , body
+        , body: gzipped
         , headers:
             { "Accept": "application/json"
             , "Content-Encoding": "gzip"
@@ -103,7 +101,7 @@ handleAff (GitHubToken token) = case _ of
     let name = PackageName.print pname
     let url = Array.fold [ pursuitApiUrl, "/packages/purescript-" <> name <> "/available-versions" ]
     Log.debug $ "Checking if package docs for " <> name <> " are published on Pursuit using endpoint " <> url
-    result <- Run.liftAff $ Fetch.withRetryRequest url
+    result <- Run.liftAff $ Fetch.withRetryRequest @String url
       { headers: { accept: "application/json" }
       }
 
--- a/app/src/Fetch/Retry.purs
+++ b/app/src/Fetch/Retry.purs
@@ -8,7 +8,7 @@ import Registry.App.Prelude
 
 import Effect.Aff (Error)
 import Effect.Aff as Aff
-import Fetch (class ToCoreRequestOptions, HighlevelRequestOptions, Response, new)
+import Fetch (class ToCoreRequestOptions, class ToRequestBody, HighlevelRequestOptions, Response, new)
 import Fetch (Response) as ReExport
 import Fetch.Core as Core
 import Fetch.Core.Request as CoreRequest
@@ -22,8 +22,9 @@ data RetryRequestError
   | StatusError Response
 
 withRetryRequest
-  :: forall input output thruIn thruOut headers
-   . Union input thruIn (HighlevelRequestOptions headers String)
+  :: forall input output thruIn thruOut headers @body
+   . ToRequestBody body
+  => Union input thruIn (HighlevelRequestOptions headers body)
   => Union output thruOut CoreRequest.UnsafeRequestOptions
   => ToCoreRequestOptions input output
   => String
@@ -31,7 +32,7 @@ withRetryRequest
   -> Aff (RetryResult RetryRequestError Response)
 withRetryRequest url r =
   withRetry retry
-    $ (Aff.attempt $ fetch @thruIn url r)
+    $ (Aff.attempt $ fetchBody @thruIn url r)
     <#>
       ( either
           (Left <<< FetchError)
@@ -60,15 +61,16 @@ withRetryRequest url r =
 -- | The upstream library will soon be updated to use Visible Type Application
 -- | and then we can drop this redefinition:
 -- | https://github.com/rowtype-yoga/purescript-fetch/issues/9
-fetch
-  :: forall input output @thruIn thruOut headers
-   . Union input thruIn (HighlevelRequestOptions headers String)
+fetchBody
+  :: forall input output @thruIn thruOut headers @body
+   . ToRequestBody body
+  => Union input thruIn (HighlevelRequestOptions headers body)
   => Union output thruOut CoreRequest.UnsafeRequestOptions
   => ToCoreRequestOptions input output
   => String
   -> { | input }
   -> Aff Response
-fetch url input = do
+fetchBody url input = do
   request <- liftEffect $ new url $ Request.convert input
   coreResponse <- Promise.Aff.toAffE $ Core.fetch request
   pure $ Response.convert coreResponse
--- a/app/src/Fetch/Retry.purs
+++ b/app/src/Fetch/Retry.purs
@@ -62,7 +62,7 @@ withRetryRequest url r =
 -- | and then we can drop this redefinition:
 -- | https://github.com/rowtype-yoga/purescript-fetch/issues/9
 fetchBody
-  :: forall input output @thruIn thruOut headers @body
+  :: forall input output @thruIn thruOut headers body
    . ToRequestBody body
   => Union input thruIn (HighlevelRequestOptions headers body)
   => Union output thruOut CoreRequest.UnsafeRequestOptions
--- a/app/src/Fetch/Retry.purs
+++ b/app/src/Fetch/Retry.purs
@@ -22,7 +22,7 @@ data RetryRequestError
   | StatusError Response
 
 withRetryRequest
-  :: forall input output thruIn thruOut headers @body
+  :: forall input output thruIn thruOut headers body
    . Union input thruIn (HighlevelRequestOptions headers body)
   => Union output thruOut CoreRequest.UnsafeRequestOptions
   => ToCoreRequestOptions input output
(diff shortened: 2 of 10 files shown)


app/src/App/Effect/Pursuit.purs
```diff
@@ -61,12 +63,15 @@ handleAff (GitHubToken token) = case _ of
     { pursuitApiUrl } <- Env.askResourceEnv
     Log.debug "Pushing to Pursuit..."
 
-    result <- Run.liftAff $
+    result <- Run.liftAff do
+      gzipped <- Gzip.compress (Argonaut.stringify payload)
+      body <- liftEffect (Buffer.toString UTF8 gzipped)
```
pete-murphy: I haven't tested this. I've tested sending a `Buffer` as request body directly without converting to string, but that's not currently supported in the upstream types for `RequestBody` https://github.com/purescript-web/purescript-web-fetch/issues/15.

foreign/test/Foreign/Gzip.purs
```diff
@@ -0,0 +1,24 @@
...
+spec :: Spec.Spec Unit
+spec = do
+  Spec.it "Compresses a string with gzip" do
+    tmp <- Tmp.mkTmpDir
+    let
+      file = Path.concat [ tmp, "out.gz" ]
+      contents = "<contents>"
+    buffer <- Gzip.compress contents
+    FS.Aff.writeFile file buffer
+    result <- _.result =<< Execa.execa "gzcat" [ file ] identity
```
pete-murphy: I don't know if I can assume `gzcat` is available where the tests are being run, I think its available on `ubuntu-latest`?

foreign/test/Foreign/Gzip.purs
```diff
@@ -13,12 +13,13 @@ import Test.Spec as Spec
...
+  Spec.describe "compress" do
+    Spec.it "Compresses a string with gzip" do
+      tmp <- Tmp.mkTmpDir
+      let
+        file = Path.concat [ tmp, "out.gz" ]
+        contents = "<contents>"
+      buffer <- Gzip.compress contents
+      FS.Aff.writeFile file buffer
+      result <- _.result =<< Execa.execa "gzcat" [ file ] identity
+      (_.stdout <$> result) `Assert.shouldEqual` Right contents
```
pete-murphy: Odd, without the `describe` wrapper, the `Gzip` test results are lumped in with the `FastGlob` results.

Before:
```
Foreign » FastGlob
  ✓︎ Prevents directory traversal
  ✓︎ Prevents symlink directory traversal
  ✓︎ Prevents traversal to a non-existing file
  ✓︎ Compresses a string with gzip
```

After:
```
Foreign » FastGlob
  ✓︎ Prevents directory traversal
  ✓︎ Prevents symlink directory traversal
  ✓︎ Prevents traversal to a non-existing file
Foreign » Gzip » compress
  ✓︎ Compresses a string with gzip
```

app/src/App/Effect/Pursuit.purs
pete-murphy: Just passing `Buffer` directly gets this to pass the integration tests (https://github.com/purescript/registry-dev/pull/659/commits/e231be6da99ef301ba59407019f1366867b4232e). Before that, was seeing `"Not in GZIP format"` errors
```
registry # [   43.420167] registry-server-init[910]: Pushing to Pursuit...
registry # [   43.625450] registry-server-init[910]: [ERROR] Pursuit publishing failed with status 500 and body
registry # [   43.626327] registry-server-init[910]: {
registry # [   43.626781] registry-server-init[910]: "cause1":"java.util.zip.ZipException: Not in GZIP format",
registry # [   43.627595] registry-server-init[910]: "servlet":"com.github.tomakehurst.wiremock.servlet.WireMockHandlerDispatchingServlet-2dca0d64",
registry # [   43.629109] registry-server-init[910]: "cause0":"java.lang.RuntimeException: java.util.zip.ZipException: Not in GZIP format",
registry # [   43.634155] registry-server-init[910]: "message":"java.lang.RuntimeException: java.util.zip.ZipException: Not in GZIP format",
registry # [   43.635515] registry-server-init[910]: "url":"/packages",
registry # [   43.636876] registry-server-init[910]: "status":"500"
registry # [   43.637356] registry-server-init[910]: }
registry # [   43.637757] registry-server-init[910]: [ERROR] Could not publish your package to Pursuit because an error was encountered (cc: @purescript/packaging): Expected to receive a 201 status from Pursuit, but received 500 instead.
```

foreign/test/Foreign/Gzip.purs
pete-murphy: Replaced with `zcat` which seems to be available.

app/src/Fetch/Retry.purs
```diff
@@ -22,16 +22,17 @@ data RetryRequestError
   | StatusError Response
 
 withRetryRequest
-  :: forall input output thruIn thruOut headers
-   . Union input thruIn (HighlevelRequestOptions headers String)
+  :: forall input output thruIn thruOut headers @body
+   . ToRequestBody body
+  => Union input thruIn (HighlevelRequestOptions headers body)
```
pete-murphy: If we don't want to have this overloaded to work with any `ToRequestBody` (which then requires specifying `@String` in several places) we could have separate functions similar to the separation between [`fetch`](https://github.com/rowtype-yoga/purescript-fetch/blob/71bf92d8bbb5bad0a6b0647cd0f86a4b762bf977/src/Fetch.purs#L59) and [`fetchBody`](https://github.com/rowtype-yoga/purescript-fetch/blob/71bf92d8bbb5bad0a6b0647cd0f86a4b762bf977/src/Fetch.purs#L73) in `purescript-fetch`.

`withRetryRequest` and `withRetryRequestBody`?

flip111:
Hey i reviewed your PR looks good to me. There is an opportunity to tweak the compression algorithm if needed. The gzip function takes a second parameter with options and you can set the compression level there with other parameters https://nodejs.org/api/zlib.html#zlibgzipbuffer-options-callback

pete-murphy:
> Hey i reviewed your PR looks good to me. There is an opportunity to tweak the compression algorithm if needed. The gzip function takes a second parameter with options and you can set the compression level there with other parameters https://nodejs.org/api/zlib.html#zlibgzipbuffer-options-callback

Thanks! I have no experience optimizing compression so I have no idea what would be the best settings here and would be open to recommendations. I'm hoping that since (I think) there haven't been any issues publishing via Pulp using the default settings for `zlib.createGzip`, we might be safe using the defaults here 🤞 .

foreign/test/Foreign/Gzip.purs
pete-murphy: Maybe this is a bug in `purescript-spec`? I opened this issue https://github.com/purescript-spec/purescript-spec/issues/144

thomashoneyman:
I'm sorry for my slow reviews this week — I'm only somewhat available, and will be fully back online next week. I'm going to take a quick look through your PR but don't have the time right now to test it out. I promise to get this over the line by next week, however, should it appear to have no issues. Thanks for taking the time to work on this!

app/src/Fetch/Retry.purs
```diff
@@ -22,16 +22,16 @@ data RetryRequestError
   | StatusError Response
 
 withRetryRequest
-  :: forall input output thruIn thruOut headers
-   . Union input thruIn (HighlevelRequestOptions headers String)
+  :: forall input output thruIn thruOut headers @body
```
thomashoneyman: Is this VTA actually used? I don't see it in the body of the function.

foreign/src/Foreign/Gzip.purs
```diff
@@ -0,0 +1,33 @@
...
+import Node.Buffer (Buffer)
+
+foreign import compressImpl :: EffectFn3 String (EffectFn1 Error Unit) (EffectFn1 Gzip Unit) Unit
+
+foreign import toRequestBodyImpl :: Buffer -> RequestBody
+
+newtype Gzip = Gzip Buffer
+
+instance ToRequestBody Gzip where
+  toRequestBody (Gzip buffer) = toRequestBodyImpl buffer
```
thomashoneyman: I typically like to write functions like this one with `unsafeCoerce` (since it's a straightforward coercion) instead of a as a foreign import, because you can tell at a glance how it works instead of having to look at the FFI to see if there's anything interesting happening.

Would you mind reimplementing the foreign import as something like this?

```purs
-- could also be 'gzipToRequestBody :: Gzip -> RequestBody' and the instance:
--   instance ToRequestBody Gzip where
--     toRequestBody = gzipToRequestBody
coerceToRequestBody :: Buffer -> RequestBody
coerceToRequestBody = unsafeCoerce
```

thomashoneyman (commented):
This is a nice, small patch! It looks totally fine to me module some minor review comments. I might be able to test it tomorrow, in which case I'll get it merged. Thanks!

foreign/test/Foreign/Gzip.purs
thomashoneyman: We can use anything available in our `nixpkgs`, regardless of what's in the GitHub Actions runner. The tests have little available — just what's on the `ubuntu-latest`, plus whatever node modules have been installed via the package-lock.json. If we want to make something else available (like `gzcat`) then we can switch the below script to use nixpkgs' `runCommand` and manually add those dependencies to the `nativeBuildInputs`.

https://github.com/purescript/registry-dev/blob/30ebf38e8b4fbe685b65f502a1c51dcf413e2d06/flake.nix#L197-L208

It'd look something like replacing `writeShellScriptBin` with this:

```nix
pkgs.runCommand "run-tests" { nativeBuildInputs = [ pkgs.gzcat ]; } ''
  <script>
''
```

However, if zcat is available, it's nice to not have to touch this. Just wanted to give you some details — you don't need to change anything about your approach.

app/src/App/Effect/Pursuit.purs
thomashoneyman: The upstream is actually https://github.com/rowtype-yoga/purescript-fetch-core, and we can more easily make changes to that library. `fetch` relies on `fetch-core` instead of `web-fetch` (until one day we merge them into `js-fetch`).

thomashoneyman: Regardless — I think your Gzip newtype workaround is a proper solution. Good intuition!

foreign/test/Foreign/Gzip.purs
thomashoneyman: Hm — seems like it. I think what you've done in this PR is fine (adding the top `describe`).

app/src/Fetch/Retry.purs
pete-murphy: Before these changes, the `body` option of `withRetryRequest` was fixed to `String`, but I've replaced the definition of `fetch` below with `fetchBody` from the `purescript-fetch` library. 

Because `body` could now be any `ToRequestBody body => body`, I needed to specify `@String` in these call sites (the compiler didn't seem to infer as `String` without the VTA, from what I tried) https://github.com/purescript/registry-dev/blob/6677362a77d4e0bd7c017a592460a01f05b4795c/app/src/App/Effect/Pursuit.purs#L104 https://github.com/purescript/registry-dev/blob/6677362a77d4e0bd7c017a592460a01f05b4795c/app/src/App/Effect/Source.purs#L139 https://github.com/purescript/registry-dev/blob/6677362a77d4e0bd7c017a592460a01f05b4795c/app/src/App/Effect/Storage.purs#L247

pete-murphy: An alternative that comes to mind, to avoid adding those `@String` changes, would be keeping both `fetch` and `fetchWithBody` implementations copied from `purescript-fetch` and then adding a `retryWithRequestBody` so we don't have to update call sites of `retryWithRequest`

pete-murphy: > I needed to specify `@String` in these call sites (the compiler didn't seem to infer as `String` without the VTA, from what I tried)

Hm, nevermind, trying again without VTAs and type inference is working fine. I think at one point there was an explicit `ToRequestBody body` constraint that required the type annotation. I copied the constraint from the upstream definition of [`fetchBody`](https://github.com/rowtype-yoga/purescript-fetch/blob/71bf92d8bbb5bad0a6b0647cd0f86a4b762bf977/src/Fetch.purs#L75), but from what I can tell it's not really necessary.

pete-murphy: Removed here 0a02e14aaa133499bf17b698ab6cc336120e71f8

foreign/src/Foreign/Gzip.purs
pete-murphy: Updated here 74cca55945b539b05619353ddbf0b34ea10ad713 with a link to https://github.com/rowtype-yoga/purescript-fetch/issues/11

app/src/Fetch/Retry.purs
pete-murphy: Resolved here https://github.com/purescript/registry-dev/pull/659#discussion_r1315261368

thomashoneyman (approved):

thomashoneyman:
Thank you!

PR #660, opened 2023-09-13
Pass through dhall types directory to server

thomashoneyman:
The server was not setting DHALL_TYPES correctly because we didn't pass it through to the nix module. Now we do send through env vars that Nix knows about. This was preventing users from publishing their packages due to an error typechecking their manifest.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,13 +38,13 @@
       }
     }/Prelude/package.dhall";
 
+    # The location of the Dhall type specifications, used to type-check manifests.
     DHALL_TYPES = ./types;
 
     # We disable git-lfs files explicitly, as this is intended for large files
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -232,10 +232,6 @@
       in
         builtins.listToAttrs (builtins.map toKeyPair noComments);
 
-      # Print an attrset of env vars { ENV_VAR = "value"; } as a newline-delimited
-      # string of "ENV_VAR=value" lines.
-      printEnv = env: pkgs.lib.concatStringsSep "\n" (pkgs.lib.mapAttrsToList (name: value: "${name}=${value}") env);
```
thomashoneyman, line 237: This was unused.

f-f (approved):

PR #663, opened 2023-09-17
Support Windows globs

f-f:
Came across this one while working on https://github.com/purescript/spago/pull/995: it turns out that [fast-glob](https://www.npmjs.com/package/fast-glob) only supports forward slashes, so paths need to be converted for that.

The library exports a `convertPathToPattern` function, which I:
- added to the implementation of `match` so that the generated matches can take Windows paths
- exported from the library so that downstream can use it to fix their paths


f-f:
I'm starting to think this is not the best way to go at it, will put this on hold for now

PR #668, opened 2023-11-08
Fix performance issue in 'allDependencies'

thomashoneyman:
In #667 I added an implementation to prune unused dependencies from legacy packages. Unfortunately, while the implementation was tested in the small, it was never tested on a package with a sizable graph. On larger packages the inefficiencies in the implementation would cause the process to run out of memory.

This implementation fixes the issue. We switch to ST so we can keep a mutable map around to track modules we've already visited and push all unvisited dependencies to an array for later processing. Once we go through the full pending array without refilling it we're done and can exit.

### DIFF
--- a/lib/src/PursGraph.purs
+++ b/lib/src/PursGraph.purs
@@ -4,9 +4,11 @@ module Registry.PursGraph where
 
 import Prelude
 
+import Control.Monad.ST as ST
 import Data.Array as Array
 import Data.Array.NonEmpty (NonEmptyArray)
 import Data.Array.NonEmpty as NonEmptyArray
+import Data.Array.ST as Array.ST
 import Data.Bifunctor (bimap)
 import Data.Codec.Argonaut (JsonCodec)
 import Data.Codec.Argonaut as CA
@@ -19,11 +21,13 @@ import Data.Newtype (class Newtype, un)
 import Data.Profunctor as Profunctor
 import Data.Set (Set)
 import Data.Set as Set
-import Data.Traversable (traverse)
 import Data.Tuple (Tuple(..))
+import Foreign.Object as Object
+import Foreign.Object.ST as Object.ST
 import Node.Path (FilePath)
 import Registry.Internal.Codec as Internal.Codec
 import Registry.PackageName (PackageName)
+import Safe.Coerce (coerce)
 
 -- | A graph of the dependencies between modules, discovered by the purs
 -- | compiler from a set of source files.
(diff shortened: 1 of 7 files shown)


thomashoneyman:
Tested on this package run in the API:
https://github.com/purescript/registry/issues/410#issuecomment-1802255081

f-f (commented):
Code looks good, but I wonder why we don't have similar issues in Spago? I run the graph code against our work codebase, which is much bigger than the stuff the Registry goes through.

thomashoneyman:
The major difference is that Spago's implementation is all folds, but I was using `bind` via the call to `traverse`. I think this is a reasonable place to use ST — at least the way it's used in the registry, which is a bit different from what Spago is doing — but I can explore a different approach if you aren't happy with the use of ST.

lib/src/PursGraph.purs
```diff
@@ -78,13 +82,29 @@ associateModules parse graph = do
...
+  ST.while (map (_ > 0) (Array.ST.length pending)) do
+    Array.ST.shift pending >>= case _ of
+      Nothing -> pure unit
+      Just (ModuleName mod) -> Object.ST.peek mod visited >>= case _ of
+        Nothing -> do
+          void $ Object.ST.poke mod unit visited
+          case Map.lookup (ModuleName mod) graph of
+            Nothing -> pure unit
+            Just { depends } -> void $ Array.ST.pushAll depends pending
+        Just _ -> pure unit
```
natefaubion, line 106: This is equivalent to writing a recursive function with an explicit `List` stack and accumulator, FWIW.

natefaubion: For example, https://github.com/aristanetworks/purescript-backend-optimizer/blob/bdef8b58cd8591bef7fd499c892989604e6b918d/src/PureScript/Backend/Optimizer/CoreFn/Sort.purs#L33-L47

thomashoneyman: Happy to write it that way if you have an example somewhere. I'm curious what the generated code will look like vs. the ST version.

natefaubion: I don't have an opinion. One is obviously using mutation, so I would expect this to be significantly faster. I just don't use ST in core backend-optimizer because it's not very portable.

thomashoneyman: In this case since we don't need portability in the registry and speed does matter (this is part of the pipeline) I would prefer to stay with ST. The generated code looks quite good.

thomashoneyman:
@f-f could you take another look at this? it's still [blocking publishing](https://github.com/purescript/registry/issues/411#issuecomment-1803070763).

f-f:
Ah yes sorry I had the tab open and got distracted with other things - my preference is that we keep the lib portable if we can. We explicitely depend on some node libs though, so I guess this is alright as well?

f-f (approved):

thomashoneyman:
Yeah, we already explicitly depend on some NPM libraries (ssh2), and Node builtins (the crypto module), so if we were to make the library portable we'd need to move that stuff out altogether and at that time could also replace the ST code. But we'd only need to do any of that if someone came along wanting to use the registry-lib with an alternate backend.

f-f:
I assume we'll want to use the Scheme backend with Spago (which depends on the registry-lib) once that takes off, but we'll sorry about it then

PR #671, opened 2023-11-17
Update Node deps to breaking changes

JordanMartinez:
Currently fails with these errors when I call `spago build`. I didn't look into the 2nd/3rd error because I'm blocked by the first one:
```
[1/3 TypesDoNotUnify] .spago/packages/spago-core/47b735c5845a084f40826ae1989c3dc6c9171f21/core/src/Log.purs:161:23

  161    Effect.liftEffect $ Process.exit 1
                             ^^^^^^^^^^^^^^

  Could not match type
    Effect
  with type
    Function Int
  while trying to match type Effect t1
    with type Int -> t2
  while checking that expression exit 1
    has type Effect t0
  in value declaration die'
  where t0 is an unknown type
        t1 is an unknown type
        t2 is an unknown type

[2/3 TypesDoNotUnify] app/src/Fetch/Retry.purs:73:51

  73    coreResponse <- Promise.Aff.toAffE $ Core.fetch request
                                                        ^^^^^^^

  Could not match type
    Request
  with type
    Request
  while checking that type Request
    is at least as general as type Request
  while checking that expression request
    has type Request
  in value declaration fetch

[3/3 TypesDoNotUnify] foreign/src/Foreign/Gzip.purs:27:19

  27    toRequestBody = gzipToRequestBody
                        ^^^^^^^^^^^^^^^^^

  Could not match type
    RequestBody
  with type
    RequestBody
  while checking that type Gzip -> RequestBody
    is at least as general as type t0 -> RequestBody
  while checking that expression gzipToRequestBody
    has type t0 -> RequestBody
  where t0 is an unknown type

           Src   Lib   All
Warnings     0     0     0
Errors       2     1     3

❌ Failed to build.
```


thomashoneyman:
@JordanMartinez this should be unblocked now (no more spago-core dependency)

JordanMartinez:
Yup! Thanks!

JordanMartinez:
CI builds!

spago.yaml
```diff
@@ -145,4 +107,4 @@ workspace:
         - typelevel-prelude
         - unsafe-coerce
         - untagged-union
-
+    fetch: 4.0.0
```
thomashoneyman: This is already in the target package set:
https://github.com/purescript/registry/blob/b035b8ebfb29d65b40efbea9edffe9a5e539dedc/package-sets/46.0.0.json#L114

```suggestion
```

thomashoneyman: So is `httpurple`, by the way, so we can remove that from extra_packages

thomashoneyman: Huh I don't know how this comment got onto _this_ commit! I must have been looking at an earlier diff. I see you deleted `httpurple` so that's all good.

thomashoneyman (approved):
Other than removing `fetch` from extra_packages, looks good to me!

spago.yaml
JordanMartinez: I think I included `fetch` initially because I was getting some other error (probably because I hadn't migrated away from `fetch-core` yet?). Either, way, thanks for taking care of this!

Issue #672, opened 2023-11-23
Proposal: authenticated publish endpoint


thomashoneyman:
The registry currently offers a single endpoint for publishing packages, `/publish`. Anyone can send a package name, location, and ref to publish the package. The registry will fetch the source, validate, it, publish, and push the docs to Pursuit.

When the registry fetches the source it will also pick up the manifest file for the package. The registry supports a `purs.json` file and it also has first-class support for a `spago.yaml` file. It has first-class support for `spago.yaml` so that Spago users don't have to generate and commit a `purs.json` file.

It's important that users don't have to commit an extra file because, as we've seen with the old `spago.dhall` / `bower.json` pairs, the file that is generated for publishing almost inevitably falls out of sync with the primary package manager file. In the publishing context this means either publishing with an unexpected, out-of-date manifest, or failing altogether and having to tag a new version just to fix the out-of-sync manifest.

We don't want the registry to have to special case a bunch of different potential package manager file formats. It was a stretch to even special-case the spago.yaml format. But I think it's a shame that package managers other than Spago will be stuck having to generate and commit a purs.json file as part of publishing. Instead, I'd like to propose an alternative: an authenticated `/publish` endpoint which allows you to send a manifest in the publish payload, rather than requiring it exist in the source code.

**Authenticated Publish**

The authenticated publish endpoint would be the same as the normal endpoint, with a new field: `manifest`.

```purs
type AuthPublishData =
  { name :: PackageName
  , location :: Maybe Location
  , ref :: String
  , compiler :: Maybe Version
  , resolutions :: Maybe (Map PackageName Version)
  -- NEW
  , manifest :: Manifest
  }
```

The purpose of this endpoint is for package managers to publish packages without having to commit a `purs.json` file and without the registry having to special-case their particular manifest format. Instead, so long as they can generate a proper `Manifest` then the registry can accept their code.

The reason the endpoint has to be authenticated is because we consider the non-authenticated API untrusted input, and we want users to have full control over the contents of their manifests. An arbitrary user shouldn't be able to just send a manifest to the registry and publish it.

**Internal Changes**

The new endpoint requires almost no changes to the registry. We would simply authenticate, then go through the normal `publish` workflow, except that if we already have a `Manifest` then we don't read one from disk. (Or we throw an exception if you sent a manifest, and have one committed, and they don't match.) Similar to how we branch for `spago.yaml` files.

**Considerations**

[...]

PR #673, opened 2023-11-24
Reimplement spago to manifest code

JordanMartinez:
Fixes the first part of https://github.com/purescript/spago/issues/1122#issuecomment-1823047315:
> 1. Reimplement spago.yaml parsing in the registry (just the sections we care about)


JordanMartinez:
I'm guessing I need to add `yaml` as an NPM dependency somewhere, but not sure how to do that.

app/src/App/PackageManagers/Spago.purs
```diff
@@ -0,0 +1,164 @@
...
+      , includeFiles
+      , excludeFiles
+      }
+
+-- | Parse a JSON string, constructing the `Toml` value described by the string.
+-- | To convert a string into a `Toml` string, see `fromString`.
+yamlParser :: String -> Either String Core.Json
+yamlParser j = runFn3 yamlDocParserImpl Left Right j
+
+foreign import yamlDocParserImpl :: forall a. Fn3 (String -> a) (Core.Json -> a) String a
```
thomashoneyman: We should put the YAML parsing in the `foreign` subdirectory along with our other FFI bindings

app/src/App/API.purs
```diff
@@ -14,6 +14,7 @@ module Registry.App.API
 
 import Registry.App.Prelude
 
+import App.PackageManagers.Spago as Spago
```
thomashoneyman: If we're going to support several "decoders", then perhaps we should name the modules `App.Manifest.X` where `X` is the manifest type. In this case, that'd be `App.Manifest.Spago`. Alternately, I'd be happy with a single top-level module called `App.Manifest` that has functions for decoding different ones. Kind of like how  we have `App.Legacy.Manifest`, but these would be ones we plan to support beyond the legacy importer.

The reason I'm hesitant about `App.PackageManagers` is that I don't think we plan to do anything specific to a package manager besides parsing the manifest format.

app/src/App/PackageManagers/Spago.js
```diff
@@ -0,0 +1,12 @@
+import Yaml from 'yaml';
```
thomashoneyman: Let's move this to `foreign/src/yaml.js` instead, as we keep our FFI bindings there in general.

app/src/App/PackageManagers/Spago.purs
```diff
@@ -0,0 +1,164 @@
...
+    widestRange :: Range
+    widestRange = Either.fromRight' (\_ -> unsafeCrashWith "Fake range failed")
+      $ Range.parse ">=0.0.0 <2147483647.0.0"
+
+    spagoRangeCodec :: JsonCodec Range
+    spagoRangeCodec = CA.prismaticCodec "SpagoRange" rangeParse printSpagoRange CA.string
+      where
+      rangeParse str =
+        if str == "*" then Just widestRange
+        else hush $ Range.parse str
```
thomashoneyman: We should move all these codecs to the top level so that they can be tested with various spago.yaml configs and we can be ready to write migrations if the format changes.

The same is true, really, of all the functions in this `where` clause; I think they should be lifted to the top level so we can at least potentially test them.

app/src/App/PackageManagers/Spago.purs
```diff
@@ -0,0 +1,164 @@
...
+import Registry.PackageName as PackageName
+import Registry.Range (Range)
+import Registry.Range as Range
+import Registry.Version as Version
+import Run (AFF, EFFECT, Run)
+import Run.Except (EXCEPT)
+import Run.Except as Except
+
+getSpagoManifest :: forall r. FilePath -> Run (LOG + EXCEPT String + AFF + EFFECT + r) Manifest
+getSpagoManifest spagoYamlPath = do
```
thomashoneyman: I think we should have functions to parse a spago manifest into a `SpagoYAML` type and another to convert it into a `Manifest` type, but actually reading the files from disk and throwing errors should be deferred to whomever is using these functions.

We often need to write different error messages depending on the circumstances in which we're parsing or converting and once we get into `Run` it's a bigger hassle to test these functions vs. being pure. I'm thinking:

```purs
parseSpagoManifest :: YamlString -> Either String SpagoYaml

convertSpagoManifest :: SpagoYaml -> Either String Manifest

spagoYamlCodec :: JsonCodec SpagoYaml

dependenciesCodec :: JsonCodec SpagoDependencies
```

...and so on. It's not too much work for someone else to read the YAML, feed it to the parser, and then convert to a `Manifest` and write their own error messages along the way.

app/src/App/PackageManagers/Spago.purs
```diff
@@ -0,0 +1,164 @@
...
+      [] -> Right (Map.fromFoldable success)
+      errs -> Left $ "The following packages did not have their ranges specified: " <> String.joinWith ", " (map PackageName.print errs)
+    pure $ Manifest
+      { version
+      , license
+      , name
+      , location
+      , description
+      , dependencies
+      , owners: Nothing -- TODO Spago still needs to add this to its config
```
thomashoneyman: Wonder if we could actually parse this out if we know what Spago _will_ do, even if they don't do it now. After all, if it's missing right now it'll just parse to `Nothing`.

thomashoneyman (changes requested):

thomashoneyman:
Thanks for taking this on, @JordanMartinez! I have a few comments, and @f-f should take a look, but on the whole I think it's a good direction. I already commented on some of this, but we should add tests to the `app` test that parse yaml strings into the spago.yaml config format so we have some verification that our parsing and errors are good. We could also consider adding tests to the `foreign` workspace that just verify we're parsing YAML strings correctly so that if we update the yaml library in the future we've got assurances it's all still working.

JordanMartinez:
Since I opened this PR before I read over https://github.com/purescript/registry-dev/issues/672, I'm actually thinking this is the wrong approach to take and my idea about storing `purs.json` file within the `git tag`'s annotation would be a better solution.

The concerns I have here are:
- needing to depend on a backend-specific library (e.g. yaml).
- out-of-sync issues with the package manager's config file. While I copied over this code and refactored it a bit here, this could still be slightly different than what Spago does.

I'd rather have the package managers consume their own config files and produce a `purs.json` file for us via some `api` subcommand (e.g. `spago api spago-to-purs`). However, that requires us to depend on (possibly multiple versions of) different package managers. So, this is also not ideal.

f-f:
> I'm actually thinking this is the wrong approach

I think this approach is by far the least disruptive, and my understanding is that we have reached a consensus that this is a good way to move forward. (see PR comments from Thomas anticipating having multiple config parsers explicitly)

Re your concerns:
1. The registry is already backend-specific and linux/macOS specific, we are not aiming for wide compatibility at this point in time
2. we'd figure out mismatches with the upstream pretty quickly as users will have their packages rejected if the parsing is not correct. Moreover, having the original config and the parser implemented here allows us to eventually reimport the package (this would go via a Trustee patch, see #80)

> I'd rather have the package managers consume their own config files and produce a purs.json file for us via some api subcommand

I'd rather not. As I tried to make clear in the whole #672 thread, I am of the opinion that passing data out-of-band is a terrible idea and a good way to dig a ditch similar to the one we're trying to climb out of with the Registry.

f-f:
Re this PR: it looks good so far, and I think it's good to go once Thomas' comments are addressed

thomashoneyman:
> > I'd rather have the package managers consume their own config files and produce a purs.json file for us via some api subcommand
> 
> I'd rather not. As I tried to make clear in the whole https://github.com/purescript/registry-dev/issues/672 thread, I am of the opinion that passing data out-of-band is a terrible idea and a good way to dig a ditch similar to the one we're trying to climb out of with the Registry.

I think @JordanMartinez meant that we support Spago configs by relying on `spago` directly in the registry, and we call `spago api spago-to-purs` to generate a purs.json file rather than implement codecs in the registry code.

That said, the Spago config is simple enough that I prefer to decode it directly rather than use a CLI. (Not that this would be the case for _all_ configs we can handle; we shell out to Dhall, for example, for spago.dhall files.)

> [concern:] needing to depend on a backend-specific library (e.g. yaml).

As @f-f said, I don't consider this a problem. I can see an argument for making the registry library backend-independent, since it can be depended on directly by package managers, but the registry application explicitly runs on Node on Linux. We already bind to plenty of Node-specific libraries for that reason.

> > [concern:] out-of-sync issues with the package manager's config file. 
> 
> we'd figure out mismatches with the upstream pretty quickly as users will have their packages rejected if the parsing is not correct. Moreover, having the original config and the parser implemented here allows us to eventually reimport the package

I agree with @f-f here that I'm not super worried about mismatches in the config parsing. If we can't parse the config then the package is rejected and it's up to the package manager to get a patch merged that fixes the config; they can always fall back to committing a purs.json file in the meantime. I'm struggling to envision a way our parsing would succeed, but the resulting manifest would be different from what the author intended.

JordanMartinez:
Ok, but I'll have to get to this another time. Likely not today.

thomashoneyman:
Sounds good, @JordanMartinez. I'm happy to pair on this to get it through quickly, if you'd like.

thomashoneyman:
Here's the tracking for adding `files` (now includes/excludes) and `owners` to our spago.yaml parsing — https://github.com/purescript/registry-dev/issues/594

app/src/App/API.purs
```diff
@@ -30,7 +30,6 @@ import Data.Set as Set
 import Data.Set.NonEmpty as NonEmptySet
 import Data.String as String
 import Data.String.CodeUnits as String.CodeUnits
-import Data.String.NonEmpty (fromString) as NonEmptyString
 import Data.String.NonEmpty.Internal (toString) as NonEmptyString
```
thomashoneyman: This is from before your code, of course, but could you go ahead and fix this import?

```suggestion
import Data.String.NonEmpty as NonEmptyString
```

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
+module Registry.App.Manifest.Spago
+  ( getSpagoManifest
```
thomashoneyman: I want to test the functions in this module (especially testing the parsing code against known-good spago.yaml files) so we need to export them. We can hide stuff like `widestRange` but all the codecs and the reading, parsing, converting functions should be exported.

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+import Registry.PackageName (PackageName)
+import Registry.PackageName as PackageName
+import Registry.Range (Range)
+import Registry.Range as Range
+import Registry.Version as Version
+import Run (AFF, EFFECT, Run)
+import Run.Except (EXCEPT)
+import Run.Except as Except
+
+getSpagoManifest :: forall r. FilePath -> Run (LOG + COMMENT + EXCEPT String + AFF + EFFECT + r) Manifest
```
thomashoneyman: We shouldn't have this function in this file. This file should only deal with parsing spago.yaml files to `SpagoYaml` and converting `SpagoYaml` to `Manifest`; think of it as a library module, whereas `API.purs` is more of an app module. We shouldn't be using things like `COMMENT` in here and ideally we'd be using as few effects as possible.

I'm OK with moving this wholesale to API.purs as a helper function or inlining it into `publish`.

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+          Comment.comment $ Array.fold
+            [ "Converted your spago.yaml into a purs.json manifest to use for publishing:\n"
+            , "```json"
+            , printJson Manifest.codec manifest
+            , "```"
+            ]
+          pure manifest
+
+readConfig :: forall r. FilePath -> Run (LOG + COMMENT + EXCEPT String + AFF + EFFECT + r) (Either String SpagoYaml)
+readConfig path = liftAff do
```
thomashoneyman: Can we rename this to `readSpagoYaml`? That's what we typically name functions like this. 

Also, this is listing several effects not in use in the function. You've listed `EXCEPT`, but no exceptions are thrown; likewise, `LOG` and `COMMENT` are not used. It seems like this whole thing could just be writen in `Aff` directly, without any use of `Run`:

```suggestion
readConfig :: forall m. MonadAff m => FilePath -> m (Either String SpagoYaml)
readConfig path = liftAff do
```

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+  liftEffect (FS.exists path) >>= case _ of
```
thomashoneyman: I think it's almost certain we've already checked for the path's existence before we try to read the file, and the `Aff.attempt` below will return a `Left` if the path doesn't exist, so I think it's OK to skip this additional `exists` call.

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+    false -> pure (Left $ "Did not find " <> path <> " file.")
+    true -> do
+      result <- Aff.attempt $ FS.Aff.readTextFile UTF8 path
```
thomashoneyman: Can we add a `readYamlFile` utility that mirrors the existing `readJsonFile` utility and just use that here? I think any time we add a format like this we should add `read*` / `write*` functions to the prelude.

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+      pure do
+        yamlStr <- lmap Aff.message result
+        docAsJson <- lmap (append "YAML: ") (Yaml.yamlParser yamlStr)
+        parseSpagoManifest docAsJson
+
+parseSpagoManifest :: Json -> Either String SpagoYaml
+parseSpagoManifest = lmap CA.printJsonDecodeError <<< CA.decode spagoYamlCodec
```
thomashoneyman: We should refer to functions that operate on `SpagoYaml` with that name rather than `SpagoManifest`.

```suggestion
parseSpagoYaml :: Json -> Either String SpagoYaml
parseSpagoYaml = lmap CA.printJsonDecodeError <<< CA.decode spagoYamlCodec
```

Also, we typically use `parse` to refer to producing something structured from a `String` (for example, see `Manifest.parse` or `Range.parse` or `Version.parse`), so I would prefer a `parse` function like:

```suggestion
parseSpagoYaml :: String -> Either String SpagoYaml
parseSpagoYaml str = do 
  json <- Argonaut.jsonParser str
  lmap CA.printJsonDecodeError $ CA.decode spagoYamlCodec json
```

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+            Maybe
+              { version :: Version
+              , license :: License
+              , location :: Maybe Location
+              , include :: Maybe (Array String)
+              , exclude :: Maybe (Array String)
+              , owners :: Maybe (NonEmptyArray Owner)
+              }
+        }
+  }
```
thomashoneyman: Can we break this down such that each record has its own type synonym? We can do the same with the codecs if you want (one codec per nested record), as we do elsewhere in the app.

```suggestion
type SpagoYaml = { package :: Maybe SpagoPackageConfig }

type SpagoPackageConfig = 
  { name :: PackageName
  , description :: Maybe String
  , dependencies :: Map PackageName (Maybe Range)
  , publish :: Maybe SpagoPublishConfig
  }

type SpagoPublishConfig =
  { version :: Version
  , license :: License
  , location :: Maybe Location
  , include :: Maybe (Array String)
  , exclude :: Maybe (Array String)
  , owners :: Maybe (NonEmptyArray Owner)
  }
```

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+widestRange :: Range
+widestRange = Either.fromRight' (\_ -> unsafeCrashWith "Fake range failed")
+  $ Range.parse ">=0.0.0 <2147483647.0.0"
+
+spagoRangeCodec :: JsonCodec Range
+spagoRangeCodec = CA.prismaticCodec "SpagoRange" rangeParse printSpagoRange CA.string
+  where
+  rangeParse str =
+    if str == "*" then Just widestRange
+    else hush $ Range.parse str
```
thomashoneyman: I'm not so sure about this. When would we just get '*' for a range? Packages should always specify ranges. I think the registry should reject that altogether instead of producing a bogus wide range. cc: @f-f.

thomashoneyman: (I know this was already in the registry and you didn't write it originally, I just noticed it in this review.)

f-f: This comes from Spago's codebase - Spago will use the `*` for the ranges of packages that are not coming from the registry (and rightfully so, we don't have any solver information for those. They are local packages, or packages fetched from git, and so on)

I think we all agree that the Registry will refuse to deal with those, and I guess you're leaning towards rejecting them at parse time?

thomashoneyman: I see. We don't need to reject them at parse time (errors will be worse), but we should reject them when we convert `SpagoYaml` to `Manifest`.

I don't think we should ever convert them to a `Range`, though (no `widestRange`). Instead, I'd rather we use an explicit data type, something like this:

```purs
data SpagoRange = Unbounded | HasRange Registry.Range
```

Then, when we convert to a `Manifest` we can verify that anything that's `Unbounded` is collected as an error, something like "<n> dependencies do not specify a range: <x>, <y>, <z>"

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+
+    decode json =
+      map Left (CA.decode PackageName.codec json)
+        <|> map Right (CA.decode packageSingletonCodec json)
+
+widestRange :: Range
+widestRange = Either.fromRight' (\_ -> unsafeCrashWith "Fake range failed")
+  $ Range.parse ">=0.0.0 <2147483647.0.0"
+
+spagoRangeCodec :: JsonCodec Range
```
thomashoneyman: Can we make a dedicated `SpagoRange` data type that admits "no range" as a possibility (or just parse to `Maybe Range`) instead of parsing missing ranges into the widest range possible?

See: https://github.com/purescript/registry-dev/pull/673/files#r1410003772

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+  packageSingletonCodec = Internal.Codec.packageMap spagoRangeCodec
+
+  to :: Map PackageName (Maybe Range) -> Array (Either PackageName (Map PackageName Range))
+  to deps =
+    map
+      ( \(Tuple name maybeRange) -> case maybeRange of
+          Nothing -> Left name
+          Just r -> Right (Map.singleton name r)
+      )
+      $ Map.toUnfoldable deps :: Array _
```
thomashoneyman: Is this annotation needed given that `Array` is in the type signature?

app/src/App/Manifest/Spago.purs
```diff
@@ -0,0 +1,190 @@
...
+
+  from :: Array (Either PackageName (Map PackageName Range)) -> Map PackageName (Maybe Range)
+  from = Map.fromFoldable <<< map
+    ( case _ of
+        Left name -> Tuple name Nothing
+        Right m -> map Just $ unsafeFromJust (List.head (Map.toUnfoldable m))
```
thomashoneyman: Could we add a comment to `from` that explains why this is safe? Why can we guaranteed do this `unsafeFromJust`?

app/src/App/Manifest/Spago.purs
f-f: > Then, when we convert to a Manifest we can verify that anything that's Unbounded is collected as an error, something like " dependencies do not specify a range: , , "

This made me realise that if packages don't go through `spago publish` then we are not going to reject the ones that depend on stray git repos, as long as the names are found in the registry (and they could refer to entirely different packages) and the ranges are legit-looking to the registry.
This is going to happen for all packages coming through the Github action.
Not sure what we should do about those?

In the general case of supporting arbitrary formats this might come up again as well: how do we come up with sensible dependency information without parsing and running the full interpreter

thomashoneyman: If they list a dependency name and range then we'll assume they mean the registry, and since we solve and compile the package as part of publishing the only way publishing would succeed is if the dependency name/range in the registry happens to work.

That seems fine to me — alternately, we could parse your package set additions / overrides if you're using them? Or we could require that you aren't using a package set for your library? Not sure what's best.

f-f: Yeah it's a bit thorny. I think we can go ahead and assume that the manifest refers to registry packages (except for cases like the `*` here, which surely means the package is not from the registry), and figure something out once we stumble on a problem with this approach

thomashoneyman (approved):

thomashoneyman:
cc: @JordanMartinez this is ready for a look if you'd like to verify it.

JordanMartinez:
@thomashoneyman Looks good to me!

PR #676, opened 2023-11-28
Write manifest files before calling purs publish

### LINKED ISSUE #675: Write purs.json before calling Pursuit publish on already published package version
See https://github.com/purescript/registry/issues/422#issuecomment-1825929900 for an example of this error.

In #667 we started to prune unused dependencies from manifests. Accordingly, we stopped immediately writing manifests and instead read them so they can (potentially) be manipulated before they are written. Here's where we read the manifest and previously wrote it:
https://github.com/purescript/registry-dev/blob/1d9e28b36964fbbb3e7918ee486adbd0ac0e6a96/app/src/App/API.purs#L402

Later, we check if the manifest package / version has been published to the registry, but the docs haven't been published to pursuit:
https://github.com/purescript/registry-dev/blob/1d9e28b36964fbbb3e7918ee486adbd0ac0e6a96/app/src/App/API.purs#L498

If so, we eventually move forward with publishing to Pursuit:
https://github.com/purescript/registry-dev/blob/1d9e28b36964fbbb3e7918ee486adbd0ac0e6a96/app/src/App/API.purs#L520

However, at this time, we haven't actually attempted to change the manifest and haven't written it to the package source directory. That doesn't happen until later — for example:
https://github.com/purescript/registry-dev/blob/1d9e28b36964fbbb3e7918ee486adbd0ac0e6a96/app/src/App/API.purs#L550-L563

This causes Pursuit publishing to fail in the case there was no purs.json file, as `purs publish` will exit. To fix this, we need to either:

1. Move the 'already published' check to the end of the pipeline, after we have potentially modified and written the new purs.json file, or
2. If we have already published the package version, skip any further processing and read the manifest we previously produced from the registry index, aborting if it isn't there.

The advantage of (1) is that it's very easy to implement. The disadvantage is that if a package is published at one time, and then the registry checks or modifications change, then if the package docs are attempted later on then it's possible the package manifest used for publishing is different (or fails altogether) vs. the original we packaged.

The advantage of (2) is that we skip directly to reusing the manifest we already computed and publish that — no chance of a mismatch, and more efficient. If we're OK with just reading from the registry index then it's also quite easy to implement. If we want to handle the possibility that the manifest is missing in the registry index then it's a little more work (we'd have to download and unpack the tarball we published to get the manifest out of it).

My suggestion is that we use approach (2), reading the already-published manifest from the registry index (unless the source code uses a purs.json, in which case we can just use that). If the manifest cannot be found in the index, abort with an error.

As a stretch goal we should add to the "integration" test in the PureScript tests such that we verify this code path:
https://github.com/purescript/registry-dev/blob/1d9e28b36964fbbb3e7918ee486adbd0ac0e6a96/app/test/App/API.purs#L71

thomashoneyman:
Fixes #675. Specifically, this PR includes two changes:

1. We make sure to write a purs.json file to disk before calling `publishToPursuit`, unless the package already contains a purs.json file. The file we write is taken from the manifest index rather than what we've generated in the pipeline run so as to avoid any mismatches with what we published before. This is especially important for packages where we pruned unused dependencies
2. We were still exercising the rest of the publish pipeline after publishing to Pursuit, ie. uploading to storage and so on. We shouldn't do that — if a package needs to be published to Pursuit only, then we should do that then terminate. This PR fixes that.

I've added a test verifying the behavior. Most of the diff is just indentation.


app/test/App/API.purs
```diff
@@ -70,7 +70,17 @@ spec = do
 
   Spec.describe "API pipelines run correctly" $ Spec.around withCleanEnv do
     Spec.it "Publish a legacy-converted package with unused deps" \{ workdir, index, metadata, storageDir, githubDir } -> do
-      let testEnv = { workdir, index, metadata, username: "jon", storage: storageDir, github: githubDir }
+      let
+        testEnv =
+          { workdir
+          , index
+          , metadata
+          , pursuitExcludes: Set.singleton (Utils.unsafePackageName "type-equality")
```
thomashoneyman, line 78: The Pursuit mock effect reads from the metadata ref, so this allows us to manually exclude packages that are present in fixtures but we want _not_ present on Pursuit.

app/src/App/API.purs
```diff
@@ -495,173 +496,184 @@ publish source payload = do
...
-  for_ (Operation.Validation.isNotPublished (Manifest manifest) (Metadata metadata)) \info -> do
-    -- If the package has been published already, then we check whether the published
-    -- version has made it to Pursuit or not. If it has, then we terminate here. If
-    -- it hasn't then we skip to Pursuit publishing.
-    published <- Pursuit.getPublishedVersions manifest.name >>= case _ of
-      Left error -> Except.throw error
-      Right versions -> pure versions
+  publishPursuitOnly <- case Operation.Validation.isNotPublished (Manifest manifest) (Metadata metadata) of
+    Nothing -> pure false
+    Just info -> do
```
thomashoneyman, line 513: This diff changes no behavior until line 535. Just indentation and a switched case.

app/src/App/API.purs
```diff
@@ -495,173 +496,184 @@ publish source payload = do
...
+              unless hadPursJson do
+                existingManifest <- ManifestIndex.readManifest manifest.name manifest.version
+                case existingManifest of
+                  Nothing -> Except.throw "Version was previously published, but we could not find a purs.json file in the package source, and no existing manifest was found in the registry."
+                  Just existing -> Run.liftAff $ writeJsonFile Manifest.codec packagePursJson existing
+              publishToPursuit { packageSourceDir: packageDirectory, compiler: payload.compiler, resolutions: verifiedResolutions, dependenciesDir }
+
+    -- In this case the package version has not been published, so we proceed
+    -- with ordinary publishing.
+    Nothing ->
```
thomashoneyman, line 556: Previously we ran this code even if the package version was already published and just needed its docs published. But we shouldn't — it should only ever run if the package version is new. 

All changes from this point onward are just indentation.

app/test/App/API.purs
f-f: I am slightly confused about this field even after your explanation - could we add some docs about it? (presumably in `Registry.Test.Assert.Run`)

f-f (approved):
Code looks good, just left a comment asking for more comments 😄

app/test/App/API.purs
thomashoneyman: Does this help clear things up?
https://github.com/purescript/registry-dev/pull/676/commits/6245a2f48c9c0cb3bb11b17a20986c66746767d1

f-f:
Lovely, thanks!

PR #677, opened 2023-11-28
Log generated manifest if converted from legacy or integrated manifest type

### LINKED ISSUE #666: Show generated manifest in logs when converting from legacy
When a user encounters an issue with the manifest we generate on their behalf from a bower.json or spago.dhall file, they don't know what the actual generated manifest contains, which makes debugging more difficult. For example, in this issue there is a solver error with the generated manifest from a spago.dhall file, but since the manifest isn't included, we can't verify the build plan that got used:

https://github.com/purescript/registry/issues/405#issuecomment-1787698557

When we generate a manifest ("Generating a manifest from your bower.json and/or spago.dhall files...") we should notify the user of the manifest we produced for their package.

thomashoneyman:
Fixes #666. If we did not use the manifest in your source code directly (e.g. we converted a spago.yaml, spago.dhall, bower.json, etc. to a purs.json) then we explicitly show it to you before continuing. Since any errors wrt dependencies will be based on the generated purs.json file this helps narrow down any issues.

### DIFF
--- a/app/src/App/API.purs
+++ b/app/src/App/API.purs
@@ -435,6 +435,12 @@ publish source payload = do
               ]
             Right manifest -> do
               Log.debug "Successfully converted a spago.yaml into a purs.json manifest"
+              Comment.comment $ Array.fold
+                [ "Converted your spago.yaml into a purs.json manifest to use for publishing:\n"
+                , "```json"
+                , printJson Manifest.codec manifest
+                , "```"
+                ]
               pure manifest
     else do
       Comment.comment $ "Package source does not have a purs.json file. Creating one from your bower.json and/or spago.dhall files..."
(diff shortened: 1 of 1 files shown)


app/src/App/API.purs
```diff
@@ -457,10 +463,14 @@ publish source payload = do
...
           let manifest = Legacy.Manifest.toManifest payload.name version existingMetadata.location legacyManifest
+          Comment.comment $ Array.fold
+            [ "Converted your legacy manifest(s) into a purs.json manifest to use for publishing:\n"
+            , "```json"
+            , printJson Manifest.codec manifest
+            , "```"
+            ]
           pure manifest
 
-  Comment.comment "Verifying package..."
```
thomashoneyman, line 462: At this point we're well underway in verification so I feel like it's not particularly helpful. I can put it back if we need it.

f-f (approved):
Looks great!

PR #683, opened 2024-01-19
Include additional VCS directories to ignore

toastal:
npm’s list is incomplete for VCSs. Since they can sometimes execute scripts without the users knowledge & since only the source, not the entire history, is needed, it should be safe to ignore them. Additionally clarified these as a separate bullet point in the spec.


toastal:
If anyone wants to accept work on supporting some of these, you have my attention :)

thomashoneyman (approved):
The idea is to ignore VCS in general, so :+1:

JordanMartinez (approved):

PR #689, opened 2024-05-29
bump char limit on package names

roryc89:
Increase char limit on package names  https://github.com/purescript/spago/issues/1086\#issuecomment-1764101919


f-f (approved):
Thanks!

PR #690, opened 2024-06-07
Port from codec-argonaut to codec-json

f-f:
Not sure if we are tracking this in this repo, but we talked about doing this in the past, and we need it for https://github.com/purescript/spago/issues/1090

Maybe the most important detail here is that we are propagating the errors from the parsers into the `DecodeError`s, so that they get through to the user when parsing things like PackageNames that are too long.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ result
 
 # Keep it secret, keep it safe.
 .env
+.envrc
--- a/app/spago.yaml
+++ b/app/spago.yaml
@@ -8,11 +8,11 @@ package:
   dependencies:
     - aff
     - ansi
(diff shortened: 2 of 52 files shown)


thomashoneyman:
At a glance this all looks like a straightforward, mechanical change. I guess we don’t have much in the way of error tests; what does the new error output look like when for example a JSON payload contains an invalid package name?

f-f:
@thomashoneyman good point - I added a test to exercise the new errors!

thomashoneyman (approved):
(Other than the formatter issue — looks good!)

f-f:
Ah yeah, autoformatter didn't run for some reason

PR #694, opened 2024-07-21
`instance Show` for `PackageName`

peterbecich:
I think this will assist with any debugging that uses `PackageName`, for instance in Spago: https://github.com/purescript/spago/pull/1253

### DIFF
--- a/lib/src/PackageName.purs
+++ b/lib/src/PackageName.purs
@@ -50,6 +50,7 @@ stripPureScriptPrefix pkg =
 -- | A Registry-compliant package name
 newtype PackageName = PackageName String
 
+derive newtype instance Show PackageName
 derive newtype instance Eq PackageName
 derive newtype instance Ord PackageName
 


f-f:
Show instances were removed in #538, and while I sometimes do miss them myself, I think that having `Debug` instances would be best. But @thomashoneyman might have more input on this

We have [`unsafeLog`](https://github.com/purescript/spago/blob/3c2b1ce92a5d9d1edbd236181eb821cabafd9c6a/src/Spago/Prelude.purs#L58) for quickly printing things out in Spago, and that has been enough for my dev needs

thomashoneyman:
I'm happy to add `Debug` instances. We ran into several issues where we'd inadvertently `Show` data types in the registry when we really ought to have been pretty-printing them for display, so we removed it from the codebase. Pretty much everything of significance in the registry is accompanied by a `print` function for display and a JSON encoder/decoder , so I generally use those if I need to 'show' a value for debugging purposes. We also leverage functions like [`unsafeStringify`](https://github.com/purescript/registry-dev/blob/ebf4944079318c2eb27180b3b7dae15da4303ddd/lib/test/Registry/Test/Utils.purs#L46-L48) for testing. Perhaps one of these is a suitable option for you?

PR #695, opened 2024-08-28
Few fixes to SSH and Owner to implement auth and transfer commands in Spago

f-f:
As per title - PR for https://github.com/purescript/spago/issues/1123 coming up soon, depends on these changes getting merged.

I already tested it by moving a package [back](https://github.com/purescript/registry/commit/f25b362ac606ef343249d341abe097b139c5ce82) and [forth](https://github.com/purescript/registry/commit/4c342ff98dda5735a683805413fc673fff05f643)


lib/src/SSH.purs
```diff
@@ -1,21 +1,27 @@
 module Registry.SSH
   ( PublicKey
   , PrivateKey
+  , ParsePrivateKeyError(..)
+  , printParsePrivateKeyError
```
thomashoneyman: We don't generally include 'parse' in the naming for errors — could we rename these?

```suggestion
  , PrivateKeyError(..)
  , printPrivateKeyError
```

lib/src/SSH.purs
```diff
@@ -1,21 +1,27 @@
...
   , Signature(..)
   , parsePublicKey
   , parsePrivateKey
-  , parsePrivateKeyWithPassword
+  , publicToOwner
```
thomashoneyman: Could we rename this as well? We refer to public/private keys with 'key' in the name:

```suggestion
  , publicKeyToOwner
```

lib/src/SSH.purs
```diff
@@ -37,18 +43,24 @@ foreign import parseKeyImpl :: forall r. Fn4 (Exception.Error -> r) (ParsedKey -
...
+  RequiresPassphrase -> "Encrypted private key requires a passphrase"
+  OtherParseError message -> message
 
 -- | Parse a password-protected private SSH key
-parsePrivateKeyWithPassword :: { key :: String, passphrase :: String } -> Either String PrivateKey
-parsePrivateKeyWithPassword { key, passphrase } =
-  case runFn4 parseKeyImpl (Left <<< Exception.message) Right key (notNull passphrase) of
-    Right parsed | not (isPrivateKey parsed) -> Left $ "Expected private key, but this is a public key of type " <> keyType parsed
-    result -> map PrivateKey result
+parsePrivateKey :: { key :: String, passphrase :: Maybe String } -> Either ParsePrivateKeyError PrivateKey
```
thomashoneyman: What's the reason to unify the parsing functions? If they should be unified, then we'll want to update the documentation comment on this function to indicate it _might_ be password-protected (or we could ditch the doc comment because it's not adding much value).

lib/src/SSH.purs
```diff
@@ -88,3 +100,12 @@ isPrivateKey :: ParsedKey -> Boolean
...
 foreign import equalsImpl :: Fn2 ParsedKey ParsedKey Boolean
+
+foreign import publicToOwnerImpl :: Fn1 PublicKey { keytype :: String, public :: String, id :: Nullable String }
+
+publicToOwner :: PublicKey -> Owner
+publicToOwner key =
+  let
+    { id: nullableId, keytype, public } = runFn1 publicToOwnerImpl key
+  in
+    Owner { keytype, public, id: Nullable.toMaybe nullableId }
```
thomashoneyman: Can we write this with do/let? We don't use let/in anywhere else.

```suggestion
publicToOwner :: PublicKey -> Owner
publicToOwner key = do
  let { id: nullableId, keytype, public } = runFn1 publicToOwnerImpl key
  Owner { keytype, public, id: Nullable.toMaybe nullableId }
```

lib/test/Registry/SSH.purs
```diff
@@ -23,12 +24,13 @@ spec = do
...
 
   Spec.it "Parses a password-protected RSA private key" do
-    case SSH.parsePrivateKey id_rsa of
-      Left err1 -> do
-        err1 `Assert.shouldEqual` "Encrypted private OpenSSH key detected, but no passphrase given"
-        case SSH.parsePrivateKeyWithPassword { key: id_rsa, passphrase: id_rsa_password } of
-          Left err2 -> Assert.fail $ "Failed to parse id_rsa private key with password: " <> err2
+    case SSH.parsePrivateKey { key: id_rsa, passphrase: Nothing } of
+      Left err1@SSH.RequiresPassphrase -> do
+        SSH.printParsePrivateKeyError err1 `Assert.shouldEqual` "Encrypted private OpenSSH key detected, but no passphrase given"
```
thomashoneyman: I think this should be `"Encrypted private key requires a passphrase"` based on your print function in the SSH module.

lib/src/SSH.purs
f-f: The error is not about the private key in general, but specifically about parsing it - it even comes from a function called `parsePrivateKey`. So I think it's clearer this way?

lib/src/SSH.purs
f-f: Ah yeah makes sense

lib/src/SSH.purs
f-f: > What's the reason to unify the parsing functions?

I feel it's the other way around - I don't see the reason why they should be separate? It's less/simpler code this way, and whoever is calling this (spago in my case) can just specify a passphrase as the argument if needed.

lib/src/SSH.purs
f-f: Thought you could use `do` only in monadic code. This is fine as well 👍

lib/src/SSH.purs
f-f: I'll remove the doc comment

lib/src/SSH.purs
thomashoneyman: Yeah, I suppose we sometimes include what kind of error (parse error, fetch error, etc.) in the error name. I'm OK with keeping it, but I'd prefer we move the 'parse' name to the error, like this:

```suggestion
  , PrivateKeyParseError(..)
  , printPrivateKeyParseError
```

f-f: Ah yeah that reads better

thomashoneyman (approved):

PR #697, opened 2024-09-07
Fail upon encountering unrecognized fields when parsing config structures

### LINKED ISSUE #1165: Issue warnings when the configuration has unrecognised fields
As noted e.g. in #1159, the fact that we just ignore fields that we don't parse means that it's possible to get the names of the fields wrong and not know about it.

I am not sure how the fix for this one would look like - I think either we'll end up doing some low-level parsing with the `yaml` library in [Config.js](https://github.com/purescript/spago/blob/68b644644faf38f2d2ccbee7b02758de60a8f1c7/src/Spago/Config.js), or hopefully we can do something [in the Codec at the PS level](https://github.com/purescript/spago/blob/68b644644faf38f2d2ccbee7b02758de60a8f1c7/core/src/Config.purs#L66)

fsoikin:
Fixes https://github.com/purescript/spago/issues/1165


lib/src/PackageSet.purs
```diff
@@ -41,9 +41,9 @@ derive newtype instance Eq PackageSet
...
 -- | implementation.
 codec :: CJ.Codec PackageSet
-codec = Profunctor.wrapIso PackageSet $ CJ.named "PackageSet" $ CJ.object
-  $ CJ.recordProp (Proxy :: _ "version") Version.codec
-  $ CJ.recordProp (Proxy :: _ "compiler") Version.codec
-  $ CJ.recordProp (Proxy :: _ "published") Internal.Codec.iso8601Date
-  $ CJ.recordProp (Proxy :: _ "packages") (Internal.Codec.packageMap Version.codec)
-  $ CJ.record
+codec = Profunctor.wrapIso PackageSet $ CJ.named "PackageSet" $ CJS.objectStrict
+  { version: Version.codec
```
f-f: We should use the `recordProp` here as well so that fields stay sorted

spago.yaml
```diff
@@ -1,9 +1,8 @@
 workspace:
   lock: true
-  package_set:
-    registry: 50.10.0
-  extra_packages:
-    codec-json: 1.2.0
+  packageSet:
```
f-f: This repo is built/deployed with Nix - that machinery supports the spago config and lockfile format up to spago 0.93.19, so we should not upgrade the config (nor the lockfile) to use the new format

fsoikin: Ah, that would explain the CI errors

spago.lock
```diff
@@ -2,208 +2,1338 @@ workspace:
...
-        - transformers
-        - tuples
-        - typelevel-prelude
-        - unicode
-        - unsafe-coerce
-        - uuidv4
-        - variant
-      test_dependencies:
-        - spec
+      core:
```
f-f: We should also regenerate the lockfile using spago 0.93.19

thomashoneyman (approved):

f-f:
@thomashoneyman could we revert this one? We are introducing strict parsing of the `Manifest`, `Metadata`, which breaks forward compatibility guarantees (i.e. older versions of this library being able to parse files produced with newer versions which might introduce new fields.

@fsoikin if the intention is to support https://github.com/purescript/spago/issues/1165 then we don't need any of these changes: the Spago config only needs strict record parsing for the `Location` type. The other types we are using [in the Config parsing](https://github.com/purescript/spago/blob/c4c9641fd77e890c3f02a253980713c8982d6567/core/src/Config.purs#L53-L58) are newtypes so are not affected by strict record parsing.

thomashoneyman:
Ah, I was just thinking about backwards compatibility. We can revert.

fsoikin:
The `Location` type does need strict parsing because the `subdir` field is optional and therefore subject to typos. But what about backwards compatibility? The same codec is used for network communication too, right? Maybe have too codecs - `Location.codec` and `Location.codecStrict`?

f-f:
I think we should let the registry do its thing and just have the strict codecs we need in spago itself - the codecs don't need to be colocated with the types

fsoikin:
But then `subdir` is subject to typos, which is what https://github.com/purescript/spago/issues/1165 is complaining about.

fsoikin:
Could be n misspelled `subDir` or `sub_dir`, or even `ref` by mixing up different formats of similar purpose.

f-f:
We can use strict codecs in spago and avoid the typos, but we should just duplicate the codec there instead of changing this one. We don't have the same forwards compatibility concerns in spago itself (we do have some, but just for a subset of the config file)

fsoikin:
Ah, I see the constructors are exported. Somehow I thought they weren't, similar to `Version` and `PackageName`.

PR #700, opened 2025-06-08
A shell script to run local server without wrapping it in Nix

fsoikin:


thomashoneyman:
I don’t love duplicating the Nix wiremock config into a local file too, and I’m hesitant to commit to maintaining it; I’d like it better if we use writeShellScriptBin into the flake to reuse the existing configs instead. I see the usefulness of not having to run in the qemu VM, of course. But if you need to merge it, that’s OK.

fsoikin:
I don't _have_ to merge it, I can just use my local copy, but I thought it might be useful to share the script. 

As for the first part of your comment, I understood almost none of it due to not understanding Nix at all. I am currently sitting in a lecture on how Nix works, and my confidence is actually decreasing over time.

fsoikin:
I'll see if I can extract the wiremock part of the Nix config so it can be run separately from everything else.

thomashoneyman:
I meant wrapping this test script in `writeShellScriptBin`, like:
https://github.com/purescript/registry-dev/blob/e7269e61d885a5ef2376bf5fce5d84b79fe1051e/flake.nix#L264-L276

...and then instead of reimplementing the wiremock configs, you can import them directly into the script. You'd have to lift them out of the integration test to the top level first, but otherwise they can be unchanged.

This is assuming you _want_ to run a local server with wiremock instances...the reason those instances all exist is for the sake of the integration test, which uses a NixOS environment that matches the one we deploy, and has deterministic outputs for things like Pursuit, git changes, and so on. You should be able to execute a command like this:

```sh
curl -X POST http://localhost:8080/api/v1/publish \
    -H "Content-Type: application/json" \
    -d '{
      "name": "effect",
      "ref": "v4.0.0",
      "compiler": "0.15.4",
      "location": {
        "githubOwner": "purescript",
        "githubRepo": "purescript-effect"
      }
    }'
```

What sort of testing are you looking to do? The registry has pretty robust tests that don't need this whole integration test setup. The integration tests are more of a last resort thing to run in CI with the exact operating system we deploy to DigitalOcean.

thomashoneyman:
If you aren't trying to run the integration test, and instead you just want to run the local server such that it can't perform writes, then you can replace the run implementations with no-ops (like what we do in the legacy importer when you provide the dry run flag).
https://github.com/purescript/registry-dev/blob/e7269e61d885a5ef2376bf5fce5d84b79fe1051e/scripts/src/LegacyImporter.purs#L114-L131

I expect this is the easier path to take rather than attempting to reuse the WireMock configs. Those are pretty specific for the integration test.

fsoikin:
Yeah, I realize the error of my ways now. This is way more complicated. I'll retry at a later point, once I gain more understanding.

PR #701, opened 2025-06-09
Update to spago 0.93.44 and Nixpkgs 25.5

f-f:
I forgot why we need to be stuck on Spago 0.93.19, so here's a PR to see what CI thinks of it.

We changed the way test dependencies are included, so I had to separate the test utilities into their own package. Otherwise it should be a pretty boring patch.


thomashoneyman:
If I remember correctly it’s because of the lock file format changing. It may be a bigger patch than you expect to update Spago 😄 I was thinking we could switch to jeslie’s Spago builder going forward.

f-f:
@thomashoneyman I did switch to jeslie0's build machinery and it works (things are good on my machine and CI seems happy)

There is only one gotcha, which is that it does something to the `foreign.js` files and that means that `purs-backend-es` doesn't see them so we have a manual copy step.
The files are in the `output` folder when it gets passed to the optimiser, but it just can't see them, and I couldn't find exactly what the bug is. If you have ideas about what it could be I'm happy to fix it but I'm also happy to leave as is since it works.

thomashoneyman:
@toastal curious if you have any thoughts here. I can try and get a review in over the next day or so, but I’m traveling so if you want to merge I am OK with it as-is.

scripts/default.nix
```diff
@@ -37,7 +36,7 @@ let
       '';
       buildPhase = ''
         ln -s ${package-lock}/js/node_modules .
-        cp -r ${spago-lock.registry-scripts}/output .
+        cp -r ${spago-lock}/output .
```
toastal: Can this be a symlink or no?

app/default.nix
```diff
@@ -32,16 +31,27 @@ let
...
-      mkdir $out
-      mv output-es $out/output
+      mkdir $out;
+      cp -r output-es $out/output;
+      # This for loop exists because purs-backend-es finds the corefn.json files
+      # just fine, but doesn't find the foreign.js files.
+      # I suspect this is because of a quirky interaction between Nix and `copyFile`,
+      # but I'm not sure how to fix it so we work around it by copying the foreign
+      # modules by hand.
+      for dir in output/*/; do
```
toastal, line 49: Why not something like `find output -type d -maxdepth 1 -print0 | xargs -0 -P$NIX_BUILD_CORES -I{} …`? Could be a bit quicker to run this in parallel.

app/default.nix
```diff
@@ -32,16 +31,27 @@ let
       "buildPhase"
       "installPhase"
     ];
-    nativeBuildInputs = [ purs-backend-es ];
+    nativeBuildInputs = [ purs-backend-es-unstable ];
     buildPhase = ''
       ln -s ${package-lock}/js/node_modules .
-      cp -r ${spago-lock.registry-app}/output .
+      cp -r ${spago-lock}/output .
```
toastal: Can this be a symlink or no?

toastal:
Other than the side quest of possibly removing _either_ `flake-utils` _or_ the dependence on Nixpkgs+Flakes[^1], the general Nix code seems just fine.

@thomashoneyman Tho ever since the PureScript community moved to Discord, I haven’t actively been following the project in my recreational time (nor have I had any work in it in recent years) so I don’t really know what has been going on with any of the tooling. If there were some PS work, I would probably be more aware of what is happening & trying to pitch in.

[^1]: the use case for `flake-utils` is providing a Flake for a project but either _not_ using Nixpkgs at all _or_ not using Flakes to managing Nixpkgs version pinning (using `npins`, `niv`, or other alternatives), in which case you would not have `lib` in your Flake—otherwise `lib` already has everything you need to ‘loop’ over architectures without incurring an extra dependency.

thomashoneyman:
The tooling development still happens on the registry-dev and spago repositories so it's in the open (very little is discussed on Discord). No worries about following along — just thought you might have some Nix ideas to contribute, as you had done a nice rework of the package a little while ago! Alas, I no longer work professionally in PureScript myself.

thomashoneyman (approved):

scripts/default.nix
f-f: Sure

app/default.nix
f-f: I can't parse that command - do you have a more precise suggestion? I'm happy with the for loop because it's very readable.

f-f: Ah you did edit in a runnable command, I must have seen an outdated version. That works just as well, but I'd still prefer the for loop

PR #702, opened 2025-06-21
Add a function that produces a `Range` that covers an exact version

garyb:
Part of https://github.com/purescript/spago/issues/1337


thomashoneyman (approved):

thomashoneyman:
We don’t have releases, so Spago will just refer to a newer commit. (At some point we’ll want to release a SemVer version of the library)

PR #706, opened 2025-12-02
Overhaul and modernize Nix flake setup

thomashoneyman:
When setting up a new machine and looking at the Nix setup we have I noticed a lot of quirkiness, nonstandard patterns, and relying on one-off little utilities like `slimlock` (thankfully in #701 we moved to `mkSpagoDerivation` from our homegrown `purix`). We also have some essentially dead code in Nix modules like `run-vm`, as AFAIK no one actually deploys the app to a VM and issues requests against it.

So I took a pass through everything to update to current best practices, improve performance especially in the dev shell, and make the codebase easier for contributors to work with. Here are some high-level changes:

- The `flake.nix` is broken into a few smaller modules (overlays, lib) with the integration test extracted out, so itas more like 250 lines instead of 900 lines
- Switched to the standard `buildNpmPackage` from the custom `slimlock` tool
- Got rid of the separate GitHub `spago test` script, as we can just run it as part of the overall `nix flake check` checks like everything else is
- Cleaned up the wiremock and git mock implementations and added comments to document the integration test so it's easier to understand why we have all this infrastructure for it
- Removed unnecessary legacy support (shell.nix, flake-compat) and dropped the test VM setup
- Included `.envrc` for developers using direnv — I personally need this, don't want to commit it on all fresh clones, and I think @f-f and I both already were on it

The codebase works the same as before, all the same commands (minus running the test VM), just a better setup internally.


f-f (approved):
This is great, thanks!

I would like to eventually get rid of the Nix VM (and replace it with something more agnostic, e.g. docker), but this is a good start 👏 👏

PR #708, opened 2025-12-04
Replace heavyweight integration tests

thomashoneyman:
The old integration test ran everything in a single Nix build sandbox, making iteration slow and debugging painful and it couldn't be run on macos. The new approach uses process-compose to orchestrate services locally:

```
nix run .#test-env     # Start server + WireMock mocks
spago run -p registry-app-e2e  # Run E2E tests (separate terminal)
```

This is meant to make it easy to iterate on the server in local development and implement tests like any other PureScript code. There's more we could do to the Wiremock setup to make it easy to work with, but this should be a good starting point. Here's a demonstration:

https://github.com/user-attachments/assets/da5eec4c-bc30-441b-a9dd-118ea02f8fb9

(This video is on a M2 Mac, demonstrating everyone contributing to the project can do this locally and quickly).

Under the hood, I opted to:

1. Keep the Nix VM integration test, but rename it to be a smoke test and only be about deploying the software the same way Colmena does and making sure the services come up as expected. In the future this is also where we could do things like kill the process and see how it recovers.

2. Add a new test environment based on [process-compose](https://github.com/F1bonacc1/process-compose), which sets up the server and associated Wiremock services, mocked git etc. so we can easily test the server. It's kind of like a Docker container, but much faster and lighter weight, particularly on Mac.

3. Use this and the new `app-e2e` tests to replace the integration test, so we still have an integration test check that happens to be a lot faster.

4. Centralize and clean up our use of env vars in the code base as this was starting to proliferate all over the place.


app/src/App/Server.purs
```diff
@@ -232,64 +232,60 @@ main = do
...
+                    Console.error $ "Healthchecks failed and failure limit reached, will not retry: " <> Fetch.Retry.printRetryRequestError error
+
+                  Succeeded _ -> do
+                    Console.error $ "Healthchecks returned non-200 status and failure limit reached, will not retry."
+
+            loop limit
+          pure unit
+
+      -- Read port from SERVER_PORT env var (optional)
+      port <- liftEffect $ Env.lookupOptional serverPort
```
f-f, line 279: here it says "optional", but I don't see a default port specified anywhere?

thomashoneyman: HTTPurple defaults to 8080. We could provide an explicit default though.

f-f: Ah I see now. Should be enough to note it in a comment!

app/src/App/Server.purs
```diff
@@ -232,64 +232,60 @@ main = do
...
+                  Failed error -> do
+                    Console.error $ "Healthchecks failed and failure limit reached, will not retry: " <> Fetch.Retry.printRetryRequestError error
+
+                  Succeeded _ -> do
+                    Console.error $ "Healthchecks returned non-200 status and failure limit reached, will not retry."
+
+            loop limit
+          pure unit
+
+      -- Read port from SERVER_PORT env var (optional)
```
f-f:
```suggestion
      -- Read port from SERVER_PORT env var (optional, HTTPurple defaults to 8080)
```

f-f (approved):
This is great, thanks @thomashoneyman for looking into this!! 👏

PR #669, opened 2023-11-13
Require all packages to solve / compile and include all valid compilers in their metadata

thomashoneyman:
Fixes #577. Fixes #255. Fixes #696. Fixes #649. Fixes #527. Fixes #684. Fixes #570.

The core problem solved here is identifying what compilers are compatible with a specific package version, such as `aff@7.0.0`. We need this to support an oft-requested feature for Pursuit: filtering search results by a compiler version (or range). It's also useful for other things; for one, it allows us to add the compiler version as a constraint to the solver to produce a build plan that works with the specific compiler given.

Metadata files now include a `compilers` key in published metadata that lists either a bare version (the version used to publish the package) or an array of versions (the full set of compilers known to work with the package). The reason for two representations is that computing the full set of compilers can take a long time; this approach lets us upload the package right away and compute the rest of the valid compilers in a fixup pass. A bare version means the full set has not been computed yet.

All packages must now be solvable. We can't compile a package version if we can't fetch its dependencies, so this becomes a requirement for all packages.

There are only 2 scenarios in which we need to compute the available compilers for a package version:

1. A new package version is published
2. A new compiler version is published

This PR is focused on the first case, and we should do a followup for the second case. (The second case is straightforward, and @colinwahl's compiler versions script already essentially implements it. It's been omitted from this PR for brevity).

A new package version can be published via the legacy importer or via a user submitting an API call, but the result is the same: eventually the `publish` pipeline is run. For that reason I've decided to compute the compiler versions for a package version as part of the publish pipeline where we're already determining resolutions and building with a specific compiler. That centralizes the logic to a single place.

Therefore this PR centers on two things: trying compilers to find all that work for a package version at the end of publishing, and updating the legacy importer to determine a valid compiler version and resolutions before calling `publish`.

I've added some tests and I've run the legacy importer locally; it's about 500 packages in so far and every failure appears to be correct. More comments in the PR.

**Update 2026-01-07:** This now also contains #709, which restricts concurrency on jobs and adds a job executor loop.

PR #709, opened 2025-12-09
Job queue, matrix builder, concurrency control

### LINKED ISSUE #649: Proxy GitHub issue API calls to the registry server
Now that we have a registry server available at https://registry.purescript.org we can have GitHub jobs initiated via GitHub issues skip running a full build in GitHub Actions and instead proxy them to the registry server. The GitHub job can be a simple script (not even written in PureScript) that sends the request to the registry server, polls for logs, and — upon encountering a new log message — comments on the GitHub issue with that message.

This has a few advantages:

1. It allows us to test the registry server more thoroughly, since jobs are actually hitting the server
2. It allows us to tear out some of the GitHub-specific code and treat the GitHub issues as just another "package manager"
3. It allows us, if we wish, to tear out the Nix setup for GitHub Actions and just use basic built-in functionality to send the request and process results. A node script, for example, would get the job done just fine, and we don't have to bother managing things like caches on GitHub.
4. It will be faster since we don't have to build the whole registry-dev repository and load in its Nix dependencies before processing user issues — we can just immediately send off the request.
5. It allows us to disable concurrent jobs, manage the number of package sets releases per day, and otherwise control the incoming jobs we have instead of being forced into the independent actions workflows done by GitHub.

We don't have to move entirely off of Actions; we can, for example, still use it to initiate processes like the legacy importer. It's just that the processing itself will happen on the server.

### LINKED ISSUE #696: Control the concurrent running of registry jobs
The registry supports an API for publishing, unpublishing, and transferring packages, as well as running a compiler matrix for one or all packages and creating new package sets. These operations touch shared resources stored on GitHub and can have relationships among themselves; these factors mean that in the case of concurrent API calls we must carefully control in what order the registry schedules these operations.

Previously we couldn't do anything about concurrent jobs because we rely on GitHub Actions, but now that there's a proper registry server we can control things. Here are some factors to keep in mind:

1. **Jobs cannot be run concurrently because they may rely on shared Git resources.**
Most data about packages is stored in Git; no two jobs can run concurrently if they read from or write to shared Git resources due to the risk of conflicts. For example, publishing a package affects its metadata, manifest entry, and potentially the new-packages.json file, and it reads the metadata and manifest entries for its dependencies at their solved versions. For simplicity, we've started with the rule 'no jobs may run concurrently', but we may choose to allow some limited concurrency (for example, allow package set jobs to run concurrently with publish jobs).

2. **Jobs may have different priorities.**
Some jobs are more urgent than others. Publishing is more important than a package set update. Since jobs cannot be run concurrently, when multiple jobs arrive at once we put them into a queue. This queue must take the job's priority into account, such that if 10 jobs are in the queue and only one is a 'Publish' job, then it should be taken next regardless of the time it arrived.

3. **Jobs may depend on one another.**
Some jobs depend on other jobs. If two jobs arrive to publish packages A@1.0.0 and B@1.0.0, and package B depends on package A at a range that includes 1.0.0, then package A should be published before package B, regardless of the order they originally arrived in.

4. **Jobs may spawn other jobs.**
After a package version is published, we go on to compute the various compiler versions it is compatible with.

These factors mean we must queue jobs instead of starting them as soon as they arrive, and the queue must be more sophisticated than a simple FIFO — we have to consider the priority of the job and its potential dependencies on other jobs also in the queue. If we choose to allow limited concurrency, then we will also have to consider what resources the job uses and only run it so long as no other job uses the same resources.

### Job Types

There are three broad types of jobs that the registry runs:

1. Package jobs
A user may publish, unpublish, or transfer a package.

2. Matrix jobs
The registry tests all compiler versions against every package version that is published. These jobs are spawned in one of two ways. First, when a package version is published, the registry creates an independent compiler test job for every compiler version the registry supports. Second, when a new compiler version is released, registry maintainers can submit a matrix request to the API, which creates an independent test job for every package version in the registry at that compiler version.
 
3. Package set jobs
A user may submit a package set update that adds, upgrades, downgrades, or removes packages from the most recent package set (downgrade/remove can only be performed by registry maintainers).

Matrix jobs targeting many package versions are special because they cannot simply be added to the job queue in any order. Instead, matrix jobs must be topologically sorted and then inserted into the queue so that they can be popped from the queue in the correct order.

### System Overview
Currently, the registry works like this: when a new job is submitted to the API, we fork a new process that adds the job to the jobs table, runs the job, and then marks it as completed. However, this is unsuitable given the restrictions described above.

Instead, I propose a different model. The registry API exposes endpoints where you can submit jobs, but instead of executing them, we immediately store each job in the database. The registry server, at startup, forks a long-running job executor process that is responsible for selecting and executing jobs. At each tick (perhaps every second), the executor follows these steps: 

1. Find the highest-priority job available. If none are available, sleep, then return to (1). 
2. Execute the job. 
3. Mark the job completed and return to (1).

We broadly group jobs into three priority buckets; within any given bucket, jobs are picked up in first-come-first-served order.

1. Package Jobs (Publish/Unpublish/Transfer)
2. Matrix Jobs
3. Package Set Jobs

We include two mechanisms to prevent the job executor loop from either dropping jobs that were interrupted (such as by a server restart) or freezing if a job takes too long.

1. At startup, the executor will scan the database for jobs that have been started but not completed. If there is only one job, it will execute the job. If there are multiple jobs (somehow), then it will choose one to execute and will reset the ‘started’ time for the others.
2. When a job is started, the job executor holds on to the fiber that’s running the job and starts a timer. If the job exceeds the timeout, then the job executor will kill and retry the job. After N retries, the job is marked as failed.

We also give the job executor an extra responsibility, as it owns the queue: when the core repositories have too many commits (more than 100k), it should squash-commit the repositories before picking up the next job so as to make sure consumers of the repositories don’t have to fetch too much data.

### Matrix Jobs

A few extra notes about matrix jobs — these are a pairing of a compiler version and a package version where we compile the package and see what compilers it works with. This happens either post-publishing (we test all compilers against that package version) or on a new compiler release (we test the new compiler against all package versions).

These jobs have a couple extra quirks to them:

1. Matrix jobs are run individually but are conceptually a bulk action  — either "all compilers against package version X" or "compiler version X against all package versions". The reason we don't actually run them in bulk is because...
2. Matrix jobs are low priority, so package jobs like publishing, unpublishing, and transferring should take priority over matrix jobs. 
3. Matrix jobs can depend on each other because they look at the `compilers` metadata field for dependencies. Thus, for example, when testing a new compiler against all packages we need to topologically sort the jobs.

The issue with splitting a conceptual bulk action into individual jobs targeting a package version / compiler version is that we should be sure that the registry doesn't change in a way that would cause two matrix jobs against package a@x.y.z to have different results for any other reason than compilation failure. For example, if we run a few jobs against a@x.y.z, and then a dependency has a new version published, then the remaining jobs against a@x.y.z may solve to different build plans and could start failing. Or, if we choose to solve a matrix job at creation time and store the result in the queue, and then a dependency is unpublished, then the existing plan can fail.

I'm not entirely sure what the best approach is, here. I think we can get away with:

1. On a new compiler release, solve and topologically sort all the package versions, and then insert them into the matrix queue in order. We can then run the queue based on creation time alone.
2. On a new package publish, solve the matrix jobs right then and store it in the database with the job so if the registry state changes in the meantime the build plans are not affected.
3. On an unpublish job, verify whether this unpublish will affect any queued matrix jobs. If so, then update the solutions for those matrix jobs.

But I'm open to ideas to make this more robust.

### Database Design

Here is the current jobs table:

```sql
CREATE TABLE jobs (
  jobId text primary key not null,
  jobType text not null,
  packageName text not null,
  ref text not null,
  createdAt text not null,
  finishedAt text,
  success integer not null default 0
);
```

This design is unsuitable to capture the various job types and priority ordering we need. Instead, we should separate jobs into multiple tables by their coarse type (package job, matrix job, package set job). Priority selection is simply looking first at the package jobs table, then matrix jobs table, and so on.

```sql
-- Common job information table
CREATE TABLE job_info (
  jobId TEXT PRIMARY KEY NOT NULL,
  createdAt TEXT NOT NULL
  startedAt TEXT,
  finishedAt TEXT,
  success INTEGER NOT NULL DEFAULT 0
);

-- Package-oriented jobs (publish/unpublish/transfer)
CREATE TABLE package_jobs (
  jobId TEXT PRIMARY KEY NOT NULL,
  jobType TEXT NOT NULL,
  packageName TEXT NOT NULL,
  packageVersion TEXT NOT NULL,
  payload JSON NOT NULL,
  FOREIGN KEY (jobId) REFERENCES job_info (jobId)
);

-- Compiler matrix jobs (one compiler, all packages)
CREATE TABLE matrix_jobs (
  jobId TEXT PRIMARY KEY NOT NULL,
  packageName TEXT NOT NULL,
  packageVersion TEXT NOT NULL,
  compilerVersion TEXT NOT NULL,
  -- the build plan, which should be computed before the job is stored in the queue
  -- so that if multiple jobs targeting one package get interrupted by a higher-
  -- priority job then the build plan is not affected.
  payload JSON NOT NULL,
  FOREIGN KEY (jobId) REFERENCES job_info (jobId)
);

-- Package set jobs
CREATE TABLE package_set_jobs (
  jobId TEXT PRIMARY KEY NOT NULL,
  payload JSON NOT NULL,
  FOREIGN KEY (jobId) REFERENCES job_info (jobId)
); 
```

### Open Questions

* Did we ever figure out what to do if a package version is unpublished and other packages depend on that version specifically? Do we allow that? Do we cascade the breakage?

* GitHub Actions are a loophole that allows for jobs to be run concurrently regardless of what the server is doing. Any solution to this issue should also include #649.

f-f:
WIP, should fix #696, fix #649


f-f:
Main chunks of work to be done at this point:
- enqueuing/running matrix jobs, need to settle on a design for this. Current direction is "build plan is fixed at submission time, and we dynamically add jobs for dependent packages after a job is completed"
- enqueing/running package set jobs, need to figure out auth story for that
- return jobs from the API; need to spec it more:
  - do we keep backwards compatibility? Spago should keep working
  - what details do we return?
  - when returning a list we should allowing selecting queued/done/in progress jobs

fsoikin:
But the first two bullets can be implemented as separate changes, right? They're not needed for strict parity.

f-f:
@fsoikin the overall goal is to merge #669 ASAP, so that we can reupload the whole registry.
That can't be merged until we have guaranteed single job execution (#696, this PR, and #649, not yet in this PR but soon), because jobs would otherwise take too long and likely conflict. Splitting off compilation for different compilers is part of making the jobs faster and support keeping the compiler information up to date.

We could be doing all these changes in separate PRs but I don't see the point since they will all need to end up in the `compilers-in-metadata` branch since we can't merge to trunk until we have the whole package.

app/src/App/Server/Router.purs
```diff
@@ -0,0 +1,110 @@
...
+
+  _, _ ->
+    HTTPurple.notFound
+  where
+  insertPackageJob :: PackageOperation -> ContT Response (Run _) Response
+  insertPackageJob operation = do
+    lift $ Log.info $ "Enqueuing job for package " <> PackageName.print (Operation.packageName operation)
+    jobId <- newJobId
+    lift $ Db.insertPackageJob { jobId, payload: operation }
+    jsonOk V1.jobCreatedResponseCodec { jobId }
```
thomashoneyman: The old code checked for running jobs before creating new ones:

```purescript
lift (Db.runningJobForPackage packageName) >>= case _ of
  Right { jobId, jobType: runningJobType } -> ...
```

Looks like the duplicate checking is no longer there, is that intentional?

app/src/App/SQLite.js
```diff
@@ -11,49 +17,155 @@ export const connectImpl = (path, logger) => {
...
-    "UPDATE jobs SET success = @success, finishedAt = @finishedAt WHERE jobId = @jobId"
-  ).run(result);
+export const selectNextPackageJobImpl = (db) => {
+  const stmt = db.prepare(`
+    SELECT job.*, info.createdAt, info.startedAt
+    FROM ${PACKAGE_JOBS_TABLE} job
+    JOIN ${JOB_INFO_TABLE} info ON job.jobId = info.jobId
+    WHERE info.finishedAt IS NULL
+    AND info.startedAt IS NULL
+    ORDER BY info.createdAt DESC
```
thomashoneyman: Why `DESC` here? Don't we want the oldest job to go out first (FIFO), like it previously was? Same for the other `selectNext*` functions.

thomashoneyman:
Also the tests are failing because of a missing `version` field in the fixtures, see:
https://github.com/purescript/registry-dev/commit/5afab581a2c436a9167b6a5f5048a89ebebc7747

app/src/App/Effect/Db.purs
```diff
@@ -70,28 +102,52 @@ type SQLiteEnv = { db :: SQLite }
...
+
+  SelectLogsByJob jobId logLevel since reply -> do
+    { fail, success } <- Run.liftEffect $ SQLite.selectLogsByJob env.db jobId logLevel since
+    unless (Array.null fail) do
+      Log.warn $ "Some logs are not readable: " <> String.joinWith "\n" fail
+    pure $ reply success
+
+  DeleteIncompleteJobs next -> do
+    Run.liftEffect $ SQLite.deleteIncompleteJobs env.db
+    pure next
```
thomashoneyman: Are we sure we want to just delete incomplete jobs? Should they instead be re-inserted into the queue? For example, reset `startedAt` to NULL so the job is picked up again.

We may also want to put more effort into making partially-completed jobs more recoverable, such as making operations as close to idempotent as possible and having a way to sweep through and catch any only partially-complete operations and decide whether to roll them back or to retry and complete them.

thomashoneyman:
> Main chunks of work to be done at this point:

In addition to your list (some responses to that below), the other point that I'm not seeing here is that we need to ensure that the github issue module only proxies requests over to the server API and writes comments on 'notify' logs but does not actually execute e.g. the publish operation itself. Otherwise we've defeated the purpose of the pull request, as we can't enforce a lock on the git commits anymore.

> * enqueuing/running matrix jobs, need to settle on a design for this. Current direction is "build plan is fixed at submission time, and we dynamically add jobs for dependent packages after a job is completed"

This makes sense to me: start with no-dependency packages, and on completion we queue dependents that are satisfied by this package existing, propagate outwards.

> * enqueing/running package set jobs, need to figure out auth story for that

What do we need to do beyond what we have today? Today, we open an issue, `pacchettibotti` signs the payload via the GitHub Actions setup, and then the job is executed by the action. With the server, the only difference would be that this payload is submitted to the server API. I'm not sure I'm seeing what the additional auth issues are. We can also have a daily GitHub Action cron job that kicks off the daily package set updater, so as far as the server is concerned it's just receiving an authenticated package set update.

The same approach is used for transfers or unpublishing: open the issue as a `purescript/packaging` maintainer and it will be auto-signed by pacchettibotti. I like this because it's public.

app/src/App/Server/Router.purs
f-f: I have noticed as well, and I am reshuffling this part of the code to add this back, among other things.

I have been thinking about splitting the "package job" table as well into different tables for publish/unpublish/transfer, because they do require different checking (publishing/unpublishing needs to check for the tuple package name and version, transferring only needs to match the package name). Thoughts?

app/src/App/SQLite.js
f-f: Yeah I have noticed too, it should be `ASC`.

And it should probably not be `LIMIT 1` either, as we likely want to do some toposorting on the result. I have not figured out yet where we draw the line though, I wouldn't like to be toposorting thousands of matrix jobs over and over, so we might want keep the limit and toposort on insertion. But then we need to `ORDER BY` something else than creation date when fishing that out. Ideas?

app/src/App/Effect/Db.purs
f-f: Yes to both, I have the first item in my todo list (that is, resetting `startedAt` instead of deleting), and for the second item we should be reasonably good at it already, the publishing pipeline is fairly resilient at this point. We shall have another pass at it though.

f-f:
@thomashoneyman 

> Also the tests are failing because of a missing version field in the fixtures, see:
https://github.com/purescript/registry-dev/commit/5afab581a2c436a9167b6a5f5048a89ebebc7747

Yeah, I have not fixed that yet because I was not sure we need the version in there at all. Help me with reminding why we need the version now 😄 

Also: I see these are on a different branch? I will cherry-pick them here, and feel free to push new stuff here as well.

> the other point that I'm not seeing here is that we need to ensure that the github issue module only proxies requests over to the server API and writes comments on 'notify' logs but does not actually execute e.g. the publish operation itself

Yeah it's planned but I want to thread through the matrix jobs and the package set jobs to the database before going to the github interaction part. It might be a nice spot where we can parallelise work if you want to have a go at it. I will push the current state of my branch so you have the latest.

> The same approach is used for transfers or unpublishing: open the issue as a `purescript/packaging` maintainer and it will be auto-signed by pacchettibotti. I like this because it's public.

Yeah if you think it's straightforward then it probably is 😄 I have not looked at this chunk of code yet so I might be overthinking stuff.

I like the public aspect of the package set updates as well, and I in fact want to make it _public-only_, in the sense that I'd put the package-set server endpoints behind encryption/authentication (with just a simple shared secret for example) so that only the code that runs on github can hit them and we don't have to worry about things coming to the registry server from other places. No strong feelings about this though if you think it's too convoluted

thomashoneyman:
Ah, if you want it to be forced public then we could indeed do a shared secret we send in the HTTP header or something. I’d like transfer and unpublish to be public too — we could potentially remove the ability for pacchettibotti signatures to be valid for those and only the shared secret can be used as a trustee override?

f-f:
> Ah, if you want it to be forced public then we could indeed do a shared secret we send in the HTTP header or something. I’d like transfer and unpublish to be public too — we could potentially remove the ability for pacchettibotti signatures to be valid for those and only the shared secret can be used as a trustee override?

@thomashoneyman we can't make unpublish and transfer public-only: package maintainers should be able to do it at the CLI (with Spago or whatever else), with their keys and so on.

I am interested in making public-only the package-sets endpoint, but maybe that's unnecessarily extending the scope here and we should stick to the bare minimum, which is proxy the GitHub calls to the server. Let's do just that here and take the rest in a different PR.

app/src/App/SQLite.js
f-f: This is now `ASC`

thomashoneyman:
[9a8d1ba](https://github.com/purescript/registry-dev/pull/709/commits/9a8d1ba152b2116bc6c0ec8f95ae674720e1dc47) implements the "thin client" approach described in #649. The core idea is that `GitHubIssue.purs` no longer runs the full registry machinery locally — instead it acts as a lightweight proxy:

- Parse the issue body to determine operation type (publish, package-set update, or authenticated operations like unpublish/transfer)
- Re-sign authenticated operations with pacchettibotti credentials if submitted by a trustee
- POST to the registry API server at the appropriate endpoint
- Poll for job completion at `/v1/jobs/{id}`, using the since and level query parameters to fetch only new logs at INFO level or above
- Post logs as GitHub comments as they arrive (we can tweak this if we're posting too rapidly)
- Close the issue on success or leave it open with an error comment on failure

This tears out most of the heavy dependencies from the GitHub workflow — no more `Storage`, `Registry`, `PackageSets`, `Pursuit`, or `Source` effect interpreters. The GitHub job now only needs enough machinery to authenticate with GitHub, parse the operation, and make HTTP calls. The actual package processing happens on the server. The polling uses a 5-second interval with a 30-minute timeout. Logs are filtered to INFO level since DEBUG output would be too noisy for GitHub comments.

I added a couple of integration tests for GitHubIssue specifically which write the event to disk, then triggers the pipeline. The point of these is to test the operation completes, sure, but more specifically that we see the issue comments, issue close events, and trustee signing we expect. The file is kind of verbose with the helpers, but those helpers will come in handy later I hope for other tests!

Along the way I fixed a few other things I noticed:

- `app/src/App/SQLite.purs`: `selectJob` used `<|>` with `ExceptT`, but `Right Nothing` terminates the chain early. So jobs that _aren't_ publish jobs wouldn't get selected. Changed to use `firstJust` helper that continues searching when a job type isn't found. Also in the FFI fixed a missing `params` variable which was causing the test failures.
- `app/src/App/Server/JobExecutor.purs`: The job executor called `runEffects env` but `env.jobId` was `Nothing`, so logs weren't getting written to the DB. Fixed it by setting `envWithJobId = env { jobId = Just jobId }`.

thomashoneyman:
I put 6c023cfd86e383b4b1de7e3825d552821142aa92 in a separate commit because it's a bigger structural change — we no longer have any reason for a `COMMENT` effect since logs all go into the db and there's no direct connection to GitHub from the registry anymore.

But we do still need a way to indicate a log is more important than `INFO` but not quite a warning or error so it can be pushed as a comment to GitHub. Maybe Spago can do something with it too. To that end I added `NOTIFY` as a log level and replaced all usages of `comment` with it.

thomashoneyman:
I'm also interested in adding a few more package fixtures so that we can e2e test the job queue. For instance, send 5 jobs at the same time, or send jobs of varying types and make sure the result order is what we expect (a job was picked up in priority order, matrix jobs finished last, etc.), that the compiler matrix completes, and so on.

app/src/App/Server/Router.purs
fsoikin: But wait, why do we need to check for a running job? Just let them run I say.

thomashoneyman: I suppose so long as everything fails fast then there wouldn’t be much waste in just queuing up duplicsyes

f-f:
> I'm also interested in adding a few more package fixtures so that we can e2e test the job queue. For instance, send 5 jobs at the same time, or send jobs of varying types and make sure the result order is what we expect (a job was picked up in priority order, matrix jobs finished last, etc.), that the compiler matrix completes, and so on.

Yes. I have a WIP commit from last week for this, will wrap it up in the next day or so and push it. It implements the `jobs` endpoint, so we can query it and check that the jobs are run in the right order, and so on.

app/src/App/Server/Router.purs
f-f: Failing fast is very valuable because every job that runs needlessly is taking time away from the other ones, and we potentially have very long queues (e.g. when a new compiler appears). This check is very cheap so I think it's a no brainer to have it. I added it in c33a3ad81bb9c6823a9cd1e4ea67338a05c23fc9

f-f:
@thomashoneyman I added a few tests in e524f004c731e7df60f247bf8b3f607ac07e6a7e and 6dc01f0175bb0cb87e75e7633405742dd69f3ad6 - they pass but the new fixture makes something else fail, I'll have a look at that.

I wanted to cover the two scenarios for the matrix jobs, (1) new compiler being detected, and triggering a whole-registry recompile, and (2) jobs for each compiler after a package publish. Let me know if you have more scenarios in mind that we could test for.

nix/test/config.nix
```diff
@@ -243,6 +254,10 @@ let
   ];
 
   s3Files = [
+    {
+      name = "effect-4.0.0.tar.gz";
+      path = rootPath + "/app/fixtures/registry-storage/effect-4.0.0.tar.gz";
+    }
```
thomashoneyman: I hadn't included the effect tarball in fixtures because our primary test publishes effect@4.0.0 — so we'd expect that effect isn't _already_ in S3. Maybe this doesn't cause any issues? But that's why it hasn't been in the fixtures already.

app/src/App/SQLite.purs
```diff
@@ -196,19 +195,20 @@ type SelectJobRequest =
...
-  case fail of
-    [] -> runExceptT $ firstJust
-      [ selectPublishJobById logs
-      , selectMatrixJobById logs
-      , selectTransferJobById logs
-      , selectPackageSetJobById logs
-      , selectUnpublishJobById logs
-      ]
-    _ -> pure $ Left $ "Some logs are not readable: " <> String.joinWith "\n" fail
+  { fail: unreadableLogs, success: logs } <- selectLogsByJob db (JobId jobId) logLevel since
```
thomashoneyman, line 202: Are we actually getting "unreadable logs"? I'm surprised we'd be seeing any kind of decode error. Is there something we need to fix?

nix/test/config.nix
f-f: The publishing E2E happens just fine with this, as the S3 list response does not include the package.

We _need_ to put there some tarball that we publish, so we can test the enqueueing of the new matrix jobs for that package version (which pull the tarball we just published)

Ideally we'd have a wiremock with memory, so it would remember which data we pushed to it - but I suspect that would require writing a small mock server ourselves...

app/src/App/SQLite.purs
f-f: `logLineFromJSRep` returns a `Left` if it can't decode the timestamp or the log level. We currently decode everything fine, this is future proofing - worth reporting if we accidentally break something.

nix/test/config.nix
thomashoneyman: It is possible to set up wiremock with memory, I just didn't do it yet!

thomashoneyman: https://github.com/purescript/registry-dev/pull/713

PR #710, opened 2025-12-13
Upgrade better-sqlite to move on from nodejs_20

thomashoneyman:
We're currently stuck pinned to nodejs_20 because of an old better-sqlite3 version. This commit upgrades the version so we can be on recent node versions instead.


thomashoneyman:
#711 builds on this with the rest of the npm deps, if it looks good can merge that into this, then this into master.

f-f (approved):

PR #711, opened 2025-12-13
Upgrade npm deps to latest

thomashoneyman:
This changeset upgrades the rest of the npm deps besides better-sqlite. Tests pass and I've verified the legacy importer runs (it exercises the `octokit` bindings). This resolves the various warnings we have going on around old dependencies.


f-f (approved):
Awesome, thanks!

PR #712, opened 2025-12-16
Update to spago 0.93.45

thomashoneyman:
This updates to the newer spago, which doesn't require us to bring in the old nodejs_20 version from nixpkgs anymore, and just pins us to the unstable version since I _think_ we don't have to care about lockfile format changes for the time being. If that changes, then our build will fail and we'll need to pin to a spago version again.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -46,11 +46,11 @@
         "registry-index": "registry-index"
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


f-f (approved):
Fantastic, thanks ❤️

PR #713, opened 2026-01-07
Refactor e2e tests with wiremock scenarios

thomashoneyman:
This adds a wiremock state machine in config.nix so that we can transition package states in mock environments. Specifically, we can do something like publish `effect-4.0.0`, at which point its tarball will be available in storage, the `available-versions` returned by Pursuit include the version, and so on.

The reason to do this is so we can try workflows like: publish effect-4.0.0, then publish console-6.0.0 (which depends on effect), and then try to unpublish effect-4.0.0 (should not be possible, because console depends on it).

I've implemented this, and I've also added a number of new e2e tests. Specifically:

* each endpoint gets its own test file focused on testing the primary use of that endpoint, success or error
* there is a GitHubIssue test for testing workflows triggered by github issues
* there is a Workflows test which tries edge cases like the publish-publish-unpublish above, or which checks on job priority like publishing effect, then prelude, and making sure prelude finishes before all the matrix jobs do.

Basic idea is to merge this into the `concurrent-jobs` branch to be our test suite as we continue iterating on things.


thomashoneyman:
The server logs have some messages about not being able to write to `scratch/.cache` which I'll look into tomorrow, but otherwise this seems like a pretty good set of tests for core registry functionality we can use as we iterate — sorry it turned out a bit bigger than expected 😬 

The one bummer here is that we now reset state between test runs, which means modifying the database, checked out repos, and so on. The upside is we have a clean state on each run and can truly inspect the results of each test. The downside is that you can no longer just start `test-env` in a terminal and do `spago run` in the other. Instead, you need to source the `test-env.sh` produced by test-env first, _then_ run `spago run`.

Small additional hassle, but I'm not sure of a better way to get around it...if you're on linux you can just do `nix build .#checks.x86_64-linux.integration`, but that's not available on macOS.

```console
E2E Tests » Endpoints » Jobs » Status endpoint
  ✓︎ can reach the status endpoint
E2E Tests » Endpoints » Jobs » Jobs list
  ✓︎ lists jobs and respects include_completed filter
E2E Tests » Endpoints » Jobs » Jobs API error handling
  ✓︎ returns HTTP 404 for non-existent job ID
E2E Tests » Endpoints » Publish » Publish workflow
  ✓︎ can publish effect@4.0.0 and filter logs
E2E Tests » Endpoints » Publish » Publish state machine
  ✓︎ returns same jobId for duplicate publish requests
E2E Tests » Endpoints » Unpublish » Publish-Unpublish workflow
  ✓︎ can publish effect@4.0.0 then unpublish it
E2E Tests » Endpoints » Transfer » Transfer workflow
  ✓︎ can transfer effect to a new location
E2E Tests » Workflows » GitHubIssue » GitHubIssue end-to-end
  ✓︎ handles publish via GitHub issue, posts comments, and closes issue on success
  ✓︎ posts failure comment and leaves issue open when job fails
  ✓︎ calls Teams API to verify trustee membership for authenticated operation
  ✓︎ posts error comment when issue body contains invalid JSON
E2E Tests » Workflows » Multi-operation » Concurrent git operations
  ✓︎ git repos remain clean after all matrix jobs complete
E2E Tests » Workflows » Multi-operation » Dependency and unpublish interactions
  ✓︎ publishing a package fails when its dependency was unpublished
  ✓︎ unpublishing a package fails when dependents exist in manifest index
E2E Tests » Workflows » Multi-operation » Job priority
  ✓︎ second publish job completes before first package's matrix jobs finish
```

app-e2e/src/Test/E2E/Endpoint/Jobs.purs
```diff
@@ -0,0 +1,69 @@
...
+      let
+        recentCount = Array.length recentJobs
+        allCount = Array.length allJobs
+
+      Assert.shouldSatisfy allCount (_ > 0)
+      Assert.shouldSatisfy recentCount (_ <= allCount)
+
+      -- Verify seeded matrix jobs exist (prelude and type-equality are seeded by test env)
+      let
+        seededPackages = [ Fixtures.prelude.name, Fixtures.typeEquality.name ]
```
thomashoneyman: This is potentially an artifact of WIP code? But the server always starts off with matrix jobs to build for existing published packages.

app/src/App/CLI/Git.purs
```diff
@@ -214,8 +214,8 @@ gitCommit { address: { owner, repo }, committer, commit, message } cwd = Except.
   -- Git will error if we try to commit without any changes actually staged,
   -- so the below command lists file paths (--name-only) that have changed
   -- between the index and current HEAD (--cached), only including files that
-  -- have been added or modified (--diff-filter=AM).
-  staged <- exec [ "diff", "--name-only", "--cached", "--diff-filter=AM" ] \error ->
+  -- have been added, modified, or deleted (--diff-filter=AMD).
+  staged <- exec [ "diff", "--name-only", "--cached", "--diff-filter=AMD" ] \error ->
```
thomashoneyman, line 218: one bug found: when unpublishing a package with only one version (i.e. `effect-4.0.0` in tests), this filter was leaving the deleted registry index file in the git worktree and the commit never went through. now, we pick up deletes as well. edge case, of course.

app/src/App/API.purs
```diff
@@ -296,9 +296,11 @@ authenticated auth = case auth.payload of
             { published = Map.delete payload.version prev.published
             , unpublished = Map.insert payload.version unpublished prev.unpublished
             }
+        -- Delete manifest first to check for dependents before any side effects.
+        -- This ensures we don't delete from storage if other packages depend on this one.
+        Registry.deleteManifest payload.name payload.version
```
thomashoneyman, line 303: if we do this in the existing order, then in this scenario:

1. package A is published
2. package B, a dependent, is published
3. package A is unpublished

we already design such that the package can't be unpublished once other packages depend on it (you'd have to unpublish _those_ first). but in the existing ordering, the tarball would already be deleted before the operation failed. now, the tarball is only deleted if you really are allowed to delete the manifest.

app-e2e/src/Test/E2E/Endpoint/Publish.purs
```diff
@@ -0,0 +1,54 @@
...
+        Just publishedMeta -> do
+          Assert.shouldSatisfy (Sha256.print publishedMeta.hash) (not <<< String.null)
+
+      manifestEntries <- Env.readManifestIndexEntry Fixtures.effect.name
+      let hasVersion = Array.any (\(Manifest m) -> m.version == Fixtures.effect.version) manifestEntries
+      unless hasVersion do
+        Assert.fail $ "Expected version " <> Version.print Fixtures.effect.version <> " in manifest index"
+
+      -- TODO: This should verify the compilers made it into the metadata? How do we know what compilers should
+      -- be added though?
```
f-f: We can list all compilers available and check the latest one is there at least, that should be good enough

app-e2e/src/Test/E2E/Endpoint/Jobs.purs
```diff
@@ -0,0 +1,63 @@
...
+import Test.Spec as Spec
+
+spec :: E2ESpec
+spec = do
+  Spec.describe "Status endpoint" do
+    Spec.it "can reach the status endpoint" do
+      Client.getStatus
+
+  Spec.describe "Jobs list" do
+    Spec.it "lists jobs and respects include_completed filter" do
```
f-f: this does more though. it publishes a package and checks the matrix jobs have run as well, I think it's worth noting.

app/src/App/CLI/Git.purs
f-f: wow, nice catch!!

app/src/App/Server/Router.purs
```diff
@@ -112,10 +112,10 @@ router { route, method, body } = HTTPurple.usingCont case route, method of
         HTTPurple.internalServerError $ "Error while fetching jobs: " <> err
       Right jobs -> jsonOk (CJ.array V1.jobCodec) jobs
 
-  Job jobId { level: maybeLogLevel, since }, Get -> do
+  Job jobId { level: maybeLogLevel, since, limit }, Get -> do
     now <- liftEffect nowUTC
     let oneHourAgo = fromMaybe now $ DateTime.adjust (negateDuration (Hours 1.0)) now
-    lift (Run.Except.runExcept $ Db.selectJob { jobId, level: maybeLogLevel, since: fromMaybe oneHourAgo since }) >>= case _ of
+    lift (Run.Except.runExcept $ Db.selectJob { jobId, level: maybeLogLevel, since: fromMaybe oneHourAgo since, limit: fromMaybe 100 limit }) >>= case _ of
```
f-f: What's the reason for parametrising the limit here? It's been fixed all along so that consumers are forced to do pagination and you can't overload the server (by requesting e.g. 1000000 entries)

app/src/App/API.purs
f-f: ah, makes sense. Let's add this clarification in the new comment

f-f:
> Instead, you need to source the test-env.sh produced by test-env first, then run spago run.

that's fine and possibly better if we can bundle that in a single invocation.

how does one terminate wiremock? is it now started by the tests, or how does that work?

thomashoneyman:
> how does one terminate wiremock? is it now started by the tests, or how does that work?

It's the same as before — `.#test-env` starts process-compose with the wiremock servers, and then you can Ctrl+C or process-compose down to shut them down.

app-e2e/src/Test/E2E/Endpoint/Publish.purs
thomashoneyman: i opted to look up the matrix jobs which were spawned, then verify those and the published compiler all made it in. we can't track the "latest" because that changes over time, but fixtures set prelude to only be compatible with a small range of compilers to avoid having to run a ton of useless compilation.

app/src/App/Server/Router.purs
thomashoneyman: fixed! sorry, misunderstood why this was missing.

thomashoneyman:
The new way to run the env is:

```console
# to start the wiremock servers, git mocks, etc.
nix run .#test-env -- --tui

# to execute the tests
spago-test-e2e
```

They're intentionally separate because this lets you iterate on tests and keep running `spago-test-e2e` without having to redo the nix build for the env. Alternately, on Linux, if you just want to do it in one command you can run the Nix `integration` check.

PR #714, opened 2026-01-07
Move `ref` from metadata to manifest

### LINKED ISSUE #527: Proposal: Store full location in published metadata versions
Looking at the package location and ref in metadata will not be sufficient to know where a package was fetched from if a `Transfer` operation has been run.

If the transfer was from Git to Git then the refs may be preserved, and this will surely be the most common case. But there are other transfers that can happen, such as transferring a package within a repository by changing its `subdir` key. In that case, old refs will point to a repo in which the given `subdir` does not exist.

I think we should specify, in package metadata, the full location + ref that was used to package each version, instead of just a top-level location. The top-level `location` field in metadata would remain, of course, because it's possible to process a transfer without actually publishing a new version.

Alternately, if we don't really care if the published version `location` is fully accurate, then perhaps we don't even store this information in the first place?

thomashoneyman:
Fixes #527 by moving the `ref` field. The ref is used for two things:

1. Historical record of where a package came from, e.g. a `Location` + reference as provided in the publish payload
2. Reconstruction of URLs for legacy package sets

Way back in the day we put it in metadata but that didn't really make sense; in the case of a transfer, the location and refs may be out of sync. Now's a good time to finally make the change because we're on the verge of a reupload that will require updating all the manifests anyway.


f-f (approved):
Looks great!!

Issue #525, opened 2022-09-29
Proposal: Reimplement 'purs publish' in the registry


thomashoneyman:
I think we should stop using `purs publish` and instead move that functionality into the registry. `purs publish` has many restrictions that prevent the registry from working properly, such as having to have a SemVer-compliant Git tag checked out with a clean source tree (except for the `purs.json` file). To support monorepos or non-GitHub locations we have to lift these restrictions. Rather than iterate on the compiler I think we should just do this in the registry, because then we can report better errors and not be tied in with compiler releases.

`purs publish` can stay in the compiler for non-registry users – the Pursuit API isn't changing – but I think eventually we will want to deprecate it altogether and have the registry be the direct pipeline to Pursuit.

We can do this because we already have, in the API pipeline:

1. The dependency resolutions for the package
2. The `purs.json` or `spago.yaml` manifest for the package
3. The `output` directory produced by the compiler; we can also generate docs if need be.

f-f:
From today's discussion in the Registry Weekly: the side of the compiler that deals with publishing is adjacent to the docs, and that's a big chunk of code that might not be straightforward to port. One of the possibilities that we discussed was to sidestep `purs publish` entirely, and change Pursuit instead.

thomashoneyman:
To re-implement `purs publish` in the registry all we have to do is construct the `Package` type used by the compiler and Pursuit. I've explored this in #328 and it's _mostly_ straightforward. I've added the type and JSON encoding here:

https://github.com/purescript/registry-dev/blob/e5120cd613b6c900f19021e9089aeb31587429fe/app/src/App/Effect/Pursuit.purs#L147-L168

I omitted fields that aren't actually used by Pursuit, include `modulesMap`. If that turns out to be used we can easily add it, since we already do this via `associateModules` in the publish pipeline.

The only type I didn't implement is that of the `modules` key, which contains the docs.json for each module the package includes or depends on. Ideally we wouldn't have to actually type the entire docs.json format because it's a behemoth — we would type it as `Json` and just read the file in from disk (the output directory for the package we're publishing). Alas, the docs.json file always sets `reExports` to an empty array, and those re-exports are only associated during `purs publish` via this function:

https://github.com/purescript/purescript/blob/6b49918b9646863e73bbedd1d47f474ba3783408/src/Language/PureScript/Docs/Collect.hs#L43

[...]

PR #715, opened 2026-01-11
Rework daily cron jobs to hit the API instead of working directly on the Registry

f-f:
~~This PR integrates into the Job Queue framework the three "Registry scripts" that we run daily to keep data tidy: the package transferrer, the package set upgrades, and the legacy importer - as detailed [here](https://github.com/purescript/registry-dev/pull/669#issuecomment-3730767008)~~

~~The patch is simpler than I though it would be, and I think that's mainly because we only deal with packages that are already in the registry, rather than packages that could possibly be missing. As a result, the package transfer script and the package set updater are deprecated, while the legacy importer stays in place since it's the only way we have to bootstrap the registry from scratch at the moment.~~

EDIT: the above approach didn't work because of the complex race conditions that come from running more code interacting with the registry concurrently with the jobs (that's why we have the queue, right!). So we are going to keep the daily cron jobs in separate scripts, but wire them up so that they hit the API instead of directly editing the Registry.

Somewhere in the middle of this PR I noticed that the matrix tests for detecting a compiler at the start were discarded, so I added them back.


app-e2e/src/Test/Main.purs
```diff
@@ -21,6 +22,10 @@ main :: Effect Unit
 main = do
   env <- mkTestEnv
   runSpecAndExitProcess' config [ consoleReporter ] $ hoistE2E env do
+    -- The scheduler runs at startup and enqueues a bunch of jobs in the DB,
+    -- so we need to run these tests without cleaning out the state first
+    Spec.describe "Scheduler" Scheduler.spec
```
f-f: This was the easiest way to wire this in, not sure if it's the best one though!

app/src/App/API.purs
```diff
@@ -195,11 +195,18 @@ packageSetUpdate details = do
 
   let changeSet = candidates.accepted <#> maybe Remove Update
   Log.notice "Attempting to build package set update."
-  PackageSets.upgradeAtomic latestPackageSet (fromMaybe prevCompiler payload.compiler) changeSet >>= case _ of
-    Left error ->
-      Except.throw $ "The package set produced from this suggested update does not compile:\n\n" <> error
-    Right packageSet -> do
-      let commitMessage = PackageSets.commitMessage latestPackageSet changeSet (un PackageSet packageSet).version
+  PackageSets.upgradeSequential latestPackageSet (fromMaybe prevCompiler payload.compiler) changeSet >>= case _ of
```
f-f, line 203: The package set script used `upgradeSequential` so that it would try out a smaller changeset if the whole failed. I have switched it here so that we keep the same behaviour for all jobs - `upgradeSequential` runs `upgradeAtomic` first anyways, so I don't see any downsides here. What am I missing?

app/src/App/Main.purs
```diff
@@ -8,22 +8,33 @@ import Effect.Aff as Aff
...
+      void $ runEffects env do
+        Log.info "Initializing registry repo..."
+        Registry.readAllMetadata
+      liftEffect do
+        case env.vars.resourceEnv.healthchecksUrl of
+          Nothing -> Console.log "HEALTHCHECKS_URL not set, healthcheck pinging disabled"
+          Just healthchecksUrl -> Aff.launchAff_ $ healthcheck healthchecksUrl
+        Aff.launchAff_ $ withRetryLoop "Scheduler" $ Scheduler.runScheduler env
+        Aff.launchAff_ $ withRetryLoop "Job executor" $ JobExecutor.runJobExecutor env
+        Router.runRouter env
```
f-f: Right, this was a bit annoying - I added another fiber to schedule jobs, and that resulted in two fibers pulling the registry at the same time, and having a race condition around the creation of the folder. I am not sure what's the best way to go at it - I thought about merging the two loops (i.e. pull jobs unless 12h have passed from the last time we run the crons, in which case loop through the registry/github and enqueue stuff, then go back to running jobs), but the package importer might just take a long time since it's hitting github for each package, and we can't afford that kind of slowdown. So for now I am synchronously pulling the registry first thing, but I am not a big fan of this since it doesn't prevent more race conditions in the future.

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+      in
+        Array.last $ Array.sort recentVersions
+
+  pure $ Map.fromFoldable $ Array.catMaybes $ flip map (Map.toUnfoldable allMetadata) \(Tuple name metadata) ->
+    map (Tuple name) $ getLatestRecentVersion metadata
+
+-- | Check for new tags on existing packages and enqueue publish jobs for
+-- | versions not yet published. This allows the registry to automatically
+-- | import new versions of packages that only have legacy manifests.
+scheduleLegacyImports :: Run ServerEffects Unit
```
thomashoneyman: What about naming this “daily publish” instead? While it is capable of supporting legacy formats it’s not really a legacy import, as we’re only picking up new versions of registered packages and presumably over time that will be all normal (non-legacy) packages

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+enqueuePublishJob :: PackageName -> Metadata -> Version -> String -> Run ServerEffects Unit
+enqueuePublishJob name (Metadata metadata) version ref = do
+  -- Check if a publish job already exists for this package version
+  existingJob <- Db.selectPublishJob name version
+  case existingJob of
+    Just _ -> Log.debug $ "Publish job already exists for " <> formatPackageVersion name version
+    Nothing -> do
+      -- Use the lowest compiler from previous version for compatibility,
+      -- falling back to latest if no previous version exists
+      compiler <- case Map.findMax metadata.published of
```
thomashoneyman: This is _probably_ fine but it’s technically possible to publish a package version that dates to an older compiler (like a bug fix to an old series of the package for an earlier compiler), so we have a reliable way to determine the compiler from the resolved packages, as is done in the legacy importer (compilerFromResolutions) and could use that here

app/src/App/API.purs
thomashoneyman: The difference between them is that:

- `upgradeAtomic` should be used when you want to be strict and not accept a partial result, e.g. a manual package set update where the submitter expects everything to work, and if it doesn't work they should be notified with errors so they can fix and resubmit
- `upgradeSequential` should be used when you want to be lenient and include as many packages as possible, even if some fail. It runs the atomic attempt first as an optimization, to try and avoid the slow one-by-one process.

thomashoneyman: I don't really want to lose this behavior, but I'm not sure how best to preserve it.

One option that comes to mind is that we could give up on having the jobs be _identical_ and could modify the DB insert type to include `isScheduled :: Boolean` for cron jobs, so the executor can choose different behaviors based on that — i.e. use `upgradeSequential` for scheduled jobs, and `upgradeAtomic` otherwise. This would be useful for other scheduled jobs like the daily publish too, because we can use this to opt-in to legacy manifest formats.

We would need to change a few `App` files:

- Add `isScheduled INTEGER NOT NULL DEFAULT 0` to `package_set_jobs`
- Add `isScheduled :: Boolean` to the `InsertPackageSetJob` types and `insertPackageSetJobToJSRep` function
- Set `isScheduled: true` in the scheduler when calling insert
- Branch on `isScheduled` in the `API.purs` file when executing a job

And unfortunately also the package set job type would change in the API.V1 to add `isScheduled` to `PackageSetJobData` — but this is perhaps a good thing?

thomashoneyman: A different option would be to add a new `mode :: Maybe (PassFail | BestEffort)` to the package set update API, defaulting to 'PassFail' to preserve the existing behavior for manual updates. This preserves the 'atomic' behavior for manual submissions. But users (including the scheduler) can also specify `BestEffort` (sequential) if they are OK with packages failing and they just want the most possible to be included.

This would have fewer changes overall. What do you think @f-f?

app/src/App/Main.purs
thomashoneyman: Good catch on this race condition. In short, both the `Scheduler` and `JobExecutor` fibers call into the `REGISTRY` effect, which triggers git clone or pulls, and in this specific case when both fibers start together they race to create the same directory, causing failures.

We _could_ merge the two loops. The daily update job could end up taking a long time though, as you said, because it fetches tags from all repos and may get rate-limited. That said, if you just want to get this PR over the line perhaps it's acceptable for now. I wrote out a more full solution below but it's a bit riskier.

thomashoneyman: I think the proper solution here is per-repo locks where we add a mutex per git repository inside the `REGISTRY` handler, and all git operations for a given repo go through this lock.  It has some advantages, but also we start running up against the limitations of `run` as an effects system.

The classic solution would be to use `AVar` as a way to produce locks in `Aff`, and use `Aff.bracket` to acquire the lock, run the action, and release the lock on completion (this is exception-safe, so on exception the lock would also be released).

However, we can't use `Aff.bracket` because it would require the actions to be run in `Aff` and our actions intersperse other effects like `LOG`. You can't enter `Aff` and then "re-enter" `Run` like that. There are workarounds but they're very ugly.

So that leaves us with more manual lock management, which ordinarily is a bummer but in our case is actually not too bad. The basic idea would be:

1. Add per-repo locks to the registry env, keyed by process in registry.purs

The point of keying by process is for cleanup in the case of an exception -- described later.

```purs
data Process = Scheduler | JobExecutor

type RepoLock = { lock :: AVar Unit, owner :: Ref (Maybe ProcessId) }

type RepoLocks = Ref (Map RepoKey RepoLock)

type RegistryEnv = { ..., repoLocks :: RepoLocks }
```

2. Add a `withRepoLock` helper function

```purs
withRepoLock :: forall r a. ProcessId -> RepoLocks -> RepoKey -> Run (LOG + AFF + EFFECT + r) a -> Run _ a
withRepoLock process locks key action = do
  lock <- getOrCreateLock locks key
  Run.liftAff $ AVar.take repoLock.lock
  Run.liftEffect $ Ref.modify_ (_ { owner = Just process }) repoLock
  result <- action
  Run.liftEffect $ Ref.modify_ (_ { owner = Nothing }) repoLock.owner
  Run.liftAff $ AVAr.put unit repoLock.lock
  pure result
```

3. Use the withRepoLock function in the registry handler

```purs
pull key = withRepoLock processId env.repoLocks key do
  -- existing logic
```

4. Clean up locks on startup/job failure

The basic approach to exception safety is that an exception will kill the fiber (fibers can also be killed by the parent, e.g. on a timeout). Then, the executor/scheduler loops get restarted. At that point we can clear any held locks for the process which restarts -- this avoids the case where an exception leaves the lock held, and then it remains stale.

```purs
clearOwnLocks :: ProcessId -> RepoLocks -> Run _ Unit
clearOwnLocks process locksRef = do
  locks <- readLocks
  for_ (Map.toUnfoldable locks :: Array _) \(Tuple key repoLock) -> do
    { owner } <- Run.liftEffect $ Ref.read repoLock
    when (process == Just owner) do
      Log.warn $ "Clearing orphaned lock for " <> process
      -- clear the lock
```

When the loops start or fail, clear the locks

```purs
runJobExecutor env = runEffects env do
  clearOwnLocks JobExecutor env.registryEnv.repoLocks
  Log.info "Starting ..."
  loop
```

When we kill a fiber, same deal:

```purs
jobResult <- liftAff do
  let timeout = Aff.delay (Milliseconds delay) $> Nothing 
  ...
  
case jobResult of
  Nothing -> do
      Log.error "timed out"
      clearOwnLocks JobExecutor ...
      pure false

  Just (Left _) -> do
    Log.error "died"
    clearOwnLocks ...
```

This should work in each scenario:

- Normal completion: the lock is released, as per `withRepoLock`
- Job timeout or error, `clearOwnLocks` runs on the job result.
- Executor/scheduler crashes, gets restarted, `clearOwnLocks` on startup

thomashoneyman: One last thought: clearing locks is for an exceptional scenario where something has gone wrong. We have other consequences besides stale locks that can happen here — for example, what if we were in the middle of a `clone`? Then we'll end up with a corrupted repo and the registry will be borked.

So we might want a more general `cleanup` function that does more than clear locks. It could also run `gitCLI [ "rev-parse", "--git-dir" ]` to check each repo is valid and delete it if it isn't. (On `pull` we already do a force-clean, so stale state is OK).

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+      let
+        payload =
+          { name
+          , location: Just metadata.location
+          , ref
+          , version
+          , compiler
+          , resolutions: Nothing
+          }
+      jobId <- Db.insertPublishJob { payload }
```
thomashoneyman: This is just going to insert a normal publish job, which means it won't work for legacy manifests. In your original comment you'd mentioned relying on signatures as a way to differentiate between a "scheduled" job / maintainer-submitted job and a normal user API call (which can't be legacy). Did that just get overlooked?

thomashoneyman (changes requested):
Noticed a few things that need to be resolved before we merge. Also, I went ahead and removed the daily jobs workflow from the registry repository since this replaces it:
https://github.com/purescript/registry/pull/525

app/src/App/API.purs
f-f: I'm not sure I follow: what's the value in preserving the all-or-nothing behaviour? Wouldn't it be preferrable to always add something to the set if possible?

Re the user flow: we could make it so that if an update submitted by the user only works partially, then they would get information on what succeeded and what failed. This would be possibly more actionable than "your whole thing failed"

app/src/App/Main.purs
f-f: Good writeup! I think we should just add this locking functionality, introducing the scheduler pretty much guarantees race conditions with any jobs that are running.

app/src/App/Server/Scheduler.purs
f-f: Agreed!

app/src/App/Server/Scheduler.purs
f-f: Oh, thanks for the pointer, will use that instead!

app/src/App/Server/Scheduler.purs
f-f: Right, I had a further look at this and it turns out we don't really special case legacy packages anymore, see around here:
https://github.com/purescript/registry-dev/blob/862e6cfc91b01121352f6618a13d1f62513330b0/app/src/App/API.purs#L464

There are a few situations where we do something different if this code is running under the legacy importer (grep for `maybeLegacyIndex`), and I think we don't need to special case any of them.

app/src/App/API.purs
thomashoneyman: Say you want to publish a package set update that adds `pkg-a` and `pkg-b`, but they're incompatible. The package set update fails, and then they go one-by-one and `pkg-a` makes it in, and `pkg-b` is rejected. But if `pkg-b` is a more important package you're now unable to get it into the set without first dropping `pkg-a`. As a maintainer I would prefer to submit a package set update and be notified if my update is invalid, rather than have it go through and then have to fix it after the fact.

app/src/App/Server/Scheduler.purs
thomashoneyman: Whoops, it's actually called `findFirstFromResolutions` and is a helper function inlined, hopefully not too much trouble to move it into the app somewhere and import it back into the importer. Or it may be too tied-in to the importer caches and you want to just create a new function.

```purs
findFirstFromResolutions :: Map PackageName Version -> Run _ (Either (Map Version CompilerFailure) Version)
findFirstFromResolutions resolutions = ...
```

app/src/App/Server/Scheduler.purs
thomashoneyman: Huh! I guess it's already pretty well-integrated. For my and others' future reference, these are the usages in `API.purs`

1. Line 373 -- fallback to registry-archive when GitHub repo is inaccessible: If fetching source fails with `InaccessibleRepo` and this is a legacy import, we fall back to fetching the tarball from the `registry-archive`. This handles packages whose GitHub repos no longer exist but whose tarballs were archived. We can get rid of this logic after the importer is no longer in use, and this isn't needed for publishing new versions of packages.

2. Line 501 -- skip location mismatch validation: Legacy packages may have been transferred after their original publication. The manifest records the old location while metadata has the current one. In legacy mode, this mismatch is logged as a warning rather than throwing an error. In the main registry we don't allow this kind of mismatch. We don't want this logic for publishing new versions.

3. Line 667 -- Auto-fix dependency errors: When unused/missing dependency validation fails, legacy imports get fixed using `conformLegacyManifest` and the provided legacy index. Normal publishes fail immediately with the validation error. We want these publishes to fail, so again OK.

4. Line 780 -- determine all valid compiler versions: After successful publish, legacy imports immediately test the package against all available compilers to populate the `compilers` metadata field. Normal publishes skip this because it's extremely slow; the server handles it via separate `MatrixBuilder` jobs.

So yeah, this all checks out. I guess my only thought, then, is whether we want to add a `Log.warn` or `Log.notice` in here to warn users that their package uses a legacy format, it will be converted to a `purs.json` file, and legacy support will be phased out.

app/src/App/API.purs
thomashoneyman: Maybe this really isn't a big deal and the convenience of just doing `upgradeSequential` is worth it. I'm OK with that. We can always switch later if it's actually a problem.

f-f: Do you have a more concrete scenario for when you would need to drop package `a` in order to get package `b` in?

If you submitted an update with two packages it's fairly safe to assume that you wanted both in - so one is better than nothing? If package `b` doesn't make it then one can iterate on the two packages and release new versions (and/or PR to other packages if those are blocking), and this is work that one needed to do anyways to get packages in.

thomashoneyman: I don't have a good concrete scenario, just a vague memory of package sets being a pain when dealing with a breaking change in a library that affects other packages in the set. Maybe this is actually a non-issue.

f-f: I propose we move forward as is, and reconsider later whether we want to add the flag you proposed - the addition should be backwards compatible anyways

app/src/App/Server/Scheduler.purs
f-f: The line I linked is a notice mentioning that the dependencies are being parsed from a legacy format. We could reword it to be somewhat more aggressive, but actually I would like to just track the amount of packages published this way so we can eventually migrate all new publishes and get rid of this code path entirely. We'll need to include this whenever we get to discussing monitoring infrastructure for all this stuff.

app/src/App/API.purs
thomashoneyman: Works for me

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+                      version = LenientVersion.version result
+                    in
+                      if Set.member version publishedVersions then Nothing -- Already published
+                      else Just { version, ref: tag.name }
+
+            for_ newVersions \{ version, ref } ->
+              enqueuePublishJob name (Metadata metadata) version ref
+
+    -- Delay between packages to spread GitHub API load
+    liftAff $ Aff.delay (Milliseconds 500.0)
```
thomashoneyman: We talked about this briefly in chat, but I'm not sure this is needed. The importer will wait on 403 with exponential backoff until it can resume (rate limiting is lifted); the only reason to introduce this pause is if we _know_ we'll get rate-limited, and waiting on the 403 is slower than sufficiently slowing our fetches to avoid being rate limited in the first place. Without evidence that's the case, and that this is slow enough, I'd rather just rely on the existing behavior.

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+
+  let
+    getLatestRecentVersion :: Metadata -> Maybe Version
+    getLatestRecentVersion (Metadata metadata) =
+      let
+        recentVersions = Array.catMaybes $ flip map (Map.toUnfoldable metadata.published) \(Tuple version { publishedTime }) ->
+          let
+            diff = DateTime.diff now publishedTime
+          in
+            if diff <= limit then Just version else Nothing
```
thomashoneyman: This is a super tiny nit but we don't really use `let/in` anywhere in the codebase, we always use `do/let`

app/src/App/Server/Scheduler.purs
```diff
@@ -0,0 +1,268 @@
...
+
+-- | The scheduler loop runs immediately, then every 24 hours.
+-- | It checks for work that needs to be enqueued (transfers, package set
+-- | updates, legacy imports) and creates the appropriate jobs.
+runScheduler :: ServerEnv -> Aff (Either Aff.Error Unit)
+runScheduler env = runEffects env do
+  Log.info "Starting Scheduler"
+  loop
+  where
+  sleepTime = Milliseconds (1000.0 * 60.0 * 60.0 * 12.0)
```
thomashoneyman: The comment and our discussions have assumed 24 hours, but this looks like only 12 hours? Is that intentional? Should this be 24 hours instead?

db/schema.sql
```diff
@@ -40,6 +40,8 @@ CREATE TABLE matrix_jobs (
 CREATE TABLE package_set_jobs (
   jobId TEXT PRIMARY KEY NOT NULL,
   payload JSON NOT NULL,
+  rawPayload TEXT NOT NULL,
+  signature TEXT,
```
thomashoneyman, line 44: Do we need to add a migration for this? Won't new writes fail due to missing columns if we don't? Or is the intention to delete the migrations altogether, delete the existing database, and start from scratch?

app/src/App/API.purs
thomashoneyman, line 203: I noticed a different issue when scanning through the code — not visible in this PR — that a partial upgrade via `sequential` can result in bumping the version number incorrectly.

Not a big deal, but basically `upgradeSequential` computes the next package-set version from the entire `changeSet` (including failed changes). So if the change set includes a failed major bump and only patch-level updates succeeded, the computed version can still jump major because it saw the failed major change. The resulting package set contents would only reflect patch updates, but the version would indicate a major change that never actually landed.

Ideally we would compute the next version from `succeeded` instead of the `changeSet`. Would that be possible to add to this PR, or do you want to mark this resolved and I can open a separate issue about it?

thomashoneyman: Gave Codex a crack at it too, as an experiment, and it flagged something else entirely — removals may be a problem with `upgradeSequential` unless we switch dependency order for them (not a problem with `upgradeAtomic`):

```
In upgradeSequential, the batch is ordered so dependencies come before dependents. That’s good for updates, but for removals it’s the opposite of what you want. If you remove a dependency first, any remaining dependents still in the set will fail to compile, so the removal gets rejected even if those dependents are also slated for removal later.

Example:
- Change set removes foo and bar, where bar depends on foo.
- Order is foo then bar.
- Removing foo first makes bar uncompilable → removal of foo fails.
- Then bar removal runs, but foo never got removed.

So for removals, you generally want dependents removed before their dependencies, i.e., reverse topological order (or handle removals atomically in a separate phase).
```

db/schema.sql
f-f: I am not sure we should be tracking this file at all actually. In any case, the latest migration (which was not changed here!) hasn't been deployed yet, so we are safe to change things until we merge the whole chunk of work. That migration drops all tables anyways, so should be alright in all scenarios.

app/src/App/Server/Scheduler.purs
f-f: 12h was with the assumption that the package import would take a long time. If it doesn't then 24h should be good.

thomashoneyman: It’s pretty fast. Typically 15-18 minutes and that’s with all the GitHub actions nix setup etc — the actual import is like 3 minutes:

https://github.com/purescript/registry/actions/runs/20674149908

app/src/App/API.purs
thomashoneyman: #716 describes the issue.

thomashoneyman: https://github.com/purescript/registry-dev/pull/717 is the fix.

app/src/App/Server/Scheduler.purs
f-f: fixed

app/src/App/Server/Scheduler.purs
f-f: sounds good - I switched to `do/let`

app/src/App/Server/Scheduler.purs
f-f: yep! I removed this artificial wait

app/src/App/Server/Scheduler.purs
f-f: Applied this idea, but I kept the existing logic for the case where the package doesn't have any dependencies. In that case it still makes sense to try to take the same compiler as the previous version, instead of falling back to the latest one.

app/src/App/Server/Scheduler.purs
f-f: This is now done

app/src/App/Main.purs
f-f: I have applied this suggestion, but I think it needs a bit more work

f-f: we ditched the lock approach entirely in the end

scripts/src/DailyImporter.purs
```diff
@@ -0,0 +1,268 @@
...
+  let formatted = formatPackageVersion name version
+
+  -- Determine the appropriate compiler version
+  compiler <- case Map.findMax metadata.published of
+    Just { key: prevVersion, value: publishedInfo } -> do
+      -- Look up the manifest for the previous version to get its dependencies
+      maybeManifest <- Registry.readManifest name prevVersion
+      case maybeManifest of
+        Just (Manifest manifest) | not (Map.isEmpty manifest.dependencies) -> do
+          -- Use previous version's dependencies to find compatible compilers
```
thomashoneyman: Why use the last version’s deps? We can detect the compiler by inserting purs, solving this package, then popping off purs to get its version (and then pass the resolutions as part of the publish job)

app-e2e/src/Test/E2E/Endpoint/Startup.purs
```diff
@@ -0,0 +1,53 @@
+-- | E2E tests for server startup behavior (non-scheduler).
+-- |
+-- | - checkIfNewCompiler: Detects new compiler and enqueues matrix jobs
+-- |
```
thomashoneyman: I don't think we need to list the test names in the header

app-e2e/src/Test/E2E/Endpoint/Startup.purs
```diff
@@ -0,0 +1,53 @@
...
+
+        matrixJobs = Array.filter isNewCompilerMatrixJob jobs
+
+        -- Get package names from matrix jobs
+        matrixPackages = Array.mapMaybe
+          ( \j -> case j of
+              MatrixJob { packageName } -> Just packageName
+              _ -> Nothing
+          )
+          matrixJobs
```
thomashoneyman: Usually write this as 

```purs
matrixJobs # Array.mapMaybe case _ of
  MatrixJob -> ...
```

app-e2e/src/Test/E2E/Endpoint/Startup.purs
```diff
@@ -0,0 +1,53 @@
...
+
+      -- Should have matrix jobs for packages with no dependencies
+      -- prelude has no dependencies, so it should get a matrix job
+      let preludeName = unsafeFromRight (PackageName.parse "prelude")
```
thomashoneyman: We have a `Utils.unsafePackageName`

app-e2e/src/Test/E2E/Endpoint/Startup.purs
```diff
@@ -0,0 +1,53 @@
...
+      unless (Array.elem preludeName matrixPackages) do
+        Assert.fail $ "Expected matrix job for prelude with compiler 0.15.11, found: "
+          <> show (Array.length matrixJobs)
+          <> " matrix jobs for packages: "
+          <> show (map PackageName.print matrixPackages)
```
thomashoneyman: Should we comma-separate these, i.e. `String.joinWith ", " (map ...)`

app-e2e/src/Test/E2E/Support/Env.purs
```diff
@@ -99,6 +101,10 @@ runE2E env = flip runReaderT env
 -- | Resets: database, git fixtures, storage mock, and logs.
 resetTestState :: E2E Unit
 resetTestState = do
+  -- Wait for any pending jobs to complete before clearing state.
+  -- This is important because startup jobs (like matrix jobs from new compiler
+  -- detection) may still be running when this is called.
+  waitForAllPendingJobs
```
thomashoneyman, line 107: I worry a bit this will slow the tests down significantly, if for instance we want to kick off a few publish jobs to test them but don't care about the matrix jobs completing. Wonder if we can have an `after_` on tests that kills the jobs instead? If you aren't noticing slowdowns in the tests then no big deal.

app-e2e/src/Test/E2E/Support/Env.purs
```diff
@@ -140,13 +147,41 @@ resetGitFixtures = do
   deleteGitClones scratchDir
   where
   resetOrigin dir = do
-    void $ gitOrFail [ "reset", "--hard", "initial-fixture" ] dir
+    -- Try to reset to post-startup tag first, fall back to initial-fixture
+    tag <- hasTag "post-startup" dir
+    let targetTag = if tag then "post-startup" else "initial-fixture"
+    void $ gitOrFail [ "reset", "--hard", targetTag ] dir
```
thomashoneyman, line 153: Could you explain the reasoning for the "post-startup" tag and why we'd want to reset to that? I was thinking we'd always reset to the initial state, making the 'startup' tests essentially a no-op.

app-e2e/src/Test/E2E/Scripts.purs
```diff
@@ -0,0 +1,292 @@
...
+    Spec.it "does not enqueue jobs for already-published versions" do
+      runDailyImporterScript
+      jobs <- Client.getJobs
+
+      -- type-equality v4.0.1 is already published, should NOT have a new job
+      let
+        isDuplicateJob :: Job -> Boolean
+        isDuplicateJob = case _ of
+          PublishJob { packageName, packageVersion } ->
+            packageName == unsafeFromRight (PackageName.parse "type-equality")
```
thomashoneyman: Can use `packageName == Utils.unsafePackageName "type-equality" && packageVersion == Utils.unsafeVersion "4.0.1"` though I'll usually define these as constants in the test file if they're being used a few times

app-e2e/src/Test/E2E/Scripts.purs
```diff
@@ -0,0 +1,292 @@
...
+    # Run.runBaseAff'
+
+  case result of
+    Left err -> liftAff $ Aff.throwError $ Aff.error $ "DailyImporter failed: " <> err
+    Right _ -> pure unit
+
+-- | Run the PackageTransferrer script in Submit mode
+runPackageTransferrerScript :: E2E Unit
+runPackageTransferrerScript = do
+  { stateDir, privateKey } <- ask
```
thomashoneyman: These setup lines are pretty much the same for each script, can they be shared in a helper?

app/fixtures/registry/metadata/prelude.json
```diff
@@ -7,8 +7,7 @@
     "6.0.1": {
       "bytes": 31129,
       "compilers": [
-        "0.15.10",
-        "0.15.11"
+        "0.15.10"
```
thomashoneyman, line 10: What's the reason to drop 0.15.11 for prelude?

app-e2e/src/Test/E2E/Support/Env.purs
f-f: There should be only two matrix tests, and they complete very quickly. No slowdown detected so far!

app-e2e/src/Test/E2E/Support/Env.purs
f-f: The "initial state" doesn't include the result of the matrix jobs that run at startup. In particular, we want `prelude` to be compatible with both `0.15.10` and `0.15.11`, so we get some matrix jobs when publishing too.

app/fixtures/registry/metadata/prelude.json
f-f: As explained above, we need this so that the startup code marks the `0.15.11` compiler as "new" and enqueues matrix jobs for the whole registry

scripts/src/PackageSetUpdater.purs
```diff
@@ -1,192 +1,240 @@
...
-      Nothing -> do
-        Log.info "No packages could be added to the set. All packages failed."
-      Just { failed, succeeded, result } -> do
+    >>= case _ of
+      Left err -> do
+        Console.error $ "Error: " <> err
+        liftEffect $ Process.exit' 1
+      Right _ -> pure unit
+
+type PackageSetUpdaterEffects = (REGISTRY + GITHUB + LOG + RESOURCE_ENV + EXCEPT String + AFF + EFFECT + ())
```
f-f, line 111: I feel like we should have a `REGISTRY_READ` and a `REGISTRY_WRITE` effect, so we can actually ensure that this can't write to the registry.

Thoughts?

scripts/src/PackageSetUpdater.purs
```diff
@@ -1,192 +1,240 @@
...
+            let
+              request :: Operation.PackageSetUpdateRequest
+              request =
+                { payload: PackageSetUpdate payload
+                , rawPayload
+                , signature: Just signature
+                }
+
+            Log.info $ "Submitting package set update..."
+            result <- Run.liftAff $ submitPackageSetJob (registryApiUrl <> "/v1/package-sets") request
```
f-f, line 182: In tests we set `registryApiUrl` to be `localhost:9000/api`, instead of just the server name. We should check if we are being consistent with appending `api` to the production url as well

scripts/src/DailyImporter.purs
f-f: Yeah well, the thing is that I wanted to avoid doing the whole manifest creation dance (fetching all the different files) in the previous iteration since it needed to run fast. I have added it now - we look at the possible manifest files and get a list of dependencies while solving for the compiler

scripts/src/PackageSetUpdater.purs
thomashoneyman: We could do that, separating the write/delete/mirror functions into a different effect, but expose both from the main registry file. I'm not sure we _need_ to, and if we did it would be convenient to have `type REGISTRY r = REGISTRY_WRITE + REGISTRY_READ + r`. Guess I'm ambivalent on this one.

scripts/src/DailyImporter.purs
thomashoneyman: Sorry, I realize I'm calling back to part of the original design that never got implemented — in this comment:
https://github.com/purescript/registry-dev/pull/669#issuecomment-3726751198

We'd discussed making `compiler` optional such that discovery is part of the publish pipeline (if omitted), where we already have the manifest and everything ready to go — so there's no dance. If you make that an optional field then you can just submit the publish job without supplying any compiler or resolutions at all, which simplifies the daily publish a bit at the cost of making the actual publish job take a little longer (it has to solve + compile if you don't provide a compiler + resolutions yourself).

thomashoneyman: Here's the relevant excerpt:

---

A publish job today requires at least the package `name`, `ref`, `compiler`, and `version`. We can determine all fields from the `tag` alone:

- We already know the `name`, because we're only publishing registered packages.
- We already know the `ref` because it is the tag
- We can make the `compiler` argument optional because the importer defines a way to discover the correct compiler for a package by solving its dependencies. (Insert `purs`, solve, then pop `purs` off.) Resolutions are already optional so this isn't incurring a new cost for the registry.
- We can determine the `version` by parsing a tag using `LenientVersion`. If the tag cannot be parsed, we don't attempt to publish.

scripts/src/PackageSetUpdater.purs
thomashoneyman: The default value is defined like this:

```
-- | The Registry API base URL
productionRegistryApiUrl :: URL
productionRegistryApiUrl = "https://registry.purescript.org/api"
```

f-f: Yeah, I saw that, just wanted to bring it up in case that looked weird to you (it does to me, but I'm happy with anything as long as we have a clear convention)

scripts/src/PackageSetUpdater.purs
f-f: I would like that, but I'll spin it off in a different issue since it's not very urgent

f-f: https://github.com/purescript/registry-dev/issues/722

scripts/src/DailyImporter.purs
f-f: oh, I completely missed that. Will attempt this!

scripts/src/PackageSetUpdater.purs
thomashoneyman: I guess it's a little weird, we could drop the `api` prefix and move it into the router if you prefer. I don't have a strong opinion here.

scripts/src/DailyImporter.purs
f-f: this is now implemented. it looks nicer than I expected 😄

scripts/src/DailyImporter.purs
```diff
@@ -168,72 +146,44 @@ runDailyImport mode registryApiUrl = do
...
   let totalSubmitted = Array.foldl (+) 0 submitted
   Log.info $ "Daily Importer complete. Submitted " <> show totalSubmitted <> " publish jobs."
 
 -- | Submit a publish job for a new package version.
--- | Attempts to find a compatible compiler by fetching the new version's manifest
--- | from GitHub and using the solver to find a compiler that works with its
--- | dependencies. Falls back to the lowest compiler from the previous version if
--- | fetching/parsing fails or the package has no dependencies, or to the latest
--- | compiler if no previous version exists.
+-- | The compiler is not specified; the registry API will discover the oldest
```
thomashoneyman: It will actually discover the _newest_ compiler, given that the solver biases towards new versions

scripts/src/DailyImporter.purs
```diff
@@ -168,72 +146,44 @@ runDailyImport mode registryApiUrl = do
...
-
   let
     payload :: Operation.PublishData
     payload =
       { name
       , version
       , location: Nothing -- Use current metadata location at publish time
       , ref
-      , compiler
+      , compiler: Nothing -- Let the API discover the oldest compatible compiler
```
thomashoneyman: same note here.

thomashoneyman (approved):

PR #717, opened 2026-01-15
Compute package set version from successes, reverse-order removals 

### LINKED ISSUE #716: Partial upgrade via `upgradeSequential` can result in bumping the version number incorrectly
I noticed an issue when scanning through #715 in which a partial upgrade via `upgradeSequential` can result in bumping the version number incorrectly.

Not a big deal, but basically `upgradeSequential` computes the next package-set version from the entire `changeSet` (including failed changes). So if the change set includes a failed major bump and only patch-level updates succeeded, the computed version can still jump major because it saw the failed major change. The resulting package set contents would only reflect patch updates, but the version would indicate a major change that never actually landed.

Ideally we would compute the next version from `succeeded` instead of the `changeSet`.

thomashoneyman:
Fixes #716. First, we use success packages only to determine the new package set version. Then, we split package set updates into updates and removals, process updates first in dependency order, and then removals in reverse dependency order.

### DIFF
--- a/app/src/App/Legacy/Manifest.purs
+++ b/app/src/App/Legacy/Manifest.purs
@@ -38,12 +38,12 @@ import Registry.App.Legacy.LenientVersion as LenientVersion
 import Registry.App.Legacy.PackageSet as Legacy.PackageSet
 import Registry.App.Legacy.Types (LegacyPackageSet(..), LegacyPackageSetEntry, LegacyPackageSetUnion, RawPackageName(..), RawVersion(..), RawVersionRange(..), legacyPackageSetCodec, legacyPackageSetUnionCodec, rawPackageNameMapCodec, rawVersionCodec, rawVersionRangeCodec)
 import Registry.Foreign.Octokit (Address, GitHubError(..))
-import Registry.Internal.Codec as Internal.Codec
-import Registry.Location as Location
-import Registry.Owner as Owner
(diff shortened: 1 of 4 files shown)


app/src/App/Legacy/Manifest.purs
```diff
@@ -38,12 +38,12 @@ import Registry.App.Legacy.LenientVersion as LenientVersion
 import Registry.App.Legacy.PackageSet as Legacy.PackageSet
 import Registry.App.Legacy.Types (LegacyPackageSet(..), LegacyPackageSetEntry, LegacyPackageSetUnion, RawPackageName(..), RawVersion(..), RawVersionRange(..), legacyPackageSetCodec, legacyPackageSetUnionCodec, rawPackageNameMapCodec, rawVersionCodec, rawVersionRangeCodec)
 import Registry.Foreign.Octokit (Address, GitHubError(..))
-import Registry.Internal.Codec as Internal.Codec
-import Registry.Location as Location
-import Registry.Owner as Owner
```
thomashoneyman, line 43: Not sure how this formatting even slipped through? Should have failed the purescript formatting lint check, but it didn't 🤷 maybe `purs-tidy check` doesn't respect the IDE sort ordering in its check?

f-f (approved):
Brilliant, thanks!

PR #718, opened 2026-01-16
Make repo locks exception safe

thomashoneyman:
The repo-locking flow in #715 has a couple issues, namely:

- There isn't a guarantee that locks are released when exceptions are thrown in `Aff` (though it does catch run exceptions)
- A stuck git operation can hold the lock forever

All along we wanted to use `Aff.bracket`, which is the typical approach to this kind of design. But this doesn't work with `Run`. In this change I've made a change to lower the action to `Aff`, capture `LOG` effects into a `Ref`, then replay them after the lock is released to preserve logging behavior. This lets us remove the orphan-cleanup `clearOwnLocks`, which was only compensating for the previous non–exception-safe path.

The new helper (`runWithLogs`) uses `Run.runCont` to interpret `LOG`, `AFF`, `EFFECT`, and `EXCEPT String`, not supporting any other effects, and the change is localized to the registry module. Logs are replayed afterwards, and we still throw exceptions if needed, but the lock will be released. The lock timeout currently produces an `Aff` error and a warning log with the owning process name.

I also added a new `Test.Registry.App.Effect.Registry` module which acquires a lock, throws an exception, and makes sure everything's still clean. Might add a couple more tests.

### DIFF
--- a/app/src/App/Effect/Registry.purs
+++ b/app/src/App/Effect/Registry.purs
@@ -256,9 +256,9 @@ withRepoLockTimeout timeout process locks key action = do
   case outcome of
     Nothing -> do
       Log.warn $ "Repo lock timed out for " <> printProcess process
-      Run.liftAff $ Aff.throwError $ Aff.error "Repo lock timed out."
+      Except.throw "Repo lock timed out."
     Just (Left err) ->
-      Run.liftAff $ Aff.throwError err
+      Except.throw $ "Repo action failed: " <> Aff.message err
     Just (Right value) ->
       pure value
   where
--- a/app/src/App/Effect/Registry.purs
+++ b/app/src/App/Effect/Registry.purs
@@ -221,7 +221,7 @@ withRepoLock
    . Process
   -> RepoLocks
   -> RepoKey
-  -> Run (LOG + AFF + EFFECT + EXCEPT String + r) a
+  -> Run (LOG + AFF + EFFECT + EXCEPT String + ()) a
   -> Run (LOG + AFF + EFFECT + EXCEPT String + r) a
 withRepoLock = withRepoLockTimeout (Milliseconds 60_000.0)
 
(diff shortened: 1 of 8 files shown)


app/src/App/Effect/Registry.purs
```diff
@@ -175,16 +178,17 @@ data Process
...
-  show JobExecutor = "JobExecutor"
-  show API = "API"
-  show ScriptLegacyImporter = "ScriptLegacyImporter"
-  show ScriptPackageDeleter = "ScriptPackageDeleter"
-  show ScriptSolver = "ScriptSolver"
-  show ScriptVerifyIntegrity = "ScriptVerifyIntegrity"
-  show ScriptCompilerVersions = "ScriptCompilerVersions"
-  show ScriptArchiveSeeder = "ScriptArchiveSeeder"
+printProcess :: Process -> String
+printProcess = case _ of
```
thomashoneyman, line 182: We don't use `show` in the code base

app/src/App/Effect/Registry.purs
```diff
@@ -211,34 +215,82 @@ getOrCreateLock locksRef key = do
...
+
+-- | Acquire a repository lock, run an action, and release the lock.
+-- | The lock prevents concurrent access to the same repository.
+withRepoLockTimeout
+  :: forall r a
+   . Milliseconds
+  -> Process
+  -> RepoLocks
+  -> RepoKey
+  -> Run (LOG + AFF + EFFECT + EXCEPT String + ()) a
```
thomashoneyman, line 236: This is a closed row restricted just to the actions we intend to 'lower' to `Aff`. If we wanted to expand locking outside this module we'd have to add explicit handling for anything else we want to run to `Aff`.

app/src/App/Effect/Registry.purs
```diff
@@ -211,34 +215,82 @@ getOrCreateLock locksRef key = do
...
+  -- It isn't possible to run exception-safe Aff code like `bracket` within
+  -- the extensible effects system. For the actions we need to support
+  -- behind a lock we only need to support the LOG effect, so we lower to
+  -- Aff, run the lock-guarded code safely, and aggregate the logs to be
+  -- flushed afterwards.
+  { logs, outcome } <- Run.liftAff do
+    logsRef <- liftEffect $ Ref.new []
+    outcome <- withRepoLockAff repoLock timeout (runWithLogs logsRef action)
+    logs <- liftEffect $ Ref.read logsRef
+    pure { logs, outcome }
```
thomashoneyman, line 250: cc: @natefaubion I couldn't think of a better way to do this, but if you have ideas I'm all ears

thomashoneyman:
Closing as we are going to drop concurrency after all.

PR #723, opened 2026-01-18
Start verifying tarball hashes when running tests

f-f:
I started thinking about how Spago would handle hashes once we regen the Registry - it turns out all is good on that front, but then I realised that we are not really checking tarball hashes when running tests here. In fact, we even had a TODO about it!

So I threaded through the bits that we needed, but tests did not pass - tried a few things, but then had a proper look at the tarball that we generate. It turns out that we were including the timezone in the metadata, even though we were using `--mtime=1970-01-01 00:00:Z` as suggested by `reproducible-builds.org`, which should refer to UTC.
Some places on the internet suggest just zeroing it with `--mtime=@0`, and that in fact does clear out timezone data from the tarball 🤷 

---

Before this change, asking for a listing of the tar contents (in UTC), yielded the wrong time:
```
$ TZ=UTC gtar -tvf app/fixtures/registry-storage/effect-4.0.0.tar.gz
drwxr-xr-x 0/0               0 1969-12-31 23:00 effect-4.0.0/
-rw-r--r-- 0/0            1455 1969-12-31 23:00 effect-4.0.0/LICENSE
-rw-r--r-- 0/0             434 1969-12-31 23:00 effect-4.0.0/bower.json
-rw-r--r-- 0/0             223 1969-12-31 23:00 effect-4.0.0/purs.json
drwxr-xr-x 0/0               0 1969-12-31 23:00 effect-4.0.0/src/
drwxr-xr-x 0/0               0 1969-12-31 23:00 effect-4.0.0/src/Effect/
-rw-r--r-- 0/0             598 1969-12-31 23:00 effect-4.0.0/src/Effect/Class.purs
-rw-r--r-- 0/0            5048 1969-12-31 23:00 effect-4.0.0/src/Effect/Uncurried.js
-rw-r--r-- 0/0           12745 1969-12-31 23:00 effect-4.0.0/src/Effect/Uncurried.purs
-rw-r--r-- 0/0              67 1969-12-31 23:00 effect-4.0.0/src/Effect/Unsafe.js
-rw-r--r-- 0/0             226 1969-12-31 23:00 effect-4.0.0/src/Effect/Unsafe.purs
-rw-r--r-- 0/0             826 1969-12-31 23:00 effect-4.0.0/src/Effect.js
-rw-r--r-- 0/0            2596 1969-12-31 23:00 effect-4.0.0/src/Effect.purs
```

The tarball was created locally (when I was in UTC+1), with `--mtime=1970-01-01 00:00:Z`

We can check that the [PAX headers](https://www.gnu.org/software/tar/manual/html_node/PAX-keywords.html) (gotta love POSIX) do set a negative `mtime`, for some reason:
```
$ python3 -c "import tarfile; t=tarfile.open('app/fixtures/registry-storage/effect-4.0.0.tar.gz'); print(t.getmembers()[0].pax_headers)"
{'mtime': '-3600'}
```

After this change, we have correct timestamps - tarball created locally with `--mtime=@0`, and my locale is in UTC+2, but it doesn't show:
```
$ TZ=UTC gtar -tvf app/fixtures/registry-storage/effect-4.0.0.tar.gz
drwxr-xr-x 0/0               0 1970-01-01 00:00 effect-4.0.0/
-rw-r--r-- 0/0            1455 1970-01-01 00:00 effect-4.0.0/LICENSE
-rw-r--r-- 0/0             434 1970-01-01 00:00 effect-4.0.0/bower.json
-rw-r--r-- 0/0             242 1970-01-01 00:00 effect-4.0.0/purs.json
drwxr-xr-x 0/0               0 1970-01-01 00:00 effect-4.0.0/src/
drwxr-xr-x 0/0               0 1970-01-01 00:00 effect-4.0.0/src/Effect/
-rw-r--r-- 0/0             598 1970-01-01 00:00 effect-4.0.0/src/Effect/Class.purs
-rw-r--r-- 0/0            5048 1970-01-01 00:00 effect-4.0.0/src/Effect/Uncurried.js
-rw-r--r-- 0/0           12745 1970-01-01 00:00 effect-4.0.0/src/Effect/Uncurried.purs
-rw-r--r-- 0/0              67 1970-01-01 00:00 effect-4.0.0/src/Effect/Unsafe.js
-rw-r--r-- 0/0             226 1970-01-01 00:00 effect-4.0.0/src/Effect/Unsafe.purs
-rw-r--r-- 0/0             826 1970-01-01 00:00 effect-4.0.0/src/Effect.js
-rw-r--r-- 0/0            2596 1970-01-01 00:00 effect-4.0.0/src/Effect.purs
```

And no PAX headers:
```
$ python3 -c "import tarfile; t=tarfile.open('app/fixtures/registry-storage/effect-4.0.0.tar.gz'); print(t.getmembers()[0].pax_headers)"
{}
```


app/src/App/Effect/Storage.purs
```diff
@@ -132,11 +140,11 @@ handleS3 env = Cache.interpret _storageCache (Cache.handleFs env.cache) <<< case
...
-  Download name version path reply -> map (map reply) Except.runExcept do
+  Download name version path integrity reply -> map (map reply) Except.runExcept do
     let package = formatPackageVersion name version
     buffer <- Cache.get _storageCache (Package name version) >>= case _ of
       Nothing -> do
-        buffer <- downloadS3 name version
+        buffer <- downloadS3 name version integrity
         Cache.put _storageCache (Package name version) buffer
         pure buffer
       Just cached ->
```
thomashoneyman: Should we include the checks when the tarball is retrieved from storage, too? Without verification, if the cached tarball is old and the metadata hash has changed then this would be silently accepted, but presumably we'd want to evict the cache and force a fresh download instead.

thomashoneyman (approved):
Nice catch!

app/src/App/Effect/Storage.purs
f-f: Ah yeah, good idea! I'll fix this up

f-f: Done!

Issue #742, opened 2026-02-01
Create a Dashboard for the Registry API


f-f:
The Registry API is not made for humans, and we would need a nice dashboard to allow us to monitor the jobs at a glance.

Scope sketch:

## API (Existing)

| Endpoint | Parameters | Notes |
|----------|------------|-------|
| `GET /api/v1/jobs` | `since` (ISO8601), `include_completed` (bool) | Returns max 100 jobs, no logs |
| `GET /api/v1/jobs/{jobId}` | `level` (LogLevel), `since` (ISO8601) | Full job with logs |

All filtering beyond time/completion would be done client-side.

### 1. Jobs List

**Server-side controls:**
- Time range selector (last hour, 24h, week, custom) → `since` param
- Include completed toggle → `include_completed` param

**Client-side filters:**
- Job type (publish, unpublish, transfer, matrix, packageset)
- Package name (substring match)
- Package version
- Compiler version (for matrix jobs)
- Status (pending, running, succeeded, failed)

**Job row displays:**
- Job type (color-coded?)
- Package name + version (where applicable)
- Status indicator
- Created timestamp

**Other:**
- Click row → navigate to job detail
- Auto-refresh toggle
- Older/Newer navigation (shifts time window)

### 2. Job Details

**Header:**
- Job type, package name, version, compiler (where applicable)
- Job ID
- Status (pending/running/succeeded/failed)

**Timestamps:**
- Created, started, finished
- Computed durations (wait time, run time)

**Payload:**
- JSON viewer (collapsible, copyable)

**Logs:**
- Level filter dropdown (DEBUG/INFO/WARN/NOTICE/ERROR)
- Log entries with timestamp, level, message
- Code blocks in messages rendered appropriately
- "Load more" button for pagination (uses `since` = last log timestamp)
- Auto-refresh toggle
- Copy logs button

f-f:
some other ideas that would be nice to have:
- graphs! e.g. to see how job submissions are concentrated in time, or to visualise how long the queue is at any given time
- a compiler matrix, that shows at a glance how compatible a compiler is.

afcondon:
![Image](https://github.com/user-attachments/assets/f074622a-996c-43a7-b9ac-eee78b4e7b68)

PR #752, opened 2026-02-07
Add 'until' to API routes that list things

f-f:
The fact that having only the `since` parameter in the API routes was not very ergonomic already came up in previous discussions with @thomashoneyman, but trying to put together a dashboard for the logs (as per #742) cemented this idea.

Here we add an `until` parameter, so that the default when neither `since` nor `until` are provided is that we fetch the last items until now, which I think is a sensible default.

Bonus: we add a CORS middleware as well here, also needed for the dashboard.

### DIFF
--- a/app/src/App/SQLite.js
+++ b/app/src/App/SQLite.js
@@ -168,14 +168,24 @@ export const selectPackageSetJobByPayloadImpl = (db, payload) => {
   return stmt.get(payload);
 };
 
-const _selectJobs = (db, { table, since, includeCompleted }) => {
+const _selectJobs = (db, { table, since, until, includeCompleted }) => {
   let query = `
     SELECT job.*, info.*
     FROM ${table} job
     JOIN ${JOB_INFO_TABLE} info ON job.jobId = info.jobId
-    WHERE info.createdAt >= ?
+    WHERE 1=1
   `;
-  let params = [since];
+  let params = [];
+
+  if (since != null) {
+    query += ` AND info.createdAt >= ?`;
+    params.push(since);
+  }
+
+  if (until != null) {
+    query += ` AND info.createdAt <= ?`;
+    params.push(until);
+  }
 
   if (includeCompleted === false) {
     query += ` AND info.finishedAt IS NULL`;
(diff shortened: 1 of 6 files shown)


app/src/App/Server/Router.purs
```diff
@@ -112,22 +121,29 @@ router { route, method, body } = HTTPurple.usingCont case route, method of
       _ ->
         HTTPurple.badRequest "Expected transfer operation."
 
-  Jobs { since, include_completed }, Get -> do
-    now <- liftEffect nowUTC
-    let oneHourAgo = fromMaybe now $ DateTime.adjust (negateDuration (Hours 1.0)) now
+  Jobs { since, until: until', include_completed }, Get -> do
+    -- If neither since nor until is provided, default until to now
+    until <- case since, until' of
+      Nothing, Nothing -> Just <$> liftEffect nowUTC
```
thomashoneyman: Right now when both `since` and `until` are missing you only default `until = now`. With `_selectJobs` ordering `ASC + LIMIT 100`, that returns the oldest 100 jobs, not the latest. 

If the intent is ‘latest until now’, then should we defaulting `since` to a recent window only when both bounds are absent (e.g., last hour), or change the SQL to select the latest 100 (`ORDER BY createdAt DESC LIMIT 100`, perhaps wrapped to return `ASC`)?

app/src/App/SQLite.js
```diff
@@ -168,14 +168,24 @@ export const selectPackageSetJobByPayloadImpl = (db, payload) => {
...
 };
 
-const _selectJobs = (db, { table, since, includeCompleted }) => {
+const _selectJobs = (db, { table, since, until, includeCompleted }) => {
   let query = `
     SELECT job.*, info.*
     FROM ${table} job
     JOIN ${JOB_INFO_TABLE} info ON job.jobId = info.jobId
-    WHERE info.createdAt >= ?
+    WHERE 1=1
```
thomashoneyman: What's this for?

app/src/App/SQLite.js
```diff
@@ -168,14 +168,24 @@ export const selectPackageSetJobByPayloadImpl = (db, payload) => {
...
+
+  if (since != null) {
+    query += ` AND info.createdAt >= ?`;
+    params.push(since);
+  }
+
+  if (until != null) {
+    query += ` AND info.createdAt <= ?`;
+    params.push(until);
+  }
```
thomashoneyman: The PureScript comment says the log window is half‑open [since, until), but this SQL uses `timestamp <= ?`. Isn't that going to duplicate log lines on the boundary between adjacent windows?

Should we change this to `<` or if the comment is mistaken, fix the comment? They're inconsistent now. (Same goes for the other use of `until` in this file.

f-f: Ah yeah, makes sense

app/src/App/SQLite.js
f-f: Either of since/until could be null, and they get concatenated with `AND` if they are not - in principle we don't know which of them is going to be there so having a "base" where clause lets us just concat them with the `AND`

app/src/App/Server/Router.purs
f-f: ah yes, this makes sense

f-f:
@thomashoneyman good catch with that review – I have reworked the design so it's cleaner:
- the range is now `[since,until)` as per the comments
- I realised that it made no sense to leave since/until unspecified since we know the lower bound for `since` (the registry launch) and the upper bound for `until` (now, as we can't have logs/jobs in the future), so I made them only optional at the API boundary, but they are always known at the database layer. This makes for simpler queries
- the missing piece was the "ordering" parameter, clients need to be able to pass that in. So now we default to `ASC` (for backwards compatibility), but clients can pass `DESC` as well

thomashoneyman (approved):

PR #754, opened 2026-02-16
Reject deprecated SPDX licenses, but be lenient with committed LICENSE files

### LINKED ISSUE #753: License detection outputs deprecated SPDX identifiers, blocking AGPL packages from Pursuit
When publishing a package with a standard AGPL-3.0 LICENSE file, the registry's license detection reports `AGPL-3.0` — which is a deprecated SPDX identifier. This creates a catch-22:

- If the manifest says `AGPL-3.0-only` or `AGPL-3.0-or-later` (the valid SPDX replacements), the registry rejects it as a mismatch against the detected `AGPL-3.0`
- If the manifest says `AGPL-3.0` to match detection, the registry accepts it but Pursuit rejects it as invalid SPDX

So there's no license value that both the registry and Pursuit will accept.

Registry logs when using `AGPL-3.0-only`:
```
‼ License mismatch detected: manifest has 'AGPL-3.0-only' but detected AGPL-3.0
```

Pursuit error when using `AGPL-3.0`:
```
The license specified in package manifest is not a valid SPDX license expression.
```

thomashoneyman:
Fixes #753. In that issue, @philippedev101 noted that license validation has a catch-22 in that:

- The registry reads licenses specified in the manifest and cross-references them with the licenses discovered in the packaged code by `licensee`. Deprecated licenses like `AGPL-3.0` are allowed, and the licenses must match.
- The registry publishes to Pursuit via `purs publish`, which requires only current SPDX identifiers. Deprecated licenses like `AGPL-3.0` are rejected.
- The `licensee` tool will emit deprecated SPDX ids like `AGPL-3.0` when it sees an AGPL license file, rather than their current / canonical forms. If the manifest specifies `AGPL-3.0-only`, which is correct and compatible with the license file, this still gets rejected by the registry due to the "mismatch".

Therefore, publishing would fail in one place or the other with no value that satisfies both. This PR fixes that by being strict for manifest parsing, but lenient with discovered `licensee` licenses:

- `Registry.License.parse` stays strict for user-authored licenses (current canonical SPDX only)
- Detected licenses are canonicalized leniently first (`License.canonicalizeDetected`), then parsed strictly. By "canonicalize" I mean licenses with an unambiguous path to be moved from a deprecated to a current id are updated.
- Ambiguous deprecated IDs fail explicitly instead of being silently dropped (as happened when e.g. parsing 

Implementation-wise, I made a few changes:

- `License` is now stored as a parsed tree (not opaque text), and `print` always emits canonical form
- `extractIds` works from that stored tree now, so there's no ability to fail
- The deprecated-ID canonicalization logic lives in PureScript with a small deterministic strategy: we have a tiny rename map and `-only` and `-or-later` rewrites.

On the app side of things, we have clear errors for when entries are detected but cannot be canonicalized, and we no longer ignore bad detected values in e.g. bower.json manifest files.


app/src/App/Legacy/Manifest.purs
```diff
@@ -74,9 +75,18 @@ bowerfileToPursJson
...
-    [] -> Left "No valid SPDX license found in bower.json"
-    multiple -> Right $ License.joinWith License.And multiple
+  let
+    parsedLicenses = license <#> \rawLicense ->
+      lmap (\err -> "  - " <> rawLicense <> ": " <> err) $ License.parse rawLicense
+    { fail: parseErrors, success: validLicenses } = partitionEithers parsedLicenses
+
+  parsedLicense <-
+    if not (Array.null parseErrors) then do
+      Left $ "Invalid SPDX license(s) in bower.json:\n" <> String.joinWith "\n" parseErrors
```
f-f: Do we want to error out even in the case where we might have some valid identifiers from the bowerfile?

If yes then we should have an explanation comment here, because I found it surprising

app/src/App/API.purs
```diff
@@ -1344,44 +1348,55 @@ validateLicense packageDir manifestLicense = do
...
         parsedLicenses :: Array License
-        parsedLicenses = Array.mapMaybe (hush <<< License.parse) detectedStrings
+        parsedLicenses = parsedDetectedLicenses # Array.mapMaybe \(Tuple _ parsed) -> hush parsed
 
       Log.debug $ "Detected licenses: " <> String.joinWith ", " detectedStrings
 
-      if Array.null parsedLicenses then do
+      if not (Array.null parseFailures) then do
+        Log.warn "Some detected licenses could not be canonicalized."
+        pure $ Just $ LicenseParseError parseFailures
```
f-f: Same as above – do we want to fail here if just some license in the set doesn't parse?

app/src/App/Legacy/Manifest.purs
thomashoneyman: Yeah, on second thought, this is a best effort cross-referencing so we should retain the existing behavior in which we drop licenses which fail parsing

thomashoneyman:
I double-checked with the strict license decoding, and 44 manifests fail with it. Can't break the manifest decoding, of course! Here are a set of packages which would fail strict parsing.

```
bookhound: LGPL-2.1 AND LGPL-2.1-only
error, jarilo, undefined, yayamll, several idiomatic-node-*: AGPL-3.0
matrices, emmet: LGPL-3.0
smolder, test-unit, sized-vectors, sized-matrices: LGPL-3.0+
nano-id: GPL-3.0 AND MIT
run-external-state: GPL-3.0 AND GPL-3.0-or-later
ordered-set@0.2.0: AGPL-3.0-or-later AND AGPL-3.0
```

So instead I'm adjusting to strict-on-publish, lenient-on-decode, to allow these existing manifests to parse.

thomashoneyman:
I'm not 100% happy with this because it's more code than I'd like to add, but on the other hand dealing with licenses is kind of a pain and it's not a huge surprise. The worst part is the split between `License.parse` as used to deal with manifests which may have deprecated licenses already listed in the registry and `parseCanonical` as is used to only accept _new_ licenses with non-deprecated SPDX identifiers. That makes this a little bit brittle because we have to be careful about which we use. Open to suggestions!

f-f (approved):

f-f:
@thomashoneyman yeah I agree - the code is a bit sprawly, but I wouldn't know how to make it better; preserving license parsing trees is a big chunk of the change, and it sounds like we need those if we want to print in canonical form.

I would say we should merge this as is because it fixes a real bug, we can have another pass at some point to improve it.

PR #761, opened 2026-03-21
Add a dashboard to visualise the Job queue and explore logs

### LINKED ISSUE #742: Create a Dashboard for the Registry API

f-f:
Fix #742 - we add a Halogen dashboard to easily access the Job Queue and see the logs for each job.

The dashboard is deployed on GitHub Pages. I thought we should have a preview for each PR that changes the dashboard as well, so I added [this](https://github.com/rossjrw/pr-preview-action), which is unfortunately not compatible with the [default publish action](https://github.com/actions/deploy-pages), so I used [this one instead](https://github.com/JamesIves/github-pages-deploy-action), which is compatible.

A notable change is that I had to split the SSH module, so that we could use `Registry.Operation` in the frontend modules.


.github/workflows/dashboard.yml
```diff
@@ -20,17 +20,13 @@ jobs:
...
-
-      - name: Setup Nix cache
-        if: github.event.action != 'closed'
-        uses: DeterminateSystems/magic-nix-cache-action@v2
+        run: npm install --no-save purescript@0.15.15 spago@1.0.3
 
       - name: Build dashboard
         if: github.event.action != 'closed'
-        run: nix develop --command npm run dashboard:build
+        run: npm run dashboard:build
```
thomashoneyman, line 35: What's the reason not to use Nix here? Is it slow, failing, something else?

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          node-version: 22
+
+      - name: Install tools
+        if: github.event.action != 'closed'
+        run: npm install --no-save purescript@0.15.15 spago@1.0.3 esbuild
+
+      # TODO: this is temporary until we fix a bug in spago for bundling single packages:
+      # there is a `purs graph` call that fails when doing that because it's trying to use
+      # all the dependencies of the package, which might not be there. So we build everything.
```
thomashoneyman: Was this fixed in https://github.com/purescript/spago/commit/44649c11d017194de7cff18ba11dda922271747c

thomashoneyman:
If we expect agents to be working on the dashboard is there anything specific you think should be added in a `AGENTS.md` (and symlinked `CLAUDE.md`) in the `dashboard` directory so that agents working there pick it up? Anything you noticed you needed to correct regarding Halogen, library choices, style, etc.?

thomashoneyman:
> A notable change is that I had to split the SSH module, so that we could use Registry.Operation in the frontend modules.

Is this outdated? I didn't see changes to the `SSH` module in the commits here.

thomashoneyman (approved):
Approving with non-blocking comments 🙏

.github/workflows/dashboard.yml
f-f: Yeah I tried with Nix but it's terribly slow because we have no cache. I think we're good just skipping Nix here because it's a fairly self-contained part of the project.

.github/workflows/dashboard.yml
f-f: Right, let me cut a release.

f-f:
> Anything you noticed you needed to correct regarding Halogen, library choices, style, etc.?

Actually nothing at all, the thing is much better at Halogen than I ever was 🙂 

> Is this outdated? I didn't see changes to the SSH module in the commits here.

No changes in behaviour, we just extract a part of `Registry.SSH` into `Registry.SSH.Signature` so the `Registry.Operation` can import just that and the frontend does not need to have any node dependencies.

PR #763, opened 2026-03-22
Fix matrix job cascading for new compiler releases

f-f:
In #758 we added support for the newly released compiler `0.15.16` - that was supposed to trigger a global recompilation with the new compiler in topological order; however only packages with no dependencies were upgraded (starting from [this commit](https://github.com/purescript/registry/commit/7aae5eb1a89d3fbf5b1cb49edb5ed4da8eea8faf)), and the cascading to compile their dependents was never triggered.

This turned out to be due to _two_ separate bugs in `App.Server.MatrixBuilder`:
1. the [version lookup](https://github.com/purescript/registry-dev/blob/54410455079a0ac3313251a469549af597cee551/app/src/App/Server/MatrixBuilder.purs#L220) was supposed to reference `manifest.version`
2. the [detection of the new compiler in the metadata](https://github.com/purescript/registry-dev/blob/54410455079a0ac3313251a469549af597cee551/app/src/App/Server/MatrixBuilder.purs#L222) was wrong - the correct one is `all (_ /= compiler) compilers`

We fix those issues, and add tests to ensure that the new behaviour is correct:
- in `Test.Registry.App.Server.MatrixBuilder` where we check that the cascading happens in a mocked setting
- in `Test.E2E.Endpoint.Startup` where we check the same thing but in an end-to-end setting

In the process of implementing the latter test, I figured there was a bug in `Registry.ManifestIndex.delete`: in the mega-patch #669 we made insertion in the index strict (with the flag `IncludeRanges`/`ConsiderRanges`), but `delete` was not updated to take those into account, which caused an issue if trying to delete a version for a package that had more than one version in the index, and ultimately allowing the deletion of a version that rendered some of the version bounds in the index impossible to solve and left the index in an invalid state.
We also fix that bug here, and add tests to verify that the change is correct.


app/test/Test/Assert/Run.purs
```diff
@@ -139,6 +140,14 @@ runBaseEffects = do
     >>> Except.catch (\err -> Run.liftAff (Aff.throwError (Aff.error err)))
     >>> Run.runBaseAff'
 
+-- | For testing Run functions that only need the REGISTRY effect.
+runRegistryMock :: forall a. Ref (Map PackageName Metadata) -> Ref ManifestIndex -> Run (EXCEPT String + LOG + REGISTRY + AFF + EFFECT + ()) a -> Aff a
+runRegistryMock metadataRef indexRef =
+  Registry.interpret (handleRegistryMock { metadataRef, indexRef })
+    >>> Log.interpret (\(Log _ _ next) -> pure next)
+    >>> Except.catch (\err -> Run.liftAff (Aff.throwError (Aff.error err)))
```
thomashoneyman: This can reuse `runBaseEffects` if we're just layering on a registry interpreter, e.g. `Registry.interpret ... >>> runBaseEffects`

app/test/App/API.purs
```diff
@@ -174,10 +174,12 @@ spec = do
         copyFixture "registry"
         copyFixture "registry-storage"
         copyFixture "github-packages"
-        -- We remove effect-4.0.0.tar.gz since the unit test publishes it from
-        -- scratch and will fail if it's already in storage. We have it in
-        -- storage for the separate integration tests.
+        -- We remove effect fixtures since the unit test publishes effect from
+        -- scratch and will fail if it's already registered. We have these in
+        -- fixtures for the separate integration tests (cascade testing).
```
thomashoneyman: No _only_ for cascade testing, but maybe that's an overly pedantic nit

app/test/App/Server/MatrixBuilder.purs
```diff
@@ -0,0 +1,109 @@
...
+      , ref: Nothing
+      }
+
+  -- | Set up a test scenario where prelude@6.0.0 (no deps) has completed its
+  -- | matrix job for 0.15.11, and effect@4.0.0 (depends on prelude) has the
+  -- | given compilers in its metadata.
+  setup effectCompilers = do
+    let
+      index = Utils.fromRight "Failed to build ManifestIndex"
+        $ ManifestIndex.insert ManifestIndex.ConsiderRanges preludeManifest ManifestIndex.empty
```
thomashoneyman: personal preference of `do` over `$` in these circumstances, e.g.

```
      index = Utils.fromRight "Failed to build ManifestIndex" do
        ManifestIndex.insert ...
          >>= ManifestIndex.insert ...
```

as the alternating of a single-char infix and then 3-char infix is a little visually jarring, but this is _so minor_ :)

thomashoneyman (approved):
A couple of trivial nits but 👍 overall, good work on the catch!

app/test/App/Server/MatrixBuilder.purs
f-f: we like all the scales 🙂

PR #767, opened 2026-03-30
Suggest fixes on dashboard PR

thomashoneyman:
@f-f These are some proposed issues and fixes for the dashboard branch — each commit has a description of why + the fix.


f-f:
I cherry-picked 9dd9c35dd9551139f6eaf25cb4a003df0bc70f5e but I wouldn't like the other one - we wouldn't want rebuilds of the dashboard for PRs that change just files in `lib` for example, as they are unlikely to affect the UI.

PR #775, opened 2026-07-21
Fix manifest description length validation

### LINKED ISSUE #773: spago publish accepts a manifest whose description exceeds the registry 300-char limit, producing an index entry spago build can't decode
## Summary

`spago publish` accepts a package whose `description` exceeds the registry
`Manifest` schema's 300-character `LimitedString` limit. The publish succeeds and
the version appears in the metadata, but the manifest written to the **registry
index** violates the schema, so consumer-side `spago build` cannot decode it.
Because the whole package entry fails to read, the package becomes uninstallable.

Publish-side validation and consume-side decoding disagree about what is a valid
manifest.

## Reproduction

`harmonia@0.1.0` (`afcondon/purescript-harmonia`) is a live, minimal reproducer.

1. A package whose `spago.yaml` `description` is 432 characters.
2. `spago publish` → **succeeds**; the version appears in metadata.
3. From a consumer depending on it (`extraPackages: { harmonia: 0.1.0 }`),
   `spago build --refresh`:

```
‼ Could not read package manifests for 'harmonia' from index. Error: $.description: Could not decode Manifest:
  LimitedString: Exceeds limit of 300 characters.

✘ The following packages do not exist in the package index:
  - harmonia
```

## Evidence

- Metadata shows `0.1.0` published (and it compiled on every 0.15.x):
  <https://raw.githubusercontent.com/purescript/registry/main/metadata/harmonia.json>
- The registry-index manifest stores a **432-char** `description`:
  <https://raw.githubusercontent.com/purescript/registry-index/main/ha/rm/harmonia>

## Expected

`spago publish` (or the registry pipeline it submits to) should **reject** a
manifest whose `description` exceeds the 300-char limit, with a clear error,
*before* writing it to the index — so that anything which publishes is always
decodable by consumers.

## Impact

- The published version is uninstallable from the registry.
- The invalid line appears to poison the **entire** package index entry: a later
  corrected version (e.g. `0.1.1`) would still leave the bad `0.1.0` line in
  `ha/rm/harmonia`, so the entry stays unreadable until `0.1.0` is unpublished.
- Recovery is awkward: `spago` exposes `registry transfer` but no `unpublish`
  (only `spago auth` mentions "transfer and unpublish operations"), so removing
  the bad version needs the registry's own process.

## Environment

- spago 1.0.3
- purs 0.15.x
- macOS (darwin 24.6.0)

## Note

The root cause may live in the registry's publish validation
(`purescript/registry` / the registry-lib `Manifest` codec) rather than in spago
itself — feel free to transfer. Filing here because `spago publish` is the
user-facing entry point that accepted the invalid manifest.

thomashoneyman:
Fixes #773. Manifest descriptions were limited to 300 characters when decoding, but manifests created from `spago.yaml` used the description string directly without validating its length. This allowed the registry to publish a manifest that it could write to the registry index but package managers could not decode, making the affected package unavailable to consumers.

This PR raises the description limit to 500 characters and introduces a `LimitedString` type that tracks its maximum length at the type level. Manifest descriptions now use `LimitedString 500` and are validated when converting from `spago.yaml` or legacy manifests, so an invalid description cannot be written to the index. The existing 300-character unpublish reasons use the same type, replacing the previous decode-only limited string codec.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -389,7 +389,7 @@ All packages in the registry contain a `purs.json` manifest file in their root d
 - `location`: a valid [`Location`](#location)
 - `ref`: a `string` representing the reference (e.g., a Git commit or Git tag) at the `location` that was used to fetch this version's source code
 - `owners` (optional): a non-empty array of [`Owner`](#owner)
-- `description` (optional): a description of your library as a plain text string, not markdown, up to 300 characters
+- `description` (optional): a description of your library as a plain text string, not markdown, up to 500 characters
 - `includeFiles` (optional): a non-empty array of globs, where globs are used to match file paths (in addition to the `src` directory and other [always-included files](#always-included-files)) that you want included in your package tarball
 - `excludeFiles` (optional): a non-empty array of globs, where globs are used to match file paths in your package source to exclude them (in addition to the [always-ignored files](#always-ignored-files)) from your package tarball.
 - `dependencies`: dependencies of your package as key-value pairs where the keys are [`PackageName`](#packagename)s and values are [`Range`](#range)s; this is a required field, but if you have no dependencies you can provide an empty object.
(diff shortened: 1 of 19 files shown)


afcondon:
Happy to have enabled you to strengthen Registry and sad to have made extra work for you!

thanks!

f-f (approved):
The type-level limit is brilliant. Good stuff 🙂

PR #783, opened 2026-07-22
Make package publication retries safe

### LINKED ISSUE #645: Roll back package uploads on metadata failure
A recent package upload failed to upload metadata to the registry after the tarball was uploaded to storage. The author retried the publishing workflow, but publishing failed because there was already a tarball in storage. See:

https://github.com/purescript/registry/issues/336#issuecomment-1660974501
https://github.com/purescript/registry/issues/336#issuecomment-1660983820

Specifically, in this section of `publishRegistry`, the `Storage.upload` and `Registry.writeMetadata` functions can throw exceptions, and the process is aborted if so:

https://github.com/purescript/registry-dev/blob/8177cd30d7b7cf3379b8b728564def0ed438298c/app/src/App/API.purs#L644-L647

We should probably catch the `Registry.writeMetadata` exception and roll back the `Storage.upload` (ie. issue a `Storage.delete`) before fully exiting the pipeline. Alternately, we could have a sort of bracket functionality built into the publish pipeline where we record what resources have been modified and on publish failure we roll back all of those modifications.

thomashoneyman:
Closes #645 by making package publication retries safe when publication is interrupted between writing:

1. the package tarball to storage,
2. package metadata to the registry repository, and
3. the manifest to the registry index.

We don't delete any artifacts after a downstream failure (since we've historically treated them as immutable). Instead, we do forward reconciliation: verify and reuse the existing tarball, repair missing metadata or manifest state, and reject any conflicting artifact or changed source.

### Problem

Publication spans several independent systems and we can't do atomic commits. Previously, if the tarball upload succeeded but writing metadata or the registry-index manifest failed, retrying publication encountered an existing storage object and stopped. This left the package in a permanently incomplete state that required manual repair.

#### Reconcile existing storage

When an upload fails, we first check whether the requested package version actually made it into storage or not. If it does, the stored tarball is downloaded and verified against the hash and size of the newly produced tarball:

- If they match, the artifact is reused and publication continues.
- A missing, corrupt, or mismatched artifact is rejected and we error out (so we can still end up needing manual repair)
- Existing storage never gets overwritten.

This also handles uploads that succeeded (the package was written) but returned an error to the caller.

#### Repair incomplete downstream state

When metadata already contains the package version, then on publish we verify the stored tarball using the immutable hash and size recorded in metadata.

If the registry-index manifest is missing, the manifest embedded in the stored tarball must match the manifest produced from the current source. This prevents a moved or modified Git ref from changing an already-uploaded release.

Once verified, publication idempotently reissues both Git writes. Reissuing both writes is intentional. A failed push can leave a local checkout with a commit that has not reached its remote; an idempotent write gives the repository effect another opportunity to push it.

#### Preserve immutability

Retries fail instead of reconciling in these cases:

- storage contains a tarball with different bytes or hash,
- the current source manifest differs from the manifest in the stored tarball, or
- the registry index already contains a different manifest for the package version.

A package version is only treated as fully published once both metadata and its registry-index manifest are present.

### Why forward reconciliation instead of rollback?

Basically, we treat the tarball as an immutable publication artifact; once it has been verified, completing the missing downstream writes is safer and more reliable than deleting and recreating it.

This makes retries convergent:

```text
no durable state
  -> storage
  -> storage + metadata
  -> storage + metadata + manifest
```

Each retry either advances toward the complete state or rejects a conflicting immutable artifact, in which case something has gone horribly wrong and we have to go fix it.

### Tests

I updated our Run-based testing to go through each of these conditions:

- successful publication,
- an upload that completes but reports failure,
- metadata failure leaving storage-only state,
- manifest failure leaving storage and metadata state,
- successful retry through each incomplete state,
- corrupt stored tarball rejection,
- changed source manifest rejection,
- conflicting pre-existing tarball rejection, and
- final convergence to complete storage, metadata, and manifest state.

### DIFF
--- a/app/test/Test/Assert/Run.purs
+++ b/app/test/Test/Assert/Run.purs
@@ -13,7 +13,6 @@ module Registry.Test.Assert.Run
 import Registry.App.Prelude
 
 import Data.Array as Array
-import Data.Exists as Exists
 import Data.Foldable (class Foldable)
 import Data.Foldable as Foldable
 import Data.FunctorWithIndex (mapWithIndex)
@@ -97,9 +96,9 @@ type TEST_EFFECTS =
       + ()
   )
 
--- | A one-shot failure injected into an effect mock. The matching handler
--- | consumes the failure before returning its error, so one test can run a
--- | failed publication and its retry against the same mock state.
+-- | A failure injected into an effect mock. Matching handlers consume failures
+-- | from the front of a plan, allowing one test to fail successive publication
+-- | retries at each durable boundary while preserving their shared state.
 data TestFailure
   = FailStorageUploadAfterWrite
   | FailMetadataWrite
@@ -107,18 +106,19 @@ data TestFailure
 
 derive instance Eq TestFailure
 
-consumeFailure :: forall m. MonadEffect m => TestFailure -> Ref (Maybe TestFailure) -> m Boolean
+consumeFailure :: forall m. MonadEffect m => TestFailure -> Ref (Array TestFailure) -> m Boolean
 consumeFailure expected ref = liftEffect do
-  actual <- Ref.read ref
-  if actual == Just expected then do
-    Ref.write Nothing ref
-    pure true
-  else
-    pure false
+  failures <- Ref.read ref
+  case Array.uncons failures of
+    Just { head, tail } | head == expected -> do
+      Ref.write tail ref
+      pure true
+    _ ->
+      pure false
 
 type TestEnv =
   { workdir :: FilePath
-  , failNext :: Ref (Maybe TestFailure)
+  , failurePlan :: Ref (Array TestFailure)
   , logs :: Ref (Array (Tuple LogLevel String))
   , metadata :: Ref (Map PackageName Metadata)
   , index :: Ref ManifestIndex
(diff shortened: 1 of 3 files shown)


app/test/Test/Assert/Run.purs
```diff
@@ -142,45 +170,15 @@ runBaseEffects = do
...
+runRegistryMock metadataRef indexRef operation = do
+  failurePlan <- liftEffect $ Ref.new []
+  operation
+    # Registry.interpret (handleRegistryMock { metadataRef, indexRef, failurePlan })
+    # runBaseEffects
 
 runGitHubCacheMemory :: forall r a. CacheRef -> Run (GITHUB_CACHE + LOG + EFFECT + r) a -> Run (LOG + EFFECT + r) a
 runGitHubCacheMemory = Cache.interpret GitHub._githubCache <<< Cache.handleMemory
 
-runCompilerCacheMock :: forall r a. Run (COMPILER_CACHE + LOG + r) a -> Run (LOG + r) a
```
thomashoneyman, line 152: I ran some timings and this was really slow, especially in the retry scenarios I added. Instead, we can just use a tmpdir and run this the same as production.

app/test/Test/Assert/Run.purs
```diff
@@ -242,8 +245,12 @@ handleRegistryMock env = case _ of
     pure $ reply $ Right $ Map.lookup name metadata
 
   WriteMetadata name metadata reply -> do
-    Run.liftEffect (Ref.modify_ (Map.insert name metadata) env.metadataRef)
-    pure $ reply $ Right unit
+    failWrite <- consumeFailure FailMetadataWrite env.failurePlan
+    if failWrite then do
+      pure $ reply $ Left "Injected metadata write failure."
```
thomashoneyman, line 250: The failure ref is just so we can inject errors like this into the evaluation and see how it propagates out.

f-f (approved):

PR #785, opened 2026-07-22
Split Registry read and write effects

### LINKED ISSUE #722: Split the REGISTRY effect into REGISTRY_READ and REGISTRY_WRITE
See [here](https://github.com/purescript/registry-dev/pull/715#discussion_r2701328290) for context - it would be good if the more sensitive interaction - committing to the Registry - would be split off into its own effect, so that scripts and other apps that just want to read from the Registry (a much more innocent situation) could be allowed to do only that

thomashoneyman:
Closes #722 (and does a bit more). In that issue we were specifically concerned with splitting `REGISTRY` into read/write effects, so it's more obvious when a function is only reading from the registry and not modifying it. However, along the way, I noticed that we don't need the `GITHUB` effect in the read handler either, so I went ahead and removed it. I did a little overall audit through the codebase to make sure each function is using the least effects required to get the job done.

Some basic notes on what was done:

- split the Registry effect into `REGISTRY_READ` and `REGISTRY_WRITE`, while retaining `REGISTRY` as a convenience alias for workflows that need both
- give Registry reads a handler that emits only `LOG + AFF + EFFECT`; only the write handler retains `GITHUB`
- audit production, server, script, E2E, and test effect rows so each path receives the minimum Registry responsibility it needs
- remove transitive GitHub credentials, Octokit clients, caches, and interpreters from PackageSetUpdater, VerifyIntegrity, and the real Registry read-handler test
- preserve the shared Registry environment, memory cache, debouncer, pull machinery, and existing publication/retry behavior

### DIFF
--- a/app-e2e/src/Test/E2E/Scripts.purs
+++ b/app-e2e/src/Test/E2E/Scripts.purs
@@ -191,7 +191,7 @@ runDailyImporterScript = do
   result <- liftAff
     $ DailyImporter.runDailyImport DailyImporter.Submit resourceEnv.registryApiUrl
     # Except.runExcept
-    # Registry.interpret (Registry.handle registryEnv)
+    # Registry.interpretRead (Registry.handleRead registryEnv)
     # GitHub.interpret (GitHub.handle { octokit, cache, ref: githubCacheRef })
     # Log.interpret (Log.handleTerminal Quiet)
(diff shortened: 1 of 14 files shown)


f-f (approved):
wonderful stuff!

PR #786, opened 2026-07-22
Implement registry version check script

### LINKED ISSUE #426: Implement `registry-version-check` script
The packaging team regularly makes manual package sets releases. Manual releases are required, for example, when a package such as Halogen is upgraded. Invariably the new Halogen version will break other packages in the set, and so the Halogen release is not added at first. Over time the dependents upgrade to the new Halogen, at which point Halogen and its dependents can all be updated in one go to produce a new package set.

The packaging team needs to be able to tell what package versions have been uploaded to the registry and are more recent than the version of those packages in the package set. For example, if Halogen 6 is in the package set and Halogen 7 has been released (but could not be added automatically because it breaks downstream packages), then we need to be able to tell that there's a Halogen 7 release waiting to go into the package set.

To accomplish this, we need to add a new script to the registry. This script, most likely named `Registry.Scripts.VersionCheck`, should:

1. Read in the contents of the most recent package set (this can be read from the `package-sets` repository for now, but it will use the registry in the future)
2. Read in the contents of the `metadata` directory in the registry
3. Find all packages with a latest version higher than its version in the package set
4. Log those package versions to the console for review

thomashoneyman:
Closes #426 by adding a `.#version-check` script to report registry versions newer than the versions in the latest package set. This reuses the package set updater candidate traversal instead of being brand-new.

I ran the script against package set `77.13.1` and found 19 pending versions. Not all of them can compile together, but 14 of them do, and that's submitted as [purescript/registry#560](https://github.com/purescript/registry/issues/560).

Five versions were intentionally left out:

- `barlow-lens@1.0.0` breaks `morello@0.4.0`, which uses the old Barlow API and declares `<0.10.0`.
- `elmish@0.14.0` breaks `elmish-hooks@0.11.0`; meanwhile, `elmish-html@0.12.0` requires Elmish 0.14, so neither Elmish candidate can move until the dependent packages are compatible.
- `hyrule@2.4.0` removes modules still imported by `bolson@0.3.9`, `deku@0.9.24`, and `ocarina@1.5.4`.
- `node-child-process@12.0.0` changes the exit API and breaks `dotenv@4.0.3` and `node-execa@5.0.0`.

So this is the point of the script: surface upgrades that automatic daily updates could not accept so maintainers can identify and submit a compatible coordinated batch.

### DIFF
--- a/.github/workflows/package-set-version-check.yml
+++ b/.github/workflows/package-set-version-check.yml
@@ -1,0 +1,55 @@
+name: Package set version check
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: package-set-version-check
+  cancel-in-progress: false
+
+jobs:
+  version-check:
+    runs-on: ubuntu-latest
+    # Planning performs at most 100 incremental compile probes against the
+    # package set, each of which downloads and compiles the proposed set.
+    timeout-minutes: 120
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Install Nix
+        uses: DeterminateSystems/determinate-nix-action@v3
+
+      - name: Setup Nix cache
+        uses: DeterminateSystems/magic-nix-cache-action@v14
+        with:
+          use-flakehub: false
+
+      - name: Generate compile-verified version check report
+        run: nix run .#package-set-version-checker -- --output version-check.md
+
+      - name: Upload report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: package-set-version-check
+          path: version-check.md
+          if-no-files-found: warn
+          retention-days: 14
+
+      - name: Open report issue
+        # The checker exits successfully without writing a report when there
+        # is no package set to check, so only open an issue when the report
+        # file exists.
+        if: ${{ hashFiles('version-check.md') != '' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh issue create \
+            --title "Package set version check ($(date --utc +%Y-%m-%d))" \
+            --body-file version-check.md
--- a/SPEC.md
+++ b/SPEC.md
@@ -784,7 +784,7 @@ A package set update is an object with two keys: `compiler`, an optional field t
 ```jsonc
 { // Sets the package set compiler version to 0.15.2
   "compiler": "0.15.2",
-  "packages" {
+  "packages": {
     // Updates the `aff` package to v8.0.0
     "aff": "8.0.0",
     // Removes the `argonaut` package from the package sets altogether
@@ -812,13 +812,15 @@ The suggested new package set can only be released if:
 1. All packages in the set depend only on other packages in the set
 2. All packages in the set can be compiled together
 
+Package set self-containment concerns dependency names, not the version ranges in package manifests. Every dependency name must occur in the package set, but its manifest range need not include the exact version selected by the set. Compiling the exact package versions in the set with the set's exact compiler version is the authoritative validity check.
+
 To verify the package set, we take the following steps:
 
-1. We apply the suggested changes to the package set and verify that the package set is still self-contained. If not, the update is rejected.
+1. We apply all suggested changes to the package set as one exact batch and verify that the package set is still self-contained. If not, the update is rejected.
 2. We install the previous package set and compile it. This ensures we are beginning from a known good state.
 3. We install the new package set (uninstalling any packages that are removed) and compile the package set again. If compilation fails then the update is rejected.
 
-When we have verified the package set is self-contained and all packages compile together then we can release the package set.
+When we have verified the package set is self-contained and all packages compile together then we can release the package set. A submitted package set update is atomic: either its entire exact payload is released or it fails without releasing a subset.
 
 ##### Releasing the Package Set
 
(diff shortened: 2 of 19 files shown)


f-f:
This is cool - two things come to mind:
- we should have a policy of dropping packages that are holding on updates for too long. E.g. picture the scenario where a package that has many dependants is blocked by a package that has not many users - the former should not be held up. Maybe we can have some sort of expiration (e.g. holding a package for 30 days triggers removal or something like that)
- we could wire this up to @pacchettibotti automatically opening issues in the repos that need updating to the newer version, to speed up the process of the package sets catching up to things

thomashoneyman:
Yeah, I agree. I don't have a specific policy in mind but directionally I'm on board. To what degree should that be encoded into this script's report? We could augment this with more data than it currently gives.

thomashoneyman:
Also, we could extend this with a planner that makes it more obvious whether the discovered versions will even work together and suggest a batch. Something like:

1. Give every candidate package two choices: its current version or reported candidate.
2. Check every package manifest against the exact proposed package-set versions.
3. Maximize the number of accepted candidates while satisfying all declared ranges.
4. Repeat until stable, since rejecting one candidate can invalidate another; like for example, rejecting Elmish 0.14 also requires rejecting `elmish-html@0.12.0`.
5. Run one read-only atomic compilation of the selected plan to prove it

For this batch, this would have discovered the 14 packages that work without me having to do a separate manual triage:

- `morello` excludes Barlow 1.0.
- `elmish-hooks` excludes Elmish 0.14, which consequently excludes the new `elmish-html`.
- `bolson` and `deku` exclude Hyrule 2.4.
- `dotenv` and `node-execa` exclude node-child-process 12.

...this doesn't help with what you're describing, but maybe we can re-think this script such that it's a more useful swiss-army-knife for helping get the package sets updated with the best options when it can't be done automatically due to breaking changes.

f-f:
> To what degree should that be encoded into this script's report?

I to get us a good glance we could have:
- who is holding up who
- and for how long; that is: how many days ago the incompatible version was published
- how many dependants each of the packages involved have

Also I think we could be running this in a weekly cronjob thing in the Registry repo, opening an issue/PR if it finds something

thomashoneyman:
d283822 has some followups that address your points:

1. The version check now produces JSON and markdown reports, which themselves cover staleness, direct and transitive dependant counts, blocked version ranges, reverse blockers, compiler compatibility, etc.
2. It also includes a planner which uses our bounded ranges; it considers current and compatible newer releases, maximizes upgrade count and then freshness, and supports coordinated upgrades
3. You can optionally compile the planned payload with `--verify` so that you _know_ the proposed plan will work
4. Adds a weekly workflow which writes to a rolling issue with that week's report (doesn't actually submit anything)

Tried it out with a live run against package set `78.1.0` / compiler `0.15.15` and it produced a workable plan — including settling on `hyrule@2.3.9` as an ugrade since `2.4.0` is blocked.

thomashoneyman:
Latest commit is a prototype, still hacking on it a little. But as I was working on the version check script, I realized we could be doing a lot more automatically — and still safely — to update the package sets. Our current automatic upgrade is basically just "try all new packages in the last 24hrs, and if any fail, go sequentially and only include ones which succeed." Very naive, and if a package doesn't work in its 24h window for whatever reason it is never retried — even if it _would_ work later on (such as a dependency finally upgrading).

So the new prototype implements one shared planner that can be used by both the package set updater and the version check workflow. Any version of a package that's already in the set which is _newer_ than what's in the set is eligible for any new plan, plus recently uploaded packages even if they've never been in the set. We restrict to a single compiler version but otherwise ignore ranges. We still try all packages first, but then we do bounded best-first latest/intermediate/current searches.

The package set updater, if it is capable of finding a nonempty addition/upgrade plan, can just submit it. If it finds removals, downgrades, etc. then we can have that notify trustees for action, but it will never be automatically done.

thomashoneyman:
I've squashed this to a single commit and trimmed it down. Summary of what shipped vs. what I described earlier in the thread:

**Kept**

- One shared planner, `Registry.App.PackageSetPlanner`, used by both the package set updater and the version-check script. Candidates are every registry version newer than the set's for packages already in the set, plus recent uploads as addition seeds, closed over absent dependency names. Manifest ranges are ignored; the compiler decides.
- Compile-guided search: one whole-set probe first, then greedy repair from compiler output, per-root retry at latest and intermediate versions, and per-component probing for coordinated upgrades. `verified` is always exactly a payload that compiled, so the updater submits it as-is. Bounded at 60 planning probes / 100 total; the typical run is one probe.
- The server is now strictly atomic: `upgradeSequential` is gone and the API recompiles the exact submitted payload or publishes nothing. SPEC §5.5/§6.3 updated.
- The version-check script (`nix run .#package-set-version-checker`) reports compile-verified automatic upgrades plus, for blocked groups, the reverse-dependency removal closure verified by a real compile, with compiler evidence and a ready-to-submit trustee payload. It finds the `elmish`/`elmish-html`/`elmish-hooks` case from set 78 automatically (it's a test fixture).

**Dropped or deferred**

- The range-based planner and JSON report from d283822 — replaced by the compile-guided planner. The compiler is the authority; the ranges were a second model to keep in sync.
- Staleness / dependant-count / "who blocks whom for how long" columns and any expiration policy (@f-f's suggestion). The report shows blockers and evidence per group; the richer triage data and a drop-after-N-days policy are follow-ups once we've seen a few real reports.
- Weekly cron + rolling issue. The workflow is `workflow_dispatch` only and opens a new issue per run; adding `schedule:` is a one-line change once we like the output.
- pacchettibotti opening issues in blocking repos.
- Downgrade-requiring upgrades aren't analyzed; the API rejects downgrades and they'd need their own authenticated path.

Issue #790, opened 2026-07-27
Implement search and versioned Prim docs for the new Pursuit


thomashoneyman:
The new Pursuit application in #789 will serve package and module documentation mostly as static HTML, but we still need a server for search. We also need to replace the way Pursuit currently gets documentation for `Prim`. (Content below is written with Sol assistance).

## Search

Current Pursuit builds an in-memory search index from the latest version of every package. It supports searching for packages, modules, declarations, and approximate type signatures, and uses reverse dependencies as part of its ranking.

We have another PureScript implementation of much of this functionality in [Spago's `docs-search`](https://github.com/purescript/spago/tree/master/docs-search). We should move or adapt the reusable parts of that implementation into `registry-dev`, in particular:

* the type query parser;
* normalization of types into searchable shapes;
* approximate type matching and ranking;
* reconstruction of signatures for constructors and type class members.

We should adapt this code to use the compiler-independent docs types from #704 rather than the compiler's internal type representation.

Spago's storage and loading model is intended for a local static documentation site, so we should not preserve the JavaScript index shards or browser-side tries. Instead, we can build an index in SQLite as packages are processed by the new Pursuit.

At minimum the index needs to represent:

* packages and package versions;
* modules;
* declarations and child declarations;
* normalized type signatures and searchable type shapes;
* package popularity or reverse-dependency counts;
* the compiler versions supported by each package version.

The docs JSON remains the source of truth. The SQLite index is derived data and should be fully rebuildable. Processing the same package version again should replace its rows transactionally, and unpublishing a version should remove them.

Normal search should probably continue to use only the latest version of each package, as Pursuit does today, while keeping older versions available through their direct documentation URLs. We should preserve the useful parts of the current search behavior, including approximate type matching, package popularity in ranking, result URLs and fragments, and the existing JSON/CORS behavior used by clients.

One reason compiler compatibility was added to registry metadata in #669 was to eventually support filtering Pursuit search results by compiler version. Since we now have that information, the new index and API should support filtering by a compiler version or range without having to regenerate declaration rows whenever a package's compiler matrix changes.

We should compare representative queries against both current Pursuit and Spago's docs search before switching production over. The algorithms do not have to be identical, but surprising changes in result quality should be intentional.

## Prim documentation

Pursuit currently gets `Prim` documentation from the PureScript compiler library it was built with. That means the available builtins implicitly change when Pursuit updates its compiler dependency, and there is no way to view the `Prim` docs associated with an older compiler.

[...]

Issue #789, opened 2026-07-27
Migrate Pursuit application to the monorepo


thomashoneyman:
As part of #750 we need to replace the parts of Pursuit which assume packages come from GitHub repositories with SemVer tags. #704 tracks defining and storing a registry-owned docs format, and #525 tracks generating that format without calling `purs publish`. Once those are in place we still need an application that can actually serve those docs.

I think the new Pursuit should live in this repository rather than in a separate repository. It can be its own PureScript package and executable, separate from the registry server itself, but then we can share types and rendering code directly and use the same Nix and deployment infrastructure. The existing Haskell Pursuit can remain the reference implementation while we build this, but it's not a port, it's a rewrite.

The rough model discussed in #704 is that Pursuit would be mostly static:

* The canonical docs JSON for each package version lives in a DigitalOcean bucket.
* We generate package and module HTML from that JSON ahead of time and upload it to the same bucket (or another static origin).
* A small Pursuit server remains for things which are necessarily dynamic, like search, redirecting to the latest version, listing available versions, and badges.
* Package source links go to the [registry package viewer](https://github.com/purescript/registry-package-viewer) instead of GitHub.

The package viewer should probably move into this repository too, even if it is still deployed as an independent static application. That gives us one place for all of the code that makes up Pursuit.

We should preserve the public parts of Pursuit that people and package managers depend on. In particular:

* Existing package and module URLs should continue to work.
* Declaration fragments should remain stable.
* Links to dependencies and re-exports should use the exact versions recorded in the docs.
* The available-versions endpoint and badges should keep working.
* Old `purescript-foo` URLs should permanently redirect to proper registry names such as `foo`, completing the Pursuit side of #388.
* Existing unversioned URLs should continue redirecting to the latest version.

The registry will be the only publisher to the new Pursuit. We don't need to recreate the existing GitHub-token upload flow because package identity and authorization have already been handled by the registry before the docs artifact is generated.

We will also need to migrate existing documentation. For packages and versions in the registry we should be able to regenerate docs from their stored tarballs. For historical packages which only exist in Pursuit, Nathan's migration code in #704 can convert the contents of `pursuit-backups` and fetch any missing README files. We should do this progressively and keep track of packages which can't be migrated, rather than making the whole migration all-or-nothing. We still need to decide whether non-registry historical packages should remain visible by default.

[...]

Issue #750, opened 2026-02-02
Complete support for monorepos / subdir key and non-GitHub locations


thomashoneyman:
The registry is designed not to depend directly on GitHub and to support multiple packages stored in one repository. However, we've explicitly disabled both of those behaviors for now:

https://github.com/purescript/registry-dev/blob/2343b10ea03fa39d8260785b58e23094a1fb41ac/app/src/App/Effect/Source.purs#L81-L96

This is the overall tracking issue for removing those restrictions. There are two legacy systems we need to deal with first: Pursuit and legacy Spago package sets.

## Pursuit

The current documentation pipeline runs `purs publish`, which requires a Git repository with a clean worktree and a SemVer-shaped tag. Pursuit itself also assumes packages are on GitHub when it fetches READMEs, builds source links, authenticates uploads, and names packages.

We're replacing that pipeline with registry-owned documentation and a new Pursuit application maintained in this repository:

- [ ] #704 defines and tests a canonical registry docs format and stores generated docs in the registry.
- [ ] #525 generates those docs during registry publication without calling `purs publish`, while still supporting retries and regeneration.
- [ ] #789 moves the Pursuit application into this repository, renders the registry docs, migrates existing documentation and URLs, and cuts `pursuit.purescript.org` over to the new application.
- [ ] #790 implements package, module, declaration, and type search along with compiler-versioned `Prim` documentation.

A working production Pursuit is a requirement for the first monorepo or non-GitHub package. We don't want to temporarily accept packages which can be installed from the registry but are absent from Pursuit.

Once this work is complete, package documentation will be generated from the compiler output already produced during registry publication, stored by the registry, and rendered without accessing a Git forge. Source links will use the registry package viewer instead of trying to construct GitHub URLs. The registry will be the only publisher to Pursuit.

This work also addresses the remaining Pursuit-specific problems described in:

- #388
- https://github.com/purescript/pursuit/issues/422

## Legacy Spago package sets and manifests

We currently produce legacy `packages.dhall` files for the old Spago package manager. Those package sets bake in assumptions about package source being under `src` in a GitHub repository and can't represent all of the locations we want the registry to support.

We also still accept Bower and legacy Spago manifests as a temporary compatibility measure.

The deprecation work is already tracked separately:

- [ ] #744 stops accepting legacy Spago and Bower manifests.
- [ ] #745 stops generating legacy package sets.

[...]

PR #791, opened 2026-08-03
Migrate registry-docgen to registry-dev

### LINKED ISSUE #704: Proposal: store docs.json in the registry
As part of #525 should we consider storing `docs.json` in the registry?

* The registry pipeline is already compiling projects with an appropriate compiler. We can easily add the appropriate codegen targets.
* The registry pipeline is the sole publisher to Pursuit nowadays. It seems unnecessary to support additional authentication workflows in Pursuit.
* If we store docs metadata in the registry, pursuit largely becomes a stateless registry client in principle, even though it likely needs to build it's own index.
* We have more freedom to adjust the actual format (especially wrt metadata) since the registry controls the data end-to-end.

Questions:
* Where is this actually stored? Package metadata is stored in a single-file ldjson format, but we would want to store version specific files, likely.
* Are there format changes we need to make? Is there redundant information in the data we currently publish to Pursuit that is otherwise available in the registry?
* We may need a data migration from pursuit-backups.
* We may need some additional postback mechanism to notify Pursuit of newly published packages if we don't want to rely on a cron job that pulls the registry metadata.

thomashoneyman:
Closes #704. This moves @natefaubion's [`purescript-registry-docgen`](https://github.com/natefaubion/purescript-registry-docgen) into this repository as the top-level `docgen` package and makes it the foundation for registry-owned package documentation, as discussed on Discord.

The goal is to build doc artifacts that we can store in a bucket, and those become the source of truth for Pursuit. HTML, search indexes, and other data can be rebuilt from the docs without needing to reach back out to GitHub or any dependency on `purs publish` directly.

## What changed

The new `docgen` package includes lots of good things built by Nathan:

- A canonical docs schema, which is versioned and has the associated codecs
- Functions to convert docs from compiler/historical Pursuit documentation
- Deterministic package generation using the registry manifest and dependency resolutions
- Re-export resolution including selective imports, hidden imports, aliases, etc.
- Package and module HTML rendering, README rendering
- Exact packag eversion links for dependencies and re-exports
- Test covering codecs, generation, etc.

We use the registry manifest for almost everything, but since we have all the Pursuit backups we can use that for the documentation content itself. 107 package versions out of the ~6000 total have different dependency ranges in the registry and the docs used technically invalid versions, so we'd want to regenerate those.

## Documentation storage

The main registry change in here, beyond the `docgen` target, is adding a separate `DocsStorage` effect with filesystem and S3 interpreters. Documentation is derived and replaceable, unlike immutable package tarballs, so it supports checking for existing artifacts, downloading, create-only uploading, explicitly replacing, and deleting docs.

The existing package storage effect has been renamed from `Storage` to `PackageStorage` to make the distinction clearer.

The production Nix configuration now supplies `DOCS_BUCKET`, defaulting to the `purescript-registry-docs` space that I created last week. DigitalOcean spaces cannot make the existence check and upload atomic, so production generation will still need to ensure there is only one writer for a given package version. That constraint is tracked in #525.

## Historical documentation

This adds scripts for rendering documentation locally and validating historical Pursuit artifacts against registry manifests.

The migration audit found:

- 6,543 historical Pursuit artifacts;
- 4,240 artifacts corresponding to registry versions considered for migration;
- 4,133 which decode and agree with their authoritative registry manifests;
- 107 which should be regenerated from registry tarballs rather than imported;

All 107 regeneration candidates have stored tarballs and at least one recorded compatible compiler. This PR adds the conversion and validation machinery but does not perform the production migration.

## Scope

This PR is for defining, validating, rendering, and storing documentation artifacts, but it doesn't do any of the following:

- generate or upload them during package publication;
- add retry or regeneration jobs;
- perform the production migration;
- replace the existing Pursuit application;
- add search or compiler-versioned `Prim` documentation;
- change package-source handling for monorepos.

Those are covered by #525, #789, #790, and #750.

The imported and adapted code is included under the repository's BSD-3-Clause license with permission from Nathan.


flake.nix
```diff
@@ -103,12 +104,28 @@
           lib = pkgs.lib;
           rootPath = self;
         };
+
+        smoke = pkgs.callPackage ./nix/test/smoke.nix {
+          inherit overlays;
+          rootPath = self;
+        };
+
+        smokeTcg =
```
thomashoneyman: I've been using Amp orbs for some registry development, and it cannot use KVM so this bails out immediately. I added this separate target so the tests can still be run there, but I can take it back out if you prefer @f-f.

docgen/src/Registry/Docgen/Decoder.purs
```diff
@@ -0,0 +1,145 @@
+module Registry.Docgen.Decoder where
```
thomashoneyman, line 1: The Docgen tree is brought in basically as-is, which includes modules like this one — not sure if we _need_ this separate JSON decoding module, but I didn't want to futz around with the implementation.

natefaubion: This was because:
* Legacy decoder doesn't need to round-trip. That would bloat the decoders significantly since there's a lot of sum types.
* We aren't depending on argonaut, AFAIK

There's not a great ecosystem for purescript-json, other than codec.

docgen/src/Registry/Docgen/Package/Render.purs
```diff
@@ -0,0 +1,559 @@
...
+          [ H.meta [ H.attr "charset" "utf-8" ]
+          , H.htmlTitle
+              [ H.text title ]
+          , H.link
+              [ H.href "https://fonts.googleapis.com/css?family=Roboto+Mono|Roboto:300,400,400i,700,700i"
+              , H.type_ "text/css"
+              , H.rel "stylesheet"
+              ]
+          , H.link
+              [ H.href "https://pursuit.purescript.org/static/res/css/normalize.css?etag=fKzu1nci"
```
natefaubion: Note that all of these refer to existing deployed assets. I think for "production" these might need some alternative way of specifying these. Especially if spago is ever going to generate docs completely locally.

docgen/src/Registry/Docgen/Commonmark.js
```diff
@@ -0,0 +1,46 @@
+import * as commonmark from 'commonmark';
+
+const absoluteUriRegex = /^(https|mailto):/i;
+
+function isAllowedUri(uri) {
+  // return uri.startsWith('#') || uri.startsWith('/') || absoluteUriRegex.test(uri);
```
natefaubion, line 6: I don't remember what this is about exactly, but it probably needs some consideration. Right now only absolute uri's are allowed but I don't know if we want to do any other kind of linking, or even munging of links.

docgen/src/Registry/Docgen/Convert.purs
```diff
@@ -0,0 +1,366 @@
...
+  | PrecApp
+  | PrecInfix
+  | PrecArrow
+  | PrecKinded
+  | PrecTop
+
+derive instance Eq TypePrec
+derive instance Ord TypePrec
+
+precOfType :: DocType -> TypePrec
```
natefaubion, line 343: Some amount of maintenance risk here, though it's relatively benign. The converter automatically brackets things for presentation. If anything new is ever added, this needs to be addressed. It's also possible there are bugs lurking here (precedence is hard!). Bug fixes here will need docs regenerated, of course.

docgen/src/Registry/Docgen/Decoder.purs
```diff
@@ -0,0 +1,145 @@
...
+traversedAtIndex (Star f) = Star $ traverseWithIndex \ix value ->
+  catchError (f value) \err ->
+    throwError $ DecodeError.withPath (JP.AtIndex ix) err
+
+guardLength :: Int -> Decoder (Array JSON) Unit
+guardLength n = Star \arr ->
+  if Array.length arr == n then
+    pure unit
+  else
+    throwError $ DecodeError.basic $ "Exected Array of length " <> show n
```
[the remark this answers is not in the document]
thomashoneyman: Fixed in 49086c3, thanks.

app/src/App/Effect/DocsStorage.purs
```diff
@@ -0,0 +1,225 @@
...
+  writeDocsFile install path docs = Run.liftAff $ Aff.attempt do
+    FS.Extra.ensureDirectory $ Path.dirname path
+    tempDir <- FS.Aff.mkdtemp $ Path.concat [ Path.dirname path, ".docs-" ]
+    Aff.finally (FS.Extra.remove tempDir) do
+      let tempPath = Path.concat [ tempDir, Path.basename path ]
+      writeJsonFile Docgen.Codec.docPackage tempPath docs
+      install tempPath path
+
+  nodeErrorCode :: Aff.Error -> String
+  nodeErrorCode error = (unsafeCoerce error :: { code :: String }).code
```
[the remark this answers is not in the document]
thomashoneyman: Fixed in 49086c3. Rather than adding a bespoke FFI binding, this now decodes the optional `code` property through `Foreign` as `Maybe String`, so an arbitrary `Error` cannot be treated as a Node system error without validation.

docgen/src/Registry/Docgen/Package/Render.purs
thomashoneyman: Addressed in 4cf4356. The three stylesheets previously loaded from the Pursuit deployment are now vendored under `docgen/assets` with their license notices. `renderDocument` takes a `DocumentAssets` record so production and local consumers choose the URLs, while `render-docs` copies the bundled assets to `/static/css`; the Nix-packaged executable ships and uses the same assets. The Google Fonts stylesheet remains an explicit, optional external asset.

docgen/src/Registry/Docgen/Convert.purs
thomashoneyman: Addressed the small maintenance risk in 4cf4356 without moving presentation precedence into the renderer: `precOfType` now matches every current `DocType` constructor explicitly, so adding a constructor requires a compiler-guided decision here. I also added focused coverage for converted function types and kind signatures used as type arguments, where parentheses are required.

docgen/src/Registry/Docgen/Package/Render.purs
```diff
@@ -201,37 +216,27 @@ htmlCodeRenderer { getRefLink } currentModule =
...
-              ]
-          , H.link
-              [ H.href "https://pursuit.purescript.org/static/res/css/extra.css?etag=d4aey1o-"
-              , H.type_ "text/css"
-              , H.rel "stylesheet"
-              ]
+          , foldMap stylesheet assets.fontStylesheet
+          , stylesheet assets.normalizeStylesheet
+          , stylesheet assets.pursuitStylesheet
+          , stylesheet assets.extraStylesheet
```
natefaubion, line 230: It's not obvious to be that we need to turn each stylesheet into an individual asset in the record. Rather, just configuring stylesheets is probably ok.

```
type DocumentAssets = 
  { stylesheets :: Array String
  }
```

docgen/src/Registry/Docgen/Package/Render/Code.purs
```diff
@@ -0,0 +1,335 @@
...
+    , space
+    , syntax "::"
+    , space
+    , renderType code kind
+    , syntax ")"
+    ]
+
+renderTypeVarsWithRoles :: forall rep. Monoid rep => CodeRenderer rep -> Array TypeVar -> Array Role -> rep
+renderTypeVarsWithRoles code@{ space } vars roles =
+  intercalate space $ Array.zipWith (\var r -> renderTypeVar code var <> renderRole code r) vars roles
```
mikesol, line 178: zipWith is evil, I find that when I use it, I basically have to ask myself if I'm ok with one of the sources entirely disappearing because the second isn't there. Just make sure to test throughly, and ideally if there's another pattern even better!

docgen/src/Registry/Docgen/Package/Render/Code.purs
```diff
@@ -0,0 +1,335 @@
...
+    ]
+
+renderDeclData :: forall rep. Monoid rep => CodeRenderer rep -> DeclDataRep -> rep
+renderDeclData code@{ keyword, line, space } { isNewtype, name, roles, signature, vars } =
+  fold
+    [ foldMap (line 0 <<< renderDeclSignature code dataKw name) signature
+    , line 0 $ keyword dataKw
+        <> space
+        <> renderQualified code NSType name
+        <> renderTypeVarsWithRoles code vars roles
```
mikesol, line 223: make sure to put a space here so that data Foo bar doesn't become data Foobar

docgen/src/Registry/Docgen/Package/Render/Code.purs
```diff
@@ -0,0 +1,335 @@
...
+    ]
+  where
+  dataKw
+    | isNewtype = "newtype"
+    | otherwise = "data"
+
+renderDeclForeignData :: forall rep. Monoid rep => CodeRenderer rep -> DeclForeignDataRep -> rep
+renderDeclForeignData code@{ line } { name, roles, signature } = fold
+  [ line 0 $ renderDeclSignature code "foreign import data" name signature
+  , line 0 $ renderTypeVarsWithRoles code [] roles
```
mikesol, line 233: why are the vars empty here? Wouldn't that drop info?

mikesol (changes requested):

docgen/src/Registry/Docgen/Package/Render/Code.purs
natefaubion: It's an invariant of the compiler's data model that all vars have a role, so I think this code is correct. I don't think we should be trying to account for bugs in the compiler.

docgen/src/Registry/Docgen/Package/Render/Code.purs
natefaubion: We would have to confirm with the existing renderer. Technically foreign data doesn't have vars syntactically, but the compiler does generate them for correctness. This whole "line" seems dubious. Presumably this doesn't render anything.

